PR #1, opened 2015-09-02
Properly parse multiple ECU names

trainman419:
My DBC file has multiple ECU names; this properly parses them.


eerimoq:
Thanks for the patch.

If you have time, consider adding a unit test so the error is not introduced again later.

PR #2, opened 2017-04-08
Add ability to read cycle times from DBC

jpsamper2009:
The idea of the fork was to be able to encode cycle times in the same way that Vector's CANdb++. I also made sure that the print function prints a CANdb++ compatible format (it seems to be picky about ordering and whitespace).

### DIFF
--- a/tests/test_cantools.py
+++ b/tests/test_cantools.py
@@ -8,7 +8,7 @@ class CanToolsTest(unittest.TestCase):
 
     def test_vehicle(self):
         f = cantools.db.File()
-        filename = os.path.join('tests', 'files', 'vehicle.dbc')
+        filename = os.path.join('files', 'vehicle.dbc')
         with open(filename, 'r') as fin:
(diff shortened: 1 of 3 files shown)


tests/test_cantools.py
```diff
@@ -24,7 +24,7 @@ def test_vehicle(self):
 
     def test_motohawk(self):
         f = cantools.db.File()
-        filename = os.path.join('tests', 'files', 'motohawk.dbc')
+        filename = os.path.join('files', 'motohawk.dbc')
```
eerimoq: The tests on Travis fails (https://travis-ci.org/eerimoq/cantools/jobs/220106672), partly due to this path change. Please fix.

You can run the tests locally with ``python setup.py test`` and ``python3 setup.py test`` in the repository root folder.

eerimoq:
Merged, but it would be appreciated if you add test cases for the new code. The risk of breaking your code in the future is huge if no tests are implemented.

PR #9, opened 2017-09-25
Fixes #8: Messages always encoded with 8 bytes.

### LINKED ISSUE #8: Messages always encoded with 8 bytes
Hello,

Thanks for putting this nice library out!

I found what I think is a defect: when I encode a message, it is always returned on full length of 8 bytes, even though a smaller length is correctly set in the description file (Kayak in my case).

Steps to reproduce:
1- Have a message of total length 1 byte defined in a .kcd file. Example (just the message itself):
```
<Message id="0x275" length="1" name="Command">
  <Notes>A command</Notes>
  <Signal name="Start" offset="0" length="8" endianess="little" />
  <Producer>
    <NodeRef id="0"/>
  </Producer>
</Message>
```
2- Encode it:
```
import cantools
db = cantools.db.load_file('Test_file.kcd')
messages = {msg.frame_id: msg for msg in db.messages}
print(messages[0x275])
# Printing: message('Command', 0x275, False, 8, 'A command')
data = messages[0x275].encode({'Start': 1})
print(len(data))
# Printing: 8
```

Not only does it returns 8 bytes long data. The Message object itself believe it is of length 8.

I am willing to find the source of the problem and submit a PR. But before, I wanted to know if this was an intentional feature or if this is a bug?

Cheers,
Axel

AxelVoitier:
Fixed by:
- Reading length attribute in kcd format.
- Shorting the returned data to specified length with python slice operator.

### DIFF
--- a/cantools/db/formats/kcd.py
+++ b/cantools/db/formats/kcd.py
@@ -116,6 +116,8 @@ def _load_message_element(message, bus_name):
             name = value
         elif key == 'id':
             frame_id = int(value, 0)
+        elif key == 'length':
+            length = int(value)
         elif key == 'format':
             is_extended_frame = (value == 'extended')
         else:
@@ -136,7 +138,7 @@ def _load_message_element(message, bus_name):
     return Message(frame_id=frame_id,
                    is_extended_frame=is_extended_frame,
                    name=name,
-                   length=8,
+                   length=length,
                    nodes=[],
                    send_type=None,
                    cycle_time=None,
--- a/cantools/db/formats/kcd.py
+++ b/cantools/db/formats/kcd.py
@@ -1,6 +1,5 @@
 # Load and dump a CAN database in KCD format.
 import logging
-import math
 
 try:
     from StringIO import StringIO
@@ -139,7 +138,7 @@ def _load_message_element(message, bus_name):
     if length == 'auto':
         if signals:
             last_signal = sorted(signals, key=lambda s: s.start)[-1]
-            length = math.ceil((last_signal.start + last_signal.length) / 8.0)
+            length = (last_signal.start + last_signal.length + 7) // 8
         else:
             length = 0
     else:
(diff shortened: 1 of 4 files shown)


AxelVoitier:
I see tests failing. Let me fix that...

cantools/db/formats/kcd.py
```diff
@@ -120,6 +121,8 @@ def _load_message_element(message, bus_name):
             is_extended_frame = (value == 'extended')
         else:
             LOGGER.debug("Ignoring unsupported message attribute '%s'.", key)
+            
+    length = message.attrib.get('length', 'auto')
```
eerimoq: I suggest removing `, 'auto'` to make the length `None` is the length attribute is missing.

cantools/db/formats/kcd.py
```diff
@@ -132,11 +135,20 @@ def _load_message_element(message, bus_name):
 
     for signal in message.findall('ns:Signal', NAMESPACES):
         signals.append(_load_signal_element(signal))
+        
+    if length == 'auto':
```
eerimoq: Change to `if length is None`.

cantools/db/formats/kcd.py
```diff
@@ -132,11 +135,20 @@ def _load_message_element(message, bus_name):
...
+        if signals:
+            last_signal = sorted(signals, key=lambda s: s.start)[-1]
+            length = math.ceil((last_signal.start + last_signal.length) / 8.0)
```
eerimoq: I suggest replacing floating point calculation with integer calculations, as `length = ((last_signal.start + last_signal.length + 7) // 8)`. Floating point calculation is probably not an issue in this case, but sometimes rounding errors occurs.

AxelVoitier:
OK, check this out.

eerimoq:
Looks good, thanks for the contribution! I'll make a new release and upload it to PyPi soon (probably tonight).

AxelVoitier:
Neat! I should make other PR in the coming days concerning other issues with Kayak files. I let you see if you want to make new release that often or not ;).

PR #10, opened 2017-09-29
Support for missing or incorrect attributes in Kayak format

AxelVoitier:
Adding support for:
- Document's version attribute
- Bus' baudrate attribute
- Message's interval attribute

Also correcting the default endianess to little, as told by https://github.com/dschanoeh/Kayak/blob/master/Kayak-kcd/src/main/resources/com/github/kayak/canio/kcd/loader/Definition.xsd

### DIFF
--- a/cantools/db/bus.py
+++ b/cantools/db/bus.py
@@ -7,9 +7,11 @@ class Bus(object):
 
     def __init__(self,
                  name,
-                 comment=None):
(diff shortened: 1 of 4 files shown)


eerimoq:
Nice job!

eerimoq:
Released to PyPi as 11.0.0!

PR #11, opened 2017-10-30
Add support for float (single and double precision) in Kayak file format

AxelVoitier:
Hello,

Kayak file format has value types "single" and "double" for IEE754 floats. This PR adds support for encoding and decoding these types.

### DIFF
--- a/cantools/db/message.py
+++ b/cantools/db/message.py
@@ -15,7 +15,10 @@ def _encode_signal(signal, data, scaling):
                 break
 
     if scaling:
-        return int((value - signal.offset) / signal.scale)
+        if signal.is_float:
+            return (value - signal.offset) / signal.scale
+        else:
+            return int((value - signal.offset) / signal.scale)
     else:
         return value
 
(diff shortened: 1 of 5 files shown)


cantools/db/message.py
```diff
@@ -108,14 +117,20 @@ def _create_message_encode_decode_formats(signals):
     for signal in signals[::-1]:
         if signal.byte_order == 'big_endian':
             continue
+    
+        if signal.is_float:
+            signal_type = 'f'
+        elif signal.is_signed:
+            signal_type = 's'
+        else:
+            signal_type = 'u'
```
eerimoq: I suggest creating a function returning the format string type. Define the function within the scope of `_create_message_encode_decode_formats()`.

Somthing like:

``` python
def get_format_string_type(signal):
    if signal.is_float:
        return 'f'
    elif signal.is_signed:
        return 's'
    else:
        return 'u'
```

AxelVoitier: Yes, I wanted to do this initially. But knowing not everybody like nested functions I did without it just to be sure ^^.

AxelVoitier:
It could be something like data_type?
But merging both properties into one also means having some sort of enum for it.

Issue #12, opened 2017-11-27
SIG_VALTYPE_ & BO_TX_BU_


blackskandix:
While parsing a few new .dbc-files, I got stuck with cantools when reaching lines with the commands:

- "SIG_VALTYPE_" 
- "BO_TX_BU_"
- "BA_DEF_REL_"
  - "BU_SG_REL_"
- "BA_DEF_DEF_REL_"
- "BA_REL_"
  - "BU_SG_REL_"

just wanted to let you know

------------------

SIG_VALTYPE sets the type of the value of a signal in a message - example:
SIG_VALTYPE_ 1649 dCANxxTime : 2; 
Sets the datatype in message (0x671), for signal (dCANxxTime) to "2" (IEEE Double)

eerimoq:
Any possibility of sharing a DBC-file that can be used to reproduce the problem?

eerimoq:
Added support for `SIG_VALTYPE_` in the DBC parser in 0183cdb2b77f038041b9b09aa5737ad960b8f0f6. 

Supporting floating point signals should be easy to add in the future, whenever they are needed by someone.

blackskandix:
sorry, I'm not allowed to share the DBC-files 
I will see, if I can make a modified version of it, as soon as I find time

Anyways - thanks for responding

eerimoq:
Would be very helpful! =)

blackskandix:
update coming soon - don't close

blackskandix:
I finally got used to the unknown commands. 
You got a pull request for db/formats/dbs.py(260):

I really can't provide you any dbc-files for test purposes - I'm sorry
But here are some sample lines for syntax understanding.

`BO_TX_BU_ 780 : ACC,Frontradar;`
Message with Id "780" is added to List of Senders in the Nodes "ACC" and "Frontradar".
It seems like a way to add nodes as senders after the basic "BO_" defintion.
As in one of my files:
```
BO_ 780 ACC_02: 8 ACC //only node "ACC" is sender in the message defintion
 SG_ ...
 .
 .
BO_TX_BU_ 780 : ACC,Frontradar; //later "Frontradar" gets listed as well
```
`BA_DEF_REL_ BU_SG_REL_ "GenSigTimeoutTime" INT 0 65535;`
"BA_DEF_REL_": relative attribute gets defined, like "BA_DEF_". 
"BU_SG_REL_": Prefix for type, like "BO_" in "BA_DEF_" for Messages
[...]

PR #15, opened 2017-12-11
added Grammar for relative attributes

blackskandix:
just the grammar, so file parsing with these keywords gets possible.
data is not further processed at all

for additional information, sample lines and explanation, please visit: #12

### DIFF
--- a/cantools/db/formats/dbc.py
+++ b/cantools/db/formats/dbc.py
@@ -264,44 +264,44 @@ def _create_grammar():
                         - scolon)
     signal_type.setName(SIGNAL_TYPE)
 
-    message_add_sender = Group(Keyword('BO_TX_BU_')
+    message_add_sender = Group(Keyword(MESSAGE_TX_NODE)
                                - frame_id
                                - colon
                                - Group(delimitedList(node))
                                - scolon)
-    #message_add_sender.setName('BO_TX_BU_') #not needed?
+    message_add_sender.setName(MESSAGE_TX_NODE)
 
-    attribute_definition_relative = Group(Keyword('BA_DEF_REL_')
-                                          - (QuotedString('"', multiline=True)
-                                             | (Keyword('BU_SG_REL_')
-                                                + QuotedString('"', multiline=True)))
-                                          - word 
-                                          - (scolon
-                                             | (Group(ZeroOrMore(Group(
-                                                (comma | Empty())
-                                                + QuotedString('"', multiline=True))))
-                                                + scolon)
-                                             | (Group(ZeroOrMore(number))
-                                                 + scolon)))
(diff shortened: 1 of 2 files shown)


cantools/db/formats/dbc.py
```diff
@@ -264,6 +264,45 @@ def _create_grammar():
                         - scolon)
     signal_type.setName(SIGNAL_TYPE)
 
+    message_add_sender = Group(Keyword('BO_TX_BU_')
+                               - frame_id
+                               - colon
+                               - Group(delimitedList(node))
+                               - scolon)
+    #message_add_sender.setName('BO_TX_BU_') #not needed?
```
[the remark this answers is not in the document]
eerimoq: setName() just makes parser error messages easier to read.

cantools/db/formats/dbc.py
```diff
@@ -264,6 +264,45 @@ def _create_grammar():
...
+
+    attribute_relative = Group(Keyword('BA_REL_')
+                   - QuotedString('"', multiline=True)
+                   - Keyword('BU_SG_REL_')
+                   - word
+                   - Keyword(SIGNAL)
+                   - frame_id
+                   - word
+                   - positive_integer
+                   - scolon)
```
eerimoq: Fix indentation to match the rest of the grammar definition.

eerimoq:
Thanks, looks good in general, just a code style comment.

Have a look at the failed tests. Looks like the parser error test case does not pass, which is expected as the exception message now includes the new grammar. I suggest adding setName() and then update the test case.

cantools/db/formats/dbc.py
```diff
@@ -264,6 +264,45 @@ def _create_grammar():
                         - scolon)
     signal_type.setName(SIGNAL_TYPE)
 
+    message_add_sender = Group(Keyword('BO_TX_BU_')
```
eerimoq: Use the already declared variable `MESSAGE_TX_NODE`. Apply similar change to other already declared keywords.

blackskandix:
not sure about the tests-part, but to me it seems like the String in tests/test_cantools.py(869):
simply needs an update to:
```
self.assertEqual(
            str(cm.exception),
            "Invalid DBC syntax at line 1, column 1: '>!<abc': Expected "
            "{VERSION | NS_ | BS_ | BU_ | BO_ | CM_ | BA_DEF_ | BA_DEF_DEF_ "
            "| BA_ | VAL_ | VAL_TABLE_ | SIG_VALTYPE_ "
            "| BO_TX_BU_ | BA_DEF_REL_ | BA_DEF_DEF_REL_ | BA_REL_ | EV_}.")
```
to match with the additional keywords. 
But I really don't know anything about your testcases.(?)

PR #18, opened 2017-12-20
Implementation BO_TX_BU_ and modifable Attributenames

blackskandix:
parsing BO_TX_BU_  (#12) lines now effects the loaded db's
I also experienced a few dbc-files, where key-attributes where not named with the default-names like "GenMsgCycleTime" - so I replaced the hard-coded strings with vars, so they can get modified.

Please look away from the wierd commit-history -  the github-website messed up a few edits, that weren't supossed to be

Feel free to merge and/or modify every part you like

Finaly, here is a tiny dbc-file to show the new functionality with the command:
[sample-BO_TX_BU_.zip](https://github.com/eerimoq/cantools/files/1574979/sample-BO_TX_BU_.zip)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,7 @@ messages and signals.
    >>> db = cantools.db.load_file('tests/files/motohawk.dbc')
    >>> db.messages
    [message('ExampleMessage', 0x1f0, 8, 'Example message used as template in MotoHawk models.')]
-   >>> example_message = db.messages[0]
+   >>> example_message = db.get_message_by_name('ExampleMessage')
    >>> pprint(example_message.signals)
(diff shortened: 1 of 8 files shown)


eerimoq:
I refactored the additional message senders code. It's part of f932fd75a1cb0d99494561ac3d0913f90ef9884a, and release 14.4.0.

Configuration of attribute names should not be global variables, but a configuration per File-instance. Otherwise bigger scripts using multiple DBC-files with different keywords may fail.

blackskandix:
ok, I'm glad it helped :)

regarding the the global vars, I worked with the importlib so far, as workaround - so I didn't needed to modify your code to much:
```
import importlib, cantools_master
importlib.reload(cantools_master.db.dbc)
cantools_master.db.dbc.ATTR_NAME_send_type = .. if .. else ..
...
db.add_dbc_file( ... )
```

I'm closing then now - I may share my modifications, if I refactor them again

Issue #24, opened 2018-02-09
For database objects, attribute_definition_defaults and attribute_definitions should defaults to dicts, not lists


rjw245:
Otherwise, I see the following error:
    default_send_type = database.attribute_definition_defaults['GenMsgSendType']
TypeError: list indices must be integers, not str

rjw245:
This also applies to File objects

eerimoq:
These attributes in the Database class is an ugly hack to preserve the information from a parsed DBC-file. The interface to access such attributes should be reworked to be clean and useful. I have not made up my mind yet of how to design the final interface. Any ideas?

rjw245:
I think in either case, you're mapping the attribute name to a value right? Dictionary seems most appropriate for both, IMO.

eerimoq:
Does it help you if they are both dictionaries? What is the use case? As said, the storage of attributes should be redesigned at some point, and maybe dictionaries will be used then.

Will you submit a PR with a fix for the wrong default values?

PR #25, opened 2018-02-12
attribute_definition_defaults and attribute_definitions default to empty dictionaries.

rjw245:
Resolves issue #24


cantools/db/database.py
```diff
@@ -34,10 +34,10 @@ def __init__(self,
         self._version = version
         self._attribute_definitions = (attribute_definitions
                                        if attribute_definitions
-                                       else [])
+                                       else {})
```
eerimoq: Leave this as a list as in DBC-format.

See https://github.com/eerimoq/cantools/blob/master/cantools/db/formats/dbc.py#L534-L541.

cantools/db/database.py
```diff
@@ -34,10 +34,10 @@ def __init__(self,
...
         self._attribute_definition_defaults = (attribute_definition_defaults
                                                if attribute_definition_defaults
-                                               else [])
+                                               else {})
```
eerimoq: Make the same change in https://github.com/eerimoq/cantools/blob/master/cantools/db/internal_database.py.

cantools/db/database.py
rjw245: Addressed in 34d11f81e538a5b8399cbb58477c67716b693977

cantools/db/database.py
rjw245: Addressed in 34d11f81e538a5b8399cbb58477c67716b693977

Issue #34, opened 2018-05-16
as_dbc_string()  database's method loses the extended flag


dmyger:
Sample of code with the issue:
```
db = cantools.db.Database()
db.add_dbc_file('some.dbc')
ext_msg = db.get_message_by_name('SomeExtendedMessage')

if ext_msg.is_extended_frame:
    db_out = cantools.db.Database()
    db_out.add_message(ext_msg)
    with open('result.dbc', 'w') as dbc_result:
        dbc_result.write(db_out.as_dbc_string())
```
Check `frame_id` in the result output and see that `0x80000000` bit is absent.
And if open `result.dbc` with your library it also show the `is_extended_frame` flag is `False`:
```
db = cantools.db.Database()
db.add_dbc_file('result.dbc')
for m in db.messages:
	print m
```

eerimoq:
The `as_`-methods are not the primary feature of this package, and should probably be removed since they are not fully functional. Either that, or I should add a limitation in the README notifying the user about their flaws. That said, feel free to implement the missing feature if you find it useful.

dmyger:
It seems, that exactly for that issue, it would be easy set a high bit for the extended frame in file `cantools/db/formats/dbc.py` for each case where is used `message.frame_id`, something like this
```
    ...
    msg.append(fmt.format(frame_id=message.frame_id | 1<<31 if message.is_extended_frame else 0,
    ...
```
Instead of removing the working functionality.

eerimoq:
Feel free to create a pull request with the change and I'll review and later merge it. Make sure to modify/add tests to ensure the issue is not reintroduced in the future. Thanks!

eerimoq:
Part of release 20.9.0.

PR #37, opened 2018-05-17
Added missing information to dbc dump.

juleq:
I know that you do not consider the as_ functions as API, but have them working is helpful to me. Thank you for your work here.

Update: Fixed double space for non muxed signal dumps.

### DIFF
--- a/cantools/db/formats/dbc.py
+++ b/cantools/db/formats/dbc.py
@@ -380,11 +380,13 @@ def _dump_messages(database):
                               senders=' '.join(message.senders)))
 
         for signal in message.signals[::-1]:
-            fmt = (' SG_ {name} : {start}|{length}@{byte_order}{sign}'
+            fmt = (' SG_ {name} {muxType}{muxId} : {start}|{length}@{byte_order}{sign}'
                    ' ({scale},{offset})'
                    ' [{minimum}|{maximum}] "{unit}" {receivers}')
             msg.append(fmt.format(
                 name=signal.name,
+                muxType='M' if signal.is_multiplexer else 'm' if signal.multiplexer_ids != None else '',
+                muxId='' if signal.multiplexer_ids == None else signal.multiplexer_ids[0],
                 start=signal.start,
                 length=signal.length,
                 receivers=', '.join(signal.receivers),
--- a/cantools/db/formats/dbc.py
+++ b/cantools/db/formats/dbc.py
@@ -370,23 +370,33 @@ def _dump_nodes(database):
 
 def _dump_messages(database):
     bo = []
+    
+    def get_frame_id(message):
+        return message.frame_id | (1 << 31 if message.is_extended_frame else 0)
+
+    def get_mux(signal):
+        result = ''
+        if signal.is_multiplexer:
+            result += ' M'
+        elif signal.multiplexer_ids != None:
+            result += ' m' + str(signal.multiplexer_ids[0])
+        return result
 
     for message in database.messages:
         msg = []
         fmt = 'BO_ {frame_id} {name}: {length} {senders}'
-        msg.append(fmt.format(frame_id=message.frame_id | (1 << 31 if message.is_extended_frame else 0),
+        msg.append(fmt.format(frame_id=get_frame_id(message),
                               name=message.name,
                               length=message.length,
                               senders=' '.join(message.senders)))
 
         for signal in message.signals[::-1]:
-            fmt = (' SG_ {name}{muxType}{muxId} : {start}|{length}@{byte_order}{sign}'
+            fmt = (' SG_ {name}{mux} : {start}|{length}@{byte_order}{sign}'
                    ' ({scale},{offset})'
                    ' [{minimum}|{maximum}] "{unit}" {receivers}')
             msg.append(fmt.format(
                 name=signal.name,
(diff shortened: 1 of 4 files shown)


eerimoq:
Hello,

please implement a test and I'll review again.

juleq:
#34 I have pulled this also in here since I was having a look at the dump functionality.

tests/test_database.py
```diff
@@ -1588,7 +1588,42 @@ def test_multiplex_bad_multiplexer(self):
...
+
+        self.assertEqual(dumped_msg.signals[0].name, "MultiplexorSig")
+        self.assertEqual(dumped_msg.signals[0].is_multiplexer, True)
+        self.assertEqual(dumped_msg.signals[0].multiplexer_ids, None)
+        self.assertEqual(dumped_msg.signals[1].name, "MultiplexedSig")
+        self.assertEqual(dumped_msg.signals[1].is_multiplexer, False)
+        self.assertEqual(dumped_msg.signals[1].multiplexer_ids[0], 0x2A)
+        self.assertEqual(dumped_msg.signals[2].name, "UnmultiplexedSig")
+        self.assertEqual(dumped_msg.signals[2].multiplexer_ids, None)
+        self.assertEqual(dumped_msg.signals[2].is_multiplexer, False)
```
eerimoq: I recommend testing like in this test: https://github.com/eerimoq/cantools/blob/master/tests/test_database.py#L1074-L1089

That is, read the DBC, check some data, and then compare `as_dbc_string()` to the original file contents.

cantools/db/formats/dbc.py
```diff
@@ -374,17 +374,19 @@ def _dump_messages(database):
     for message in database.messages:
         msg = []
         fmt = 'BO_ {frame_id} {name}: {length} {senders}'
-        msg.append(fmt.format(frame_id=message.frame_id,
+        msg.append(fmt.format(frame_id=message.frame_id | (1 << 31 if message.is_extended_frame else 0),
```
eerimoq: This line is over  80-90 characters long. Either add line breaks, or create a local function called `get_frame_id(message)` and put the logic in that.

cantools/db/formats/dbc.py
```diff
@@ -374,17 +374,19 @@ def _dump_messages(database):
...
 
         for signal in message.signals[::-1]:
-            fmt = (' SG_ {name} : {start}|{length}@{byte_order}{sign}'
+            fmt = (' SG_ {name}{muxType}{muxId} : {start}|{length}@{byte_order}{sign}'
                    ' ({scale},{offset})'
                    ' [{minimum}|{maximum}] "{unit}" {receivers}')
             msg.append(fmt.format(
                 name=signal.name,
+                muxType=' M' if signal.is_multiplexer else ' m' if signal.multiplexer_ids != None else '',
+                muxId='' if signal.multiplexer_ids == None else signal.multiplexer_ids[0],
```
eerimoq: Lines too long. I think I prefer creating a local function `get_mux()` or similar that returns the mux string.

cantools/db/formats/dbc.py
```diff
@@ -445,7 +447,10 @@ def _dump_attribute_definitions(database):
                                          name=attribute[2],
                                          type_=attribute[3],
                                          choices=choices))
-
+        elif attribute[0] in [ATTRIBUTE_DEFINITION]:
```
eerimoq: Why not `elif attribute[0] == ATTRIBUTE_DEFINITION:`?

tests/test_database.py
```diff
@@ -1588,7 +1588,42 @@ def test_multiplex_bad_multiplexer(self):
...
+        self.assertEqual(attr[0][0], "BA_DEF_")
+        self.assertEqual(attr[0][1], "BusType")
+        self.assertEqual(attr[0][2], "STRING")
+
+    def test_extended_id_dump(self):
+        filename = os.path.join('tests', 'files', 'test_extended_id_dump.dbc')
+        db = cantools.db.load_file(filename)
+        dumped_db = cantools.db.load_string(db.as_dbc_string())
+        reg_id_msg = dumped_db.get_message_by_frame_id(0x100)
+        ext_id_msg = dumped_db.get_message_by_frame_id(0x1C2A2A2A)
```
eerimoq: Change `0x1C2A2A2A` to `0x1c2a2a2a`.

tests/test_database.py
juleq: Another example: Vector seems to output two newlines after the VERSION"" token. Your test files have one. If you have also used CANdb++, they might have slightly changed the dbc output filter used in CANdb++...

eerimoq: The goal is to output DBC-files on the same format as CANdb++. Unfortunately, I've not used that tool myself, so I don't know its DBC-file formatting. Leave your file as outputted by CANdb++, and I'll consider changing the whitespace formatting later.

juleq:
Thank you. Please comment here if I can help out with the Vector tooling I have access to. That would be CANoe 9.x and CANalyzer 10.x. Both are accompanied by CANdb++, but I would have to look up their versions. When starting from scratch, CANdb++ offers several templates. I used the empty one. Alternatives are J1939, CANoe, etc. I think they differ in predefined attributes. Best.

eerimoq:
Any help to improve the quality of cantools is welcome. Reference test data is always useful to achieve high quality, but I can't recall any test in particular that needs improvement or should be added.

Issue #38, opened 2018-05-22
Question: state of message/signal attributes?


rjw245:
Hi, just wanted to ask a question.

The last I checked, cantools did not properly support database-, message-, and signal-level attributes passed to constructors. Has this since been updated? I am hoping to build up a CAN database using cantools classes and assign certain attributes to the objects within the database, and finally render this to a DBC file. Is this possible at this time?

Thanks

rjw245:
I last asked about this in issue #27

eerimoq:
No, attributes are still not properly implemented, sorry.

Issue #27, opened 2018-02-25
Attribute definitions do not appear in output of Database::as_dbc_string()


rjw245:
I am instantiating a Database with an attribute definition like so:

```
Database(messages, nodes, buses, version,
                           attribute_definitions=[
                                                      ('BA_DEF_',
                                                       'BO_',
                                                       'GenMsgCycleTime',
                                                       'INT',
                                                       [ '0',
                                                         '1000000'
                                                       ],
                                                      )
                                                 ]
)
```

And one of the messages has a non-null cycle_time attribute set, which does show in the DBC. However, the attribute definition itself does not. I can't seem to figure out why.

I did notice that the _dump_attribute_definitions() function in formats/dbc.py seems to expect attribute definitions in a different format than they show when going FROM  a dbc TO  a Database object; see below:

```
>>> db = cantools.db.load_file("mydbc.dbc")
>>> print db.__dict__
...
 '_attribute_definitions': [(['BA_DEF_', 'BO_', 'GenMsgCycleTime', 'INT', (['0', '1000000'], {})], {})],
...
```
As you can see there is a lot more nesting going on than what I passed to the constructor -- the simpler format seems to be what _dump_attribute_definitions expects.

rjw245:
Never mind, I think I had missed something in my testing -- instantiating the Database's attribute_definitions the way I did DOES cause them to appear in the DBC. However, they did not when I passed them in in the format they appear when loading FROM a DBC. I think the two different formats should be reconciled in some way.

eerimoq:
Indeed, the attribute storage is a hack. It is not recommended to be used before redesigned. A redesign is not likely to happen anytime soon unless someone submits a pull request with the implementation.

PR #40, opened 2018-06-12
Added signal, message and node level attributes. 

juleq:
Changed whitespace handling and had to do minor whitespace tweaks to existing test files. 
Corrected minor quirks such as "Vector_XXX" is not a valid node name. CANdb will remove it. 
Or: Standard CAN ids outside of the 11bit range. CANdb++ will not accept that.

Sorry for putting all of that in one commit.


juleq:
I will work on the coverage. Any idea what I could do with the line ending issue that makes the Python 2 tests fail?

juleq:
This might also be of interest for the poster of #27 and #38.

juleq:
ENUM index resolving :).

juleq:
My last commit was supposed to increase coverage. It did so for Python 2 but it did lower it for the Python 3 run. Do you see why?

cantools/db/database.py
```diff
@@ -24,6 +24,7 @@ def __init__(self,
                  nodes=None,
                  buses=None,
                  version=None,
+                 attributes=None,
```
eerimoq: What's the reason for adding a list attributes to the database? I mean, isn't this information already part of messages, signals and nodes? Am I missing something?

cantools/db/message.py
```diff
@@ -527,6 +527,13 @@ def is_multiplexed(self):
 
         return bool(self._codecs['multiplexers'])
 
+    def get_dbc_frame_id(self):
+        frame_id = self.frame_id
+        if self.is_extended_frame:
+            frame_id |= 0x80000000
+
+        return frame_id
+
```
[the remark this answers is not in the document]
juleq: Agreed. Will update PR.

cantools/db/database.py
eerimoq: I don't have access to the Vector tools myself, could you take a few relevant screenshots to give me a better understanding of the database structure?

The main goal for this package is to make it easy to use (mainly to encode and decode messages), and at the same time complete. The second most important goal is to make the interface independent of the database format. When it comes to attributes the goal should be to make them useful to the user. I think a dictionary of attribute definitions (BA_DEF_) with the attribute name as key and various data (range, type, ...) as value may be useful. The attribute definition default (BA_DEF_DEF_) should be part of this value, and not be in a separate dictionary as today.

When it comes to signal, message and node attributes, they are probably most often accessed from the signal, message or node objects. `send_type` and `cycle_time` are two examples of message attributes. Maybe a separate attribute list could be useful, but I'm not sure. Is there a use case?

juleq: You are right in that my perspective was very much from the dbc side of things. I just thought that serialisation (dump) of attributes would be a mess again if I scatter them all over the place. But from a usage perspective 1.) iterating the aggregate is ugly when one could use dicts and 2.) navigating multiple collections to harvest all info on one attribute is too.

I think I will come up with a more usage friendly approach. If you have further wishes that I shoud consider when implementing, do not hesitate.

My use case BTW is just dbc IO. I read them in, manipulate them, and dump them out. Since I use the dbcs with CANoe Interaction Layers in test setups, I need support for all these colorful attributes that control the layers behaviour.

A second use case for me is to render html doc from dbcs.

eerimoq: I would be lovely if you can write an example showcasing your use case. Could help others as well.

Other than that I think my previous comment summarizes my current view of how to implement it. Though, I'm open to have internal representations of the attributes if it simplifies the implementation and makes the code easier to understand and maintain.

At last, most importantly, user friendliness is key to success!

eerimoq:
Also, all members prefixed with and underscore should not be considered private and should not be used directly by the user. So I guess a we should create a property allowing the user access to attribute definitions, and other useful data if any.

cantools/db/database.py
eerimoq: Also, all members prefixed with and underscore should not be considered private and should not be used directly by the user. So I guess a we should create a property allowing the user access to attribute definitions, and other useful data if any.

juleq: Hi there,

I will proceed as follows: Make an independent attribute definition dict (by name). Database will hold that. The attributes themselves will be objects containing value and a reference to the respective definition. The attributes will be stored in dicts in node/signal/message.

I think that was your proposal, correct?

eerimoq: It's very hard to design the message/signal/node-classes to be database format agnostic. Preferably DBC-naming of certain attributes should not be part of these classes, today `send_type`, etc. is the abstraction in these classes. I realize that I don't know enough about other database formats to make a good design of cantools package, so we have to implement something based on what we know now and what we need. I think we should proceed as outlined in your comment. Maybe think once more about the attribute dictionary in signals/messages/nodes. At least the new implementation should not clobber other database formats implementations.

juleq: Ok. I am on it right now. To avoid redundant storage, i want to change send_type into properties that lookup in the dicts on the fly.

Since it would be ugly to have a back reference to the top level (database) to access the definitions, I propose to give node/message/signal each their own reference to the definition dict. That solves the issue of fallback to default when an attribute is not available. Do you agree?

juleq: Lets discuss that later. I will stick to the current implementation right now. Doing lookups is complicating interoperability with other formats.

eerimoq: Please create a new PR with the minor fixes (standard id, whitespace and valid node names), and I'll merge that before the rest is ready.

eerimoq: I don' know it it makes any sense, but what do you think about this:

https://github.com/eerimoq/cantools/blob/dbc-attributes/examples/dbc_builder.py

Since I have absolutely no idea about what different formats have in common I figured, why not just store all format specific data in a separate database attribute; for DBC the dbc attribute. I also kept attributes as lists just as you, and added a get method. Nodes and messages are already using the same pattern.

juleq: It is a neat idea to keep format specifics in such a container. It is unlikely that other formats use exactly the same properties on attributes, so why litter the generic message with specifics. And one can still implement transformation functions if one need to translate attributes when changing formats.

I am half way there and already close to your proposal.

Do you insist on the intermediate PR? Would be a lot of extra effort for me. I promise I will up my granularity in the future :).

eerimoq: Nah, I thought you paused the implementation and I was afraid I would lose valuable fixes.

juleq: In your proposal, is the „dbc“ one shared object or is it just a „namespace“ so separate e.g. the signal-specific attribute collection?

If it is kind of a namespace, do I use a local class for that?

juleq: Hi. Short update:

I had started using dicts for the node/message/signal attributes and also for the defaults. That deviates from your proposal but eases access by name. I have switched to lists briefly but hesitatet because I would need traversal functions (e.g. get_attribute_by_name) in several places.

Do you think you could live with that?

I have a working version now that still lacks the dbc specifics separation (that is message.attributes instead of message.dbc.attributes). If you could give me a hint how to do this separation, I would also implement that (Its just a few weeks of Python for me so far :).

eerimoq: I have no string feeling about lists vs dicts at the moment. It could always be changed later. Most importantly the user interface should be easy to use.

This is one alternative:

``` python
class Database(object):

    @property
    def dbc(self):
        # Return a DBC object, probably called self._dbc.
```

juleq:
I´ll work on the coverage, but you could already have a look.

eerimoq:
I think it looks decent, but it's a lot of changes and it's hard to understand all implications of the changes. It would help my understanding if you could write an example and a few more tests.

juleq:
Yes, I will do so. Maybe I can do the first commits tomorrow.

When I am done with this I could also tell you some findings that I made along the way. E.g. dbc files in the test folder that CanDb++ will not parse due to invalid order of contents (parser is sequence sensitive) or EV_ meaning Environment Variable and not Event. Regarding the whitespaces: I have corrected a few but have not caught every single strangeness their output logic yields. E.g. one space or one newline if a group or a field is missing. Will be a lottle mor trial and error in the dumpsters :).

eerimoq:
Lovely! A DBC dump that actually is usable in other tools. =)

juleq:
I am doing an example right now. I would want to add setters to various properties (in classes msg, sig...) of so that I can read in, manipulate without touching the _privates and dump out again. Objections against setters?

eerimoq:
Should be ok.

juleq:
Since the two methods get_message_by_name and get_message_by_frame_id use dictionary lookup, offering setters for message.name and message.frame_id will produce inconsistencies. An easy way around would to change the lookup to a search (like e.g. get_signal_by_name) and take a performance hit with large databases.

Another more complicated approach would be to make the lookup a class functionality of the message class. That way the setter could alter the dictionary when it is called.

The least invasive approach would be to just issue a warning as soon as the problematic setters are called. E.g. "Changing the frame id will make the results of database.get_message_by_frame_id inconsistent. Dumping and rereading will fix the dict again."

I think manipulation of these properties is quite a feature and I would really like to have that. I would be ok with the first approach (search instead of lookup). As a user, you can always save a reference for consecutive accesses or build your own dict.

What do you think?

juleq:
Hmm. Just saw that the dicts are your primary way of accessing for encode/decode. I guess that makes it performance sensitive. I will think of another way.

juleq:
Status update:
- Regarding the message dict: I have chosen to do a search if lookup fails. If search is also failing, I raise the KeyError just like before. If I find something, I update the respective dictionary.
- Regarding the dbc specifics containment: Have not done yet. I thought that you could give me feedback on the example first.
- Regarding coverage: I have pushed to get a fresh coveralls result. I have tried a python coverage package to check coverage locally, but I was not able to produce comparable results. If you do local coverage testing, I would take a hint :)

I consider this push fully functional but I will add a test to increase coverage again.

juleq:
Regarding coverage locally: Ok, I missed the --source switch. It is working now.

cantools/db/attribute_definition.py
```diff
@@ -0,0 +1,98 @@
...
+    @property
+    def kind(self):
+        """The attribute kind (BU_, BO_, SG_), or ``None`` if unavailable.
+
+        """
+
+        return self._kind
+
+    @property
+    def type_(self):
```
eerimoq: I think I prefer `type_name` over `type_`, as ending a name with underscore in the interface looks weird. Internally I'm fine with ending variables with underscore if the name is a Python keyword.

cantools/db/attribute_definition.py
```diff
@@ -0,0 +1,98 @@
...
+    @property
+    def name(self):
+        """The attribute name as a string.
+
+        """
+
+        return self._name
+
+    @property
+    def defaultValue(self):
```
eerimoq: Use lower case with underscore, `default_value`. This applies to all methods, functions and variables.

examples/dbc_io/main.py
```diff
@@ -0,0 +1,35 @@
...
+#import cantools
+import sys
+sys.path.append('/mnt/d/JQU/dbc/cantools')
+import cantools
+
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
+DBC_IO_DBC_PATH = os.path.join(SCRIPT_DIR, 'dbc_input.dbc')
+
+infile = DBC_IO_DBC_PATH
+outfile = infile.replace('_input', '_output')
```
eerimoq: What if _input is part of the user file path? I suggest:

```
SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
INPUT_DBC_PATH = os.path.join(SCRIPT_DIR, 'dbc_input.dbc')
OUTPUT_DBC_PATH = os.path.join(SCRIPT_DIR, 'dbc_output.dbc')

```

examples/dbc_io/main.py
```diff
@@ -0,0 +1,35 @@
...
+
+# Read the dbc
+db = cantools.db.load_file(infile)
+# Get a message to manipulate
+msg = db.get_message_by_name('TheMessage')
+# Manipulate a message attribute. Try/except will catch accesses to non existent attributes
+try:
+    print("Input cycle time: " + str(msg.dbc.attributes['GenMsgCycleTime'].value))
```
eerimoq: I prefer: 

``` python
    print("Input cycle time:", msg.dbc.attributes['GenMsgCycleTime'].value)
```

examples/dbc_io/main.py
```diff
@@ -0,0 +1,35 @@
...
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
+DBC_IO_DBC_PATH = os.path.join(SCRIPT_DIR, 'dbc_input.dbc')
+
+infile = DBC_IO_DBC_PATH
+outfile = infile.replace('_input', '_output')
+
+# Read the dbc
+db = cantools.db.load_file(infile)
+# Get a message to manipulate
+msg = db.get_message_by_name('TheMessage')
```
eerimoq: Rename `msg` to `message`. I know both names exists in the code base, but I prefer `message`.

examples/dbc_io/main.py
juleq: Do you mean just the space or do you want to get rid of the try/except to not fail silently? If the latter, would propose to output a meaningful message instead of pass. That way, the user gets an idea of the preferred usage (would say that is try/except).

eerimoq: I just meant the print-statement.

Best case all exception thrown by cantools includes a descriptive error message. I'm fine with showing the backtrace to the user, but it's often nicer to catch the exception and just print its message (or another message if the exception message is unclear).

juleq: I could catch and raise again.

juleq: Ok, got it now. The concatenation. However, I also raise instead of pass now.

eerimoq:
I do not run code coverage analysis locally.

juleq:
I would consider this done if you have no objections. Can´t wait to see it pulled :).

juleq:
Will continue tomorrow with styling. You are right, having them as class property did gain nothing. And I have not even thought about the multiple dbc use case...

juleq:
I am not sure about the policy regarding blank lines: I have identified so far:

- After local variable definition
- Before and after if statements
- Before return

I would further assume the following, but your feedback said otherwise:
- Before and after local functions
- Before and after for if it enhances readability or separates groups ( e.g. iterating nodes/messages/signals for the same purpose)

Could you clarify? Thanks.

eerimoq:
I wrote a coding style guide in the documentation. It is by no means complete, but should be a good start.

http://cantools.readthedocs.io/en/latest/#coding-style

eerimoq:
Let me know when you are ready and I'll merge. I can make the last coding style changes myself once merged.

juleq:
I am ready. Thank you for that style guide. I would suggest including local functions and maybe a nested statement (for/if) as an example.

eerimoq:
Sorry, I should have added that. I'll make the additions to the coding style rules and clean up the code =)

rjw245:
Only just seeing this -- thanks for implementing this juleq

juleq:
@rjw245 Thank @eerimoq for his patience with me (and the heavy lifting)... :)

PR #41, opened 2018-06-13
Add utf-8 encoding when loading dbc from file

JulianWgs:
This option should either be standard or at least an option since it causes errors, when extracting string information from the created database.

System: Python 3.6.5 64bit, Windows 10

It breaks python 2.7 support.

### DIFF
--- a/cantools/db/database.py
+++ b/cantools/db/database.py
@@ -102,7 +102,7 @@ class Database(object):
 
         """
 
-        with open(filename, 'r') as fin:
+        with open(filename, 'r', encoding="utf-8") as fin:
             self.add_dbc(fin)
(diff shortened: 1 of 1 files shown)


eerimoq:
An optional encoding argument was added to [load_file()](https://github.com/eerimoq/cantools/blob/master/cantools/db/__init__.py#L42-L70). We should probably make the same fix in all places where `open()` is called.

JulianWgs:
Thanks for the quick response! The new release indeed fixes the problem. Sorry for not being up to date.

eerimoq:
The new release was created yesterday as a fix for this issue. So you were up to date.

PR #57, opened 2018-08-29
Support for messages with no senders in dump subcommand.

bessman:
Vector's CANdb++ tool can create .dbc-files containing messages with no senders. The cantools dump subcommand currently chokes on such files. This change adds support for messages with no senders.

Technically, only the format_and function needs to be changed, but I also changed the format_or function for consistency and robustness.

### DIFF
--- a/cantools/__init__.py
+++ b/cantools/__init__.py
@@ -140,6 +140,9 @@ def _do_dump(args):
         if cycle_time is None:
             cycle_time = '-'
 
+        if len(message.senders) == 0:
+            message.senders.append('-')
+
         print()
         print('  Name:       {}'.format(message.name))
         print('  Id:         0x{:x}'.format(message.frame_id))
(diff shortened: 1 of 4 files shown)


eerimoq:
Wih your change the dump subcommand output is `Senders:    None`. I suggest making the empty list output `Senders:    -`, as `None` is an internal Python type and it should not be exposed to the user. This can be implemented with special handling of the empty list in the `_do_dump` function instead of changing `format_and/or`.

eerimoq:
Will you add a test of the bug fix, you do I have to do it myself? =)

It should be fairly easy to create a new DBC/KCD-file in tests/files and copy-paste the existing dump testcase in test_database.py.

bessman:
Give me a sec. I can't upload the .dbc I'm actually working with, so I'm trying to reproduce the problem in a new file.

PR #60, opened 2018-10-02
Parse numbers as Decimals in more situations to relax parser on some DBCs

elafargue:
Some DBCs define some signals attributes as INT but later on use a decimal value in BA_.

This parses every number as Decimal, which solves the parsing errors and has no noticeable
side effect I could tell.

### DIFF
--- a/cantools/database/can/formats/dbc.py
+++ b/cantools/database/can/formats/dbc.py
@@ -616,9 +616,9 @@ def _load_attributes(tokens, definitions):
 
         definition = definitions[attribute[1]]
 
-        if definition.type_name in ['INT', 'HEX', 'ENUM']:
+        if definition.type_name in ['HEX', 'ENUM']:
             value = int(value)
-        elif definition.type_name == 'FLOAT':
+        elif definition.type_name in ['INT', 'FLOAT']:
             value = Decimal(value)
 
         return Attribute(value=value,
--- a/tests/files/decimal.dbc
+++ b/tests/files/decimal.dbc
@@ -1,0 +1,38 @@
+VERSION ""
+
+NS_ :
+        NS_DESC_
+        CM_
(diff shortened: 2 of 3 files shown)


eerimoq:
I suggest adding a test case with your previous comment as test case description. It help us understand why INT is parsed as Decimal in the future.

elafargue:
Good point, added the test, let me know if it makes sense!

eerimoq:
Btw, do you know if CanDB++ allows floating point values for INT-attributes? I mean, that tool is the closest to a specification of the DBC format we can get. If that tool does not allow decimal values, neither should cantools.

elafargue:
Yes, CanDB++ is fine with those.

eerimoq:
Hi,

I created a new database in CanDB++ and defined an integer message attribute with default value 0 and range 0 to 10. When I tried to change this attribute to 1.5 on a message CanDB++ complained and automatically set it to 0.

However, if I manually write a DBC-file and set the message attribute to 1.5 CanDB++ can load it. The value displayed is 1, and if I save the unmodified database the value in the DBC-fiel is still 1.5.

As CanDB++ does not allow entering floating point numbers for an integer attribute I am reluctant to merge this PR. Maybe older versions of CanDB++ allowed it, but not the latest. Also, setting an integer type to a floating point value does not make sense. The float type should be used instead.

elafargue:
It really just improves compatibility with no perceivable side effects, but
it's your call!

eerimoq:
The question is; why does an integer attribute have a floating point value?

According to my test above CanDB++ does not allow it. Silently accepting badly formatted files does not help the user in my opinion. It's better to let her know about the problem.

Which tools does this PR improve compatibility with?

eerimoq:
@elafargue Any updates?

PR #65, opened 2018-11-07
Support choice tables in dump subcommand

karlding:
This change implements choice tables in the `dump` subcommand.


cantools/database/can/message.py
```diff
@@ -315,6 +315,28 @@ def format_level_lines(signal_names):
...
+
+        for signal in self.signals:
+            if signal.choices:
+                lines.append(signal.name)
+
+                choice_lines = []
+                for value, text in sorted(signal.choices.items()):
+                    choice_lines.append('{} {}'.format(value, text))
+
+                lines += add_prefix('    ', choice_lines)
```
eerimoq: I suggest slightly simplified code:

``` python
for signal in self.signals:
    if signal.choices:
        lines.append(signal.name)

        for value, text in sorted(signal.choices.items()):
            lines.append('    {} {}'.format(value, text))
```

cantools/subparsers/dump.py
```diff
@@ -42,6 +42,13 @@ def _do_dump(args, _version):
             for line in message.signal_tree_string().splitlines()
         ]))
         print()
+        print('  Signal choices:')
+        print()
+        print('\n'.join([
+            ('    ' + line).rstrip()
+            for line in message.signal_choice_string().splitlines()
+        ]))
```
eerimoq: I suggest not dumping "Signal choices:", including the title, if there are no signals with choices. Less verbose output is nice when possible.

karlding:
Did you want me to add a command line test using `tests/files/multiplex_choices.dbc`, or is this along the lines of what you were thinking?

I wasn't sure if you wanted me to use that DBC, since there's another message inside besides the one with a multiplexed signal that has choice values.

eerimoq:
You can create a new DBC-file, say ``dump_signal_choices.dbc``, for dump signal choices testing. The file should contain a message with two signals, both with choices. Other dump scenarios are already covered by existing dump tests it seems.

eerimoq:
Part of release 28.13.0, available on PyPI!

PR #68, opened 2018-11-15
Add support for generated CAN message IDs

karlding:
This adds support for generating CAN message ID `#define`s so you can use it to select the proper decoding function, instead of maintaining a list of message IDs and dealing with "magic numbers". You could also pass that to your CAN TX function, instead of filling in the ID manually.

I guess we _could_ use `enum` values instead, but I'm not too clear on whether that would be MISRA compliant (if that's something you're concerned about).


cantools/subparsers/generate_c_source.py
```diff
@@ -47,6 +47,8 @@
 #    define EINVAL -22
 #endif
 
+{can_id_defines}
```
eerimoq: The id is called `frame_id` throughout the code base, and I suggest the same naming here. That is, rename `can_id` to `frame_id`.

cantools/subparsers/generate_c_source.py
```diff
@@ -409,7 +411,12 @@ def _generate_message(database_name, message):
                                        encode_code=encode_code,
                                        decode_code=decode_code)
 
-    return struct_, declaration, definition
+    enum_code = '#define {}_CAN_ID_{} (0x{:02x}U)'.format(
```
eerimoq: Let's rename this to `frame_id_defines` and stick to defines instead of enums.

PR #72, opened 2018-11-24
Add support for generated CAN enumerated values

karlding:
This adds support for generating signal enumerated choice table `#define`s.

An alternative approach would probably involve generating C `enum` types instead, since the raw `struct` isn't what's being encoded and decoded. So something like:

```c
typedef enum {
  STATE_X = 0,
  STATE_Y,
  STATE_Z,
  NUM_FOO_STATE
} foo_state;

struct foo_debug_t {
  // ...
  foo_state state;
  // ...
}

ssize_t foo_debug_encode(
    uint8_t *dst_p,
    struct foo_debug_t *src_p,
    size_t size) {
  // ...

  if (src_p->state >= NUM_FOO_STATE) {
    return EINVAL;
  }
  
  uint8_t val = {
    STATE_X_VAL,
    STATE_Y_VAL,
    STATE_Z_VAL
  };
  // encode using val[src_p->state] as the physical value

  // ...
}

int foo_debug_decode(
    struct foo_debug_t *dst_p,
    uint8_t *src_p,
    size_t size) {
  // ...
  // convert raw to physical value and translate into an enum
  uint8_t val = src_p[0];

  switch (val) {
    case STATE_X_VAL:
    {
      dst_p->state = STATE_X;
      break;
    }
    case STATE_Y_VAL:
    {
      dst_p->state = STATE_Y;
      break;
    }
    case STATE_Z_VAL:
    {
      dst_p->state = STATE_Z;
      break;
    }
    default:
    {
      // Either raise an error, or coerce into a different field or something
      break;
    }
  }
  // ...
}
```

The complexity with this alternative approach lies in how people use enumerated values. If people only expect enumerated values to be sent (ie. `STATE_{X,Y,Z}_VAL`, instead of all possible values that fit in the specified signal length, then this works fine. Otherwise, you would need to have a way to decode to the non-enumerated value, or lose the benefit of any abstraction the `enum` provides by hiding the physical value.


tests/files/c_source/motohawk.h
```diff
@@ -40,6 +40,9 @@
 
 #define MOTOHAWK_FRAME_ID_EXAMPLE_MESSAGE (0x1f0U)
 
+#define MOTOHAWK_EXAMPLE_MESSAGE_ENABLE_VAL_DISABLED (0U)
+#define MOTOHAWK_EXAMPLE_MESSAGE_ENABLE_VAL_ENABLED (1U)
```
eerimoq: I think we should rename `VAL` to `CHOICE`, and possibly move `CHOICE` to the same position as `FRAME_ID` in the frame id defines. That is, `MOTOHAWK_CHOICE_EXAMPLE_MESSAGE_ENABLE_ENABLED`. Alternatively we move both `FRAME_ID` and `CHOICE` to the end of the defines to make naming consistent throughout the generated code. See the example below, where all symbols starts with `motohawk_example_message_[enable_]` or `MOTOHAWK_EXAMPLE_MESSAGE_[ENABLE_]`.

``` c
#define MOTOHAWK_EXAMPLE_MESSAGE_FRAME_ID (0x1f0U)

#define MOTOHAWK_EXAMPLE_MESSAGE_ENABLE_ENABLED_CHOICE (1U)

struct motohawk_example_message_t {
    ...
};

ssize_t motohawk_example_message_encode(...);

int motohawk_example_message_decode(...);
```

tests/files/c_source/vehicle.h
```diff
@@ -256,6 +256,56 @@
 #define VEHICLE_FRAME_ID_RT_SB_ACCEL (0x94a6c22U)
 #define VEHICLE_FRAME_ID_RT_SB_GYRO_RATES (0x94a6d22U)
 
+#define VEHICLE_RT_SB_INS_STATUS_INS_STATUS_VAL_CONVERGED (2U)
+#define VEHICLE_RT_SB_INS_STATUS_INS_STATUS_VAL_INITIALISED (1U)
+#define VEHICLE_RT_SB_INS_STATUS_INS_STATUS_VAL_UNINITIALISED (0U)
```
eerimoq: Let's sort the choices in ascending numerical order.

eerimoq: A choice value may be negative, and even a 64 bits value. We should generate the suffix (`U` above) based on the signal type.

tests/files/c_source/vehicle.h
```diff
@@ -256,6 +256,56 @@
...
+
+#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_YAW_VAL_VALID (1U)
+#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_YAW_VAL_INVALID (0U)
+#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_PITCH_VAL_VALID (1U)
+#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_PITCH_VAL_INVALID (0U)
+#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_ROLL_VAL_VALID (1U)
+#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_ROLL_VAL_INVALID (0U)
```
eerimoq: Lets separate the signals with a blank line as it makes it easier to read. I think it adds more value than blank line between messages.

``` c
#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_YAW_VAL_VALID (1U)
#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_YAW_VAL_INVALID (0U)

#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_PITCH_VAL_VALID (1U)
#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_PITCH_VAL_INVALID (0U)

#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_ROLL_VAL_VALID (1U)
#define VEHICLE_RT_SB_INS_ATTITUDE_VALIDITY_ROLL_VAL_INVALID (0U)
```

tests/files/c_source/vehicle.h
karlding: I'm not sure what you mean here. Currently this should only add `U` if the signal is unsigned (`signal.is_signed`). Wouldn't this work correctly?

Or are you saying that you want to cast against the specific type depending on the length of the signal?

tests/files/c_source/vehicle.h
karlding: Done

tests/files/c_source/vehicle.h
karlding: Done

eerimoq: Ohh, sorry, I didn't see that the `U` was only added for unsigned signals.

Yeah, I was thinking of a specific suffix depending on the type length. I know it's unlikely a choice value would not fit in an integer, but it could happen. I'm fine with adding support for this later.

eerimoq:
Thanks for the contribution! =)

PR #73, opened 2018-11-26
Add .PHONY targets to Makefile

karlding:
Currently the `Makefile` does not specify a `.PHONY` target, which causes `make` to treat the target as a file. Since the current `make` targets `{test,test_c,test-sdist,release-to-pypi}` do not create these target files, if a file of the same name exists in the repository the target will not work correctly (and will not run).

```bash
# Make a folder called test
# (I was using this for storing some DBC files I was testing)
mkdir -p test
# Doesn't do anything since the prerequisite is already fulfilled
make test
```

Adding `.PHONY` specifies that the target is not a file, such that the target will run the recipe regardless of whether there is a file of the same name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+.PHONY: test
 test:
 	python2 setup.py test
 	python3 setup.py test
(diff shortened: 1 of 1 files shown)


eerimoq:
Thanks! :)

PR #75, opened 2018-11-29
Remove leading underscores from C include guards

karlding:
Hey,

I'm trying to reduce the scope of undefined behaviour in the generated C code (hopefully leading to more portable code), and thought I'd start off with this quick change!

According to the C specification (7.1.3 Reserved Identifiers), include guards should not start with leading underscores:

> Each header declares or defines all identifiers listed in its associated subclause, and optionally declares or defines identifiers listed in its associated future library directions subclause and identifiers which are always reserved either for any use or for use as file scope identifiers.
> 
>   * All identifiers that begin with an underscore and either an uppercase letter or another underscore are always reserved for any use.
>   * All identifiers that begin with an underscore are always reserved for use as identifiers with file scope in both the ordinary and tag name spaces.

According to this, identifiers beginning with underscores are technically reserved for implementation. See this [StackOverflow](https://stackoverflow.com/questions/17307540/include-guard-conventions-in-c) post or [Wikipedia article](https://en.wikipedia.org/wiki/Include_guard#Discussion) for more info.

This changes the include guard format from `__FILE_H__` to `FILE_H`, thereby reducing the scope of undefined behaviour.

**Edit**: Seems like this is one of the linter errors raised in the MISRA bug, so I guess this also fixes the MISRA 2012 Rule 21.1 violations in the header files.


eerimoq:
Hello,

thanks, looks good!

I guess the goal of the generated C source is that it is correct, easy to read, reasonably efficient in runtime, and reasonably portable. All of this to make people want to use it in their projects.

PR #82, opened 2018-12-03
Fix floating point comparison usage in test harness

karlding:
Floating point numbers should not be directly compared. Instead, they
should be compared against an epsilon value which denotes the maximum
allowed error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,9 @@ CFLAGS := \
 	-Wall \
 	-Wextra \
 	-Wpedantic \
-	-Wdouble-promotion \
 	-Wconversion \
+	-Wdouble-promotion \
+	-Wfloat-equal \
 	-Werror
 CFLAGS += $(shell $(CC) -Werror $(CFLAGS_EXTRA) -c tests/dummy.c 2> /dev/null && echo $(CFLAGS_EXTRA))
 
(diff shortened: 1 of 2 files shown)


eerimoq:
I'm not sure this is the best approach to make `-Wfloat-equal` give no warnings. Test cases should verify that the input to the encode function is the same as the output from the decode function. Only verifying that encode and decode did not change the floating point value more than an epsilon is not acceptable in my opinion.

I see two alternative solutions:

1. Only compile generated source files with `-Wfloat-equal`, not main.c.
2. Use `memcpy()` in main.c to convert floating point values to integers, just as done in the generated code, and then compare the integers.

I think I prefer the first alternative.

karlding:
Hmm, you're right, I suppose that is true.

Would you be opposed to using a `memcmp` instead? IMO, that would be the most explicit in comparing the actual contents of the two values, and you don't need to do a `memcpy` like your second proposal and manually perform the conversion.

```c
// ...
double signal1_result = -129.448;
assert(memcmp(&decoded.signal1, &signal1_result, sizeof(decoded.signal1)) == 0);
// ...
```

eerimoq:
Good idea, `memcmp()` shounds like the best alternative. =)

PR #83, opened 2018-12-04
Enable clang-7 builds on Travis CI

karlding:
This change enables clang builds on Travis CI.

The main reason why I think it'd be useful to also enable clang builds is:

1. Compiling with different compilers helps increase the chances of catching non-portable usage and undefined behaviour
2. Clang has a bunch of sanitizers which are very useful for development. I've enabled UBSan and ASan, which should be the most beneficial towards the goal of stamping out undefined behaviour.
3. It isn't uncommon to find projects using Clang, now that the generated code size is getting better.

I've gone through and enabled a bunch of warnings that I thought would be useful. Turns out the generated code doesn't need any additional changes for this giant list.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,10 @@ addons:
   apt:
     update: true
     sources:
+      - llvm-toolchain-trusty-7
       - ubuntu-toolchain-r-test
     packages:
+      - clang-7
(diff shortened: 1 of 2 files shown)


eerimoq:
Nice job!

PR #87, opened 2018-12-11
Remove UNUSED macro to silence MISRA diagnostic

karlding:
Remove `UNUSED` macro and `(void)` cast instead, fixing the Function-like macro advisory diagnostic.

Not sure if you want to take this change, since this is an advisory MISRA guideline. But I guess if people compile with flags to warn when macros are redefined and have their own `UNUSED` macro, then this could cause warnings? But IMO it shouldn't pollute the global namespace if it isn't absolutely necessary.

### DIFF
--- a/cantools/database/can/c_source.py
+++ b/cantools/database/can/c_source.py
@@ -268,9 +268,9 @@ ssize_t {database_name}_{message_name}_pack(
     const struct {database_name}_{message_name}_t *src_p,
     size_t size)
 {{
-    (void)(dst_p);
-    (void)(src_p);
-    (void)(size);
+    (void)dst_p;
+    (void)src_p;
+    (void)size;
 
     return (0);
 }}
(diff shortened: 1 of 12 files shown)


cantools/database/can/c_source.py
```diff
@@ -270,9 +268,9 @@
     const struct {database_name}_{message_name}_t *src_p,
     size_t size)
 {{
-    UNUSED(dst_p);
-    UNUSED(src_p);
-    UNUSED(size);
+    (void)(dst_p);
+    (void)(src_p);
+    (void)(size);
```
eerimoq: Let's make it `(void)dst_p;` instead of `(void)(dst_p);`.

eerimoq:
It would be interesting to know how many applications have ever been written with zero MISRA warnings, and if they are possible to understand by a human. Function-like macros are certainly useful, and so are using implicit casts and not having to create temporary variables just because it is considered complex to cast and assign at the same line. Also only allowing symbol names of less than 32 characters does not really build quality. The list is long... =)

That said, we should still keep the number of MISRA warnings to a minimum as people actually have to deal with them because their company says so.

juleq:
@eerimoq: You can have a look at the compliance statement for the RTOS Micrium to get a feeling. 

https://doc.micrium.com/display/osiiidoc/MISRA+C%3A2012

PR #94, opened 2019-01-10
Add support for --single-line in monitor subparser

karlding:
I found that the `--single-line` option under the decode subparser rather useful when looking at many messages, and was missing it when using the `monitor` subparser.

This implements support for the `--single-line` argument for the sake of consistency with the `decode` subparser.


eerimoq:
Released to PyPI.

PR #102, opened 2019-01-12
Replaced All Instances Of `ssize_t` with `int`

garethellis0:
This allows for compatibility with systems that do not have `uinstd.h`, like windows and arm. 

My personal motivation for this change was that my team is using these headers in an embedded ARM environment.

Please let me know if you'd like any changes made!

### DIFF
--- a/cantools/database/can/c_source.py
+++ b/cantools/database/can/c_source.py
@@ -41,7 +41,6 @@ HEADER_FMT = '''\
 
 #include <stdint.h>
 #include <stdbool.h>
-#include <unistd.h>
 
 #ifndef EINVAL
 #    define EINVAL 22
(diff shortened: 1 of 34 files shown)


eerimoq:
I'm not sure how this should be fixed. Just replacing the include with a define is not a good solution as there will be a compiler error is `ssize_t` is already defined. The easiest fix is to just replace `ssize_t` with `int`, or possible add another argument to the `generate_c_source` subcommand to control the generator behavior. Thing is, I kinda like `ssize_t` as it makes it clearer that it's a size. Not sure I want to replace it with `int`.

garethellis0:
Hmm, ok, would you mind if I moved it over to `int` now? I absolutely agree that it's not as nice of a solution, but I think that given the context in which these headers are used (across multiple types of non-posix compliant systems), it would be a good change.

eerimoq:
Change it to `int` as it seems to be the best solution.

garethellis0:
Changed!

eerimoq:
`unistd.h` is still included in the generated code.

garethellis0:
D'oh! Removed all instances of `#include <unistd.h>`

PR #104, opened 2019-01-25
Added Support For Representing Discrete Choices As Enums In Generated C Code

garethellis0:
This change should be entirely backwards compatible, it is controlled via a flag (`--generate-enums-for-choices`) in the C code generator command line utility.


tests/files/c_source/enums.h
```diff
@@ -0,0 +1,191 @@
...
+typedef enum {
+    OFF = 0,
+    NOMINAL = 1,
+    OVERHEAT = 2
+} enums_inverter_state;
+
+typedef enum {
+    OK = 0,
+    FAULTED = 1
+} enums_wheel_speed_sensor_state;
```
eerimoq: There is a high risk of name clashes without the `<database>_<message>_<signal>_`-prefix on all enums and their values. With prefixes the generated enums would look like:

``` c
enum enums_vehicle_status_inverter_state_choice_t {
    enums_vehicle_status_inverter_state_off_choice_t = 0,
    enums_vehicle_status_inverter_state_nominal_choice_t = 1,
    enums_vehicle_status_inverter_state_overheat_choice_t = 2
};

enum enums_vehicle_status_wheel_speed_sensor_state_choice_t {
    enums_vehicle_status_wheel_speed_sensor_state_ok_choice_t = 0,
    enums_vehicle_status_wheel_speed_sensor_state_faulted_choice_t = 1
};
```

In some systems only a subset of the signal values are named choices and the rest are an integer. For example, a 8 bits signal may name value 0 "NoChange" and value 1 "Error", while values 2-255 represents some voltage. I've absolutely no idea how common this is, but I've certainly seen it. Generating defines instead of enums is more flexible.

What is the reason for generating enums instead of defines?

garethellis0: - excellent point on the clashes, that's easy enough to add
- using just a subset of signals as named choices is a little trickier to address, I'm not too sure what to do about this.
- the rational behind this change is that it tightens up interfaces. When using this generated code, it will be more clear what values a given member variable in a struct should accept, which leads to more readable, understandable and debuggable code, and reduces errors that just using a raw value introduces. This change would likely propagate outwards as well, as it would allow the passing of enums (which have their meaning and restrictions inherent in the type) instead of just raw values when working with functions manipulating CAN messages.

eerimoq: The enums must have an additional value for invalid unpacked values, or something similar.

garethellis0: Hmm. At this point it's looking like this is more trouble then it's worth. I'm going to close this PR, sorry for wasting your time!

eerimoq: Ok!

PR #106, opened 2019-01-25
Added #include <stdlib.h> To Generated C Code

garethellis0:
Looks like when I removed the `uinstd` includes in #102, I inadvertantly removed the include for `size_t` as well. Here is a fix.

I sincerely apologize for the oversight on my part.


cantools/database/can/c_source.py
```diff
@@ -41,6 +41,7 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <stdlib.h>
```
eerimoq: I think we should include stddef.h instead.

https://en.wikipedia.org/wiki/C_data_types#stddef.h

garethellis0: Sounds good to me, changed!

PR #109, opened 2019-02-03
big endian start bit offset is already handled outside of kcd.py

M4GNV5:
This was changed for the dbc backend in 18b4cc31efb18f5a3f313de4f061d84b911f1123 but was not changed in the kcd backend.

### DIFF
--- a/cantools/database/can/formats/kcd.py
+++ b/cantools/database/can/formats/kcd.py
@@ -28,13 +28,6 @@ NAMESPACES = {'ns': NAMESPACE}
 ROOT_TAG = '{{{}}}NetworkDefinition'.format(NAMESPACE)
 
 
-def _start_bit(offset, byte_order):
-    if byte_order == 'big_endian':
-        return (8 * (offset // 8) + (7 - (offset % 8)))
-    else:
-        return offset
-
-
 def _get_node_name_by_id(nodes, node_id):
     for node in nodes:
         if node['id'] == node_id:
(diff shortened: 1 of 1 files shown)


M4GNV5:
I'm a little bit confused because of the tests which fail.

I assume the major problem here is the `offset` of big endian values. The kcd spec says:
> Least significant bit offset of the signal relative to the least significant bit of the messages data payload.

I'm not quite sure whats the least significant bit of the message data payload, but from my understanding the offset for big endian values is calculated using the `8 * idiv(x; 8) + 7 - mod(x; 8)` formulae.
However that would mean [the_homer.kcd](https://github.com/eerimoq/cantools/blob/master/tests/files/the_homer.kcd#L408-L410) included in this repo is wrong, as `A`, `B` and `C` collide. Maybe it should be this instead:

```xml
<Signal name="A" offset="7" length="17" endianess="big"/>
<Signal name="B" offset="22" length="1" endianess="big"/>
<Signal name="C" offset="21" length="5" endianess="big"/>
```

Which would result in 23 bits to be interpreted like this:
```
0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22
A A A A A A A A A A A  A  A  A  A  A  A  B  C  C  C  C  C
```

So maybe the test file is wrong?

eerimoq:
Is there a bug in the KCD parser, or should this be closed?

PR #110, opened 2019-02-11
DBC parsing: add support for Environment Variables

allaudet:
Implemented a new `environment_variables` _property_ inside `DbcSpecifics` _class_, which contains an ordered dictionary of all environment variables.

### DIFF
--- a/cantools/database/can/environment_variable.py
+++ b/cantools/database/can/environment_variable.py
@@ -1,0 +1,161 @@
+# A CAN environment variable.
+
+class EnvironmentVariable(object):
+    """A CAN environment variable.
+
+    """
+
+    def __init__(self,
+                 name,
+                 env_type,
+                 minimum,
+                 maximum,
+                 unit,
+                 initial_value,
+                 env_id,
+                 access_type,
+                 access_node,
+                 comment):
+        self._name = name
+        self._env_type = env_type
+        self._minimum = minimum
+        self._maximum = maximum
+        self._unit = unit
+        self._initial_value = initial_value
+        self._env_id = env_id
+        self._access_type = access_type
+        self._access_node = access_node
+        self._comment = comment
+
+    @property
+    def name(self):
+        """The environment variable name as a string.
+
+        """
+
+        return self._name
+
+    @name.setter
+    def name(self, value):
+        self._name = value
+
+    @property
+    def env_type(self):
+        """The environment variable type value.
+
+        """
+
+        return self._env_type
+
+    @env_type.setter
+    def env_type(self, value):
+        self._env_type = value
+
+    @property
+    def minimum(self):
+        """The minimum value of the environment variable.
+
+        """
+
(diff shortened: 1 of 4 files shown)


cantools/database/can/environment_variable.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+
+        return self._name
+
+    @name.setter
+    def name(self, value):
+        self._name = value
+
+    @property
+    def type(self):
```
eerimoq: Preferably replace ``type`` and ``id`` with other words to make syntax highlighters happy. Maybe ``kind`` instead of ``type``, or ``env_type``? It's up to you, I don't know what's best.

cantools/database/can/formats/dbc.py
```diff
@@ -315,13 +316,15 @@ class DbcSpecifics(object):
...
                  attribute_definitions=None,
+                 environment_variables=None,
                  value_tables=None):
         self._attributes = attributes
         self._attribute_definitions = attribute_definitions
 
         if value_tables is None:
             value_tables = odict()
 
+        self._environment_variables = environment_variables
```
eerimoq: Move to line 323, before all ``value_tables`` code.

cantools/database/can/environment_variable.py
```diff
@@ -0,0 +1,161 @@
...
+
+        return self._unit
+
+    @unit.setter
+    def unit(self, value):
+        self._unit = value
+
+    @property
+    def initial_value(self):
+        """The environment variable initial_value as a string.
```
eerimoq: This is not a string according to the test case. Same goes for minimum, maximum, and other members.

cantools/database/can/environment_variable.py
allaudet: True, I think `env_id` would be the best to deal with `id`, so `env_type` by extension.

cantools/database/can/formats/dbc.py
allaudet: No problem

cantools/database/can/environment_variable.py
allaudet: done

eerimoq: I meant the docstring was wrong. The code itself was ok. There is no need to call ``num`` or ``int`` anywhere in this file. At least they are not called in the similar classes ``AttributeDefinition`` or ``Attribute``.

allaudet: Yes, the docstring is already changed.
About the `int`and `num` calls, is it not the class's responsibility to ensure the integrity of the values it houses?

eerimoq: Let's make this class consistent with other classes for now, and maybe change all classes at once later.

allaudet: Right, then, so be it :+1:

PR #112, opened 2019-02-26
Adding signal initial value from dbc attribute

IvanEscamilla:
Sets the initial value to the signal when parsing a dbc file.


eerimoq:
Please update an existing test case or add another testing the new functionality. Other than that it looks good.

eerimoq:
Close in lack of response.

IvanEscamilla:
Sorry for the late response, I was too busy past days. I updated the existing test case and added new ones.

eerimoq:
Did you forget to push the changes? There are no tests in this PR.

IvanEscamilla:
Yes, I forgot to add the tests on this PR.

eerimoq:
Let me know when this is ready for review.

IvanEscamilla:
I think is ready I just don't know why is failing on test_generate_c_source case

eerimoq:
You just have to revert some changes in vehicle.dbc.

IvanEscamilla:
Ready for review

tests/test_database.py
```diff
@@ -66,6 +67,27 @@ def test_vehicle(self):
...
+        self.assertEqual(db.get_message_by_name('RT_IMU06_Accel').get_signal_by_name('Validity_Accel_Vertical').initial , 0)
+        self.assertEqual(db.get_message_by_name('RT_IMU06_Accel').get_signal_by_name('Accuracy_Accel').initial , 127)
+        self.assertEqual(db.get_message_by_name('RT_IMU06_Accel').get_signal_by_name('Accel_Longitudinal').initial , 32767)
+        self.assertEqual(repr(db.get_message_by_name('RT_IMU06_Accel').get_signal_by_name('Accel_Longitudinal')),
+                         "signal('Accel_Longitudinal', 16, 16, 'little_endian', True, 32767, "
+                         "0.001, 0, -65, 65, 'g', False, None, None, "
+                         "'Longitudinal acceleration.  This is positive when the vehicle "
+                         "accelerates in a forwards direction.')")
+        self.assertEqual(db.get_message_by_name('RT_IMU06_Accel').get_signal_by_name('Accel_Lateral').initial , -30000)
+        self.assertEqual(db.get_message_by_name('RT_IMU06_Accel').get_signal_by_name('Accel_Vertical').initial , 16120)
```
eerimoq: Try to keep the line length below ~80 characters.

cantools/database/can/formats/dbc.py
```diff
@@ -1021,6 +1021,14 @@ def get_signal_name(frame_id_dbc, name):
...
             return name
 
+    def get_signal_initial_value(frame_id_dbc, name):
+        signal_attributes = get_attributes(frame_id_dbc, name)
+
+        try:
+            return signal_attributes['GenSigStartValue'].value
+        except (KeyError, TypeError):
+            return None
+
```
eerimoq: Are you intentionally ignoring the default attribute value?

IvanEscamilla: you are right I am not passing the default attribute value, let me correct it

tests/test_database.py
IvanEscamilla: Ok I will do it

cantools/database/can/formats/dbc.py
IvanEscamilla: I was looking at the Documentation and if the signal has no initial value a None should be returned, that is why I was returning the None if the signal has no GenSigStartValue attribute defined in the DBC file,  the default attribute value that you are referring is the default value from the DBC file?

eerimoq: I'm not really sure myself how this should work. But yes, I was referring to the default value of the GenSigStartValue attribute in the DBC-file. I was just curious if the default value should be used or not, if you had even considered it.

I'm fine with only adding initial values to signals that have a GenSigStartValue attribute in the DBC-file, and give all other signals the initial value ``None``, just as you already implemented. We can change the behavior later if necessary.

IvanEscamilla: Yes I considered it, actually I think is not difficult to add that implementation, but I think it can be added later if it is really necessary, right now I am also good with this approach.

PR #114, opened 2019-03-04
Generate C++11 Code and do not clear all message data on unpack

walmis:
1. C++11 Code support, uses Linux can_frame structure to decode data, usage example:

 ```
can_frame rx_frame;
PDM_FRONT::FR_REQ message;
message.unpack(frame);

//then to access raw unpacked values:

print( message.raw.en_left_turn_indicator );

// or decode

print( message.en_left_turn_indicator() );

packing and setting of values is similar.

It also exports "choice" to string convesion and units as strings. It's a header only export and it links to the generated C code.

```

2. Instead of memsetting whole message data to 0, generated code now clears values on demand, for example:
```    
    dst_p->ucmin_ext = 0;
    dst_p->ucmin_ext |= unpack_right_shift_u16(src_p[0], 0u, 0xffu);
    dst_p->ucmin_ext |= unpack_left_shift_u16(src_p[1], 8u, 0xffu);
    dst_p->ucmax_ext = 0;
    dst_p->ucmax_ext |= unpack_right_shift_u16(src_p[2], 0u, 0xffu);
    dst_p->ucmax_ext |= unpack_left_shift_u16(src_p[3], 8u, 0xffu);
    dst_p->nucmin = 0;
    dst_p->nucmin |= unpack_right_shift_u16(src_p[4], 0u, 0xffu);
    dst_p->nucmin |= unpack_left_shift_u16(src_p[5], 8u, 0xffu);
    dst_p->nucmax = 0;
    dst_p->nucmax |= unpack_right_shift_u16(src_p[6], 0u, 0xffu);
    dst_p->nucmax |= unpack_left_shift_u16(src_p[7], 8u, 0xffu);```


eerimoq:
What's the reason for removing memset? Without it parts of the message data is potentially uninitialized when pack returns, and must instead be initialized by the user before calling the pack function.

walmis:
The reason I removed memset is that I need multiplexed data to remain on multiplexed messages. For example in mutiplexed message with parts A and B. when B data arrives, A data is zeroed out.

eerimoq:
Can't you pack it all at once?

walmis:
> Can't you pack it all at once?
This is for unpacking. For packing the memset is as it was before.

eerimoq:
Ok, I still think it's bad habit to return uninitialized data structures. Maybe generate a global structure of all signals in the database instead, or at least per message?

eerimoq:
How should we proceed with this PR? =)

Issue #103, opened 2019-01-13
Multiplexed messages in monitor output


karlding:
Currently, when using multiplexed messages, they get decoded under the same message like so:

```
Received: 104, Discarded: 49, Errors: 0
   TIMESTAMP  MESSAGE
       2.750  BatteryVoltageTemperature(
                  BatteryIndex_mux: 28,
                  Voltage_28: 39294,
                  Temperature_28: 23122
              )
```

When you have multiplexed messages that come at high frequency, this is somewhat hard to parse, as it updates too quickly.

I think it would be useful to either:

1. Display all the muxed values under separate messages
2. Display all the muxed values under the same message

`(1)` would look something like:

```
Received: 104, Discarded: 49, Errors: 0
   TIMESTAMP  MESSAGE
       2.750  BatteryVoltageTemperature(
                  BatteryIndex_mux: 28,
                  Voltage_28: 39294,
                  Temperature_28: 23122
              )
       
       2.754  BatteryVoltageTemperature(
                  BatteryIndex_mux: 29,
                  Voltage_29: 39477,
                  Temperature_29: 23022
              )
```

And `(2)` would look something like:

```
Received: 104, Discarded: 49, Errors: 0
   TIMESTAMP  MESSAGE
       2.754  BatteryVoltageTemperature(
                  BatteryIndex_mux: 29,
                  Voltage_28: 39294,
                  Voltage_29: 39477,
                  Temperature_28: 23122
                  Temperature_29: 23022
              )
```

The multiplexer value would denote the last MUX'd value received (or it could optionally be hidden).

I guess this would also be applicable to the `decode` subparser as well.

What are your thoughts?

eerimoq:
In some systems it's irrelevant which multiplexer value (or even message) the signal is part of. For these systems a union of all signals for all multiplexer values would make sense. One could even argue that the message should not be displayed at all, but only the union of all signals in the system.

As I understand it you are using the multiplexer value as an identifier of the source of the voltage and temperature, so a union of all signals in the message wont work as only the latest will be displayed. Am I missing something here?

juleq:
@karlding: So you‘re in the BMS club, heh? Cheers :).

For my telemetry, I do it like @eerimoq suggests: I have update listeners attached to all my signals. On update, a signal value is copied to a signal buffer (union).

The application samples the signal buffer. 

Further, each signal has an invalidation timer attached to it. If a signal has not been updated for e.g. three consecutive times, it is considered frozen and invalidated.

[...]

PR #118, opened 2019-03-09
Separate MUX'd signals in monitor subparser output

karlding:
Totally forgot I had this patch lying around, after getting caught up
with school again. I cleaned it up, so might as well toss it up for an
initial review.

I'm not sure if this is the right general approach, but I basically
decided that it would be ideal not to modify the `Signal` and `Message`
classes, since this is purely presentation (and those are exposed).

Currently this doesn't affect filtering, which is a plus in my books.
This still displays the message name as the original name, instead of
appending the MUX identifier.

### DIFF
--- a/cantools/subparsers/monitor.py
+++ b/cantools/subparsers/monitor.py
@@ -14,6 +14,7 @@ except ImportError:
 import can
 from .. import database
 from .utils import format_message
+from .utils import format_multiplexed_name
 
 
 class QuitError(Exception):
@@ -226,9 +227,14 @@ class Monitor(can.Listener):
             self._discarded += 1
             return
 
+        name = message.name
+        if message.is_multiplexed():
+            name = format_multiplexed_name(message, data, True)
+
         if self._single_line:
             formatted = format_message(message, data, True, True)
-            self._formatted_messages[message.name] = [
+
+            self._formatted_messages[name] = [
                 '{:12.3f} {}'.format(timestamp, formatted)
             ]
         else:
@@ -236,10 +242,10 @@ class Monitor(can.Listener):
             lines = formatted.splitlines()
             formatted = ['{:12.3f}  {}'.format(timestamp, lines[1])]
             formatted += [14 * ' ' + line for line in lines[2:]]
-            self._formatted_messages[message.name] = formatted
+            self._formatted_messages[name] = formatted
 
-        if message.name not in self._filtered_sorted_message_names:
-            self.insort_filtered(message.name)
+        if name not in self._filtered_sorted_message_names:
+            self.insort_filtered(name)
 
     def update_messages(self):
         modified = False
--- a/cantools/subparsers/utils.py
+++ b/cantools/subparsers/utils.py
@@ -17,8 +17,10 @@ def _format_signals(message, decoded_signals):
         if isinstance(value, str):
             value = "'{}'".format(value)
 
+        signal_name = signal.name
+
         formatted_signals.append(
-            '{}: {}{}'.format(signal.name,
+            '{}: {}{}'.format(signal_name,
                                value,
                               ''
                               if signal.unit is None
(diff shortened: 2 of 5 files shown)


eerimoq:
I can't make up my mind whether this is something generally useful, or just solving the special use case when the mux-signal is used as a data identifier. Also, there may be multiple mux-signals in a single message (extended multiplexing), which the logic in this patch does not handle gracefully.

I'm tempted to merge this PR anyway as it shouldn't make things worse for any user, but there is a risk that this behavior will change in the future.

karlding:
Hmm, that's probably fair.

Out of curiosity, how do other tools handle displaying the extended multiplexing case? Maybe it makes sense to see what PEAK PCAN-Explorer/Vector CANalyzer/Kvaser CANKing does and do something similar?

I have never needed to use the extended multiplexer case, but I vaguely recall that for a single mux signal, PCAN-Explorer separates them onto their own rows. It's probably a good idea to check and document the behaviour if the idea is to go that route.

Do you want to move this discussion to the original issue I raised (#103)?

eerimoq:
Wait a minute. I just realized I misunderstood how your muxed message looks. I was under the impression that the message only had three signals, mux, voltage and temperature, but your test code clearly shows that there is one mux signal and a pair of voltage and temperature signals per mux index. 

We should implement alternative ``(2)`` instad of  ``(1)`` as described in the original issue. That is, just display a union of all received signals in the message.

karlding:
Just to clarify, should old values be retained as muxed signals are received?

For example, it would start out with this:

```
BATTERY_VT(
    BATTERY_VT_INDEX: 0,
    MODULE_VOLTAGE_00: 39064,
    MODULE_TEMP_00: 11,
    MODULE_VOLTAGE_01: 0,
    MODULE_TEMP_01: 0,
    MODULE_VOLTAGE_02: 0,
    MODULE_TEMP_02: 0,
    MODULE_VOLTAGE_03: 0,
    MODULE_TEMP_03: 0,
    MODULE_VOLTAGE_04: 0,
    MODULE_TEMP_04: 0,
    MODULE_VOLTAGE_05: 0,
    MODULE_TEMP_05: 0,
    MODULE_VOLTAGE_06: 0,
    MODULE_TEMP_06: 0,
    MODULE_VOLTAGE_07: 0,
    MODULE_TEMP_07: 0,
    MODULE_VOLTAGE_08: 0,
    MODULE_TEMP_08: 0,
    // ...
    MODULE_VOLTAGE_35: 0, 
    MODULE_TEMP_35: 0
)
```

and then upon receiving the next MUX signal

```
BATTERY_VT(
    BATTERY_VT_INDEX: 1, <--- change
    MODULE_VOLTAGE_00: 39064,
    MODULE_TEMP_00: 11,
    MODULE_VOLTAGE_01: 39068, <--- change
    MODULE_TEMP_01: 10, <--- change
    MODULE_VOLTAGE_02: 0,
    MODULE_TEMP_02: 0,
    MODULE_VOLTAGE_03: 0,
    MODULE_TEMP_03: 0,
    MODULE_VOLTAGE_04: 0,
    MODULE_TEMP_04: 0,
    MODULE_VOLTAGE_05: 0,
    MODULE_TEMP_05: 0,
    MODULE_VOLTAGE_06: 0,
    MODULE_TEMP_06: 0,
    MODULE_VOLTAGE_07: 0,
    MODULE_TEMP_07: 0,
    MODULE_VOLTAGE_08: 0,
    MODULE_TEMP_08: 0,
    // ...
    MODULE_VOLTAGE_35: 0, 
    MODULE_TEMP_35: 0
)
```

eerimoq:
Yeah, retaining already received signals was the idea, but I realize that the message timestamp will only match the most recently received signals in such an implementation. We discussed this in the original issue  and it was a reason for implementing ``(1)``. As timestamps are often important to the user it might be better to implement ``(1)`` after all. I think I'll leave it for you to decide, I'm confused at the moment.

Let me know when you are satisfied with the PR and I'll merge. =)

karlding:
What if we just enumerated all the possible mux values and had each as a distinct "Message"? This would handle the extended multiplexing case, and also keeps the timestamp.

Here's how PCAN-Explorer handles the single multiplexing case:

pcan-explorer

eerimoq:
Sounds like a good idea! There is a theoretical risk of running out of memory if the message has many mux-signals, but I think that's not a problem in practice.

Another alternative is to implement ``(2)`` and retain signal values, and add a timestamp per signal in the timestamp column when ``--single-line`` is not given. I suggest that we start by implementing your idea, and make improvements when required.

karlding:
@eerimoq I pushed my changes, let me know if you have any concerns!

eerimoq:
I'll have have a look next week.

Issue #120, opened 2019-03-11
access to signals on the can bus


sgnes:
Hello,

I'm looking for a tool to access to the signals on the can bus, like get_signal_value_phys(message name, signal name), and after check the document for cantools, I don't find this kind of interface.

Would you please help check this point?
If we don't have ,shall we have a class like "Device" to hold the signals values, and user can get it easily?
Thanks.

eerimoq:
You're right, there is no such API. Feel free to design it and submit a PR.

sgnes:
@eerimoq Thanks for your reply, I will try to create a PR soon.

PR #121, opened 2019-03-12
Dev add device class

sgnes:
Hello,

Please check this change, ref to issue #120 .

### DIFF
--- a/cantools/database/can/database.py
+++ b/cantools/database/can/database.py
@@ -338,6 +338,8 @@ class Database(object):
         If `strict` is ``True`` all signal values must be within their
         allowed ranges, or an exception is raised.
 
+        if this frame id is not belong to this database, will return None.
+
         >>> db.encode_message(158, {'Bar': 1, 'Fum': 5.0})
         b'\\x01\\x45\\x23\\x00\\x11'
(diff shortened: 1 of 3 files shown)


sgnes:
I will change it later to remove the change for message class, also I will create test case later .

sgnes:
@eerimoq, would you please tell me why my new added device.py has 0% coverage, I tested it locally, and I have 100%.

eerimoq:
Hello!

Thanks for the effort, but I'm not really sure whether this is useful or not. What is you use case? Can you give an example of when this functionality is needed? I just want to make sure not to add too much code that is seldom (or never) used. There is already the tester class that probably no one uses.

sgnes:
For some test , we need to know the signal vlaue on the can bus, and also send aspecific signal  value to check the action of the tested ecu. With this interface it will be easier.
For now this works with normal message, but as no tp is used, not working with message bigger than 8bytes.

eerimoq:
Did you try using the [tester class](https://github.com/eerimoq/cantools/blob/master/cantools/tester.py)?

sgnes:
Seems has the similar functionality, i will try it.

PR #125, opened 2019-03-25
Fix deprecation warnings for SocketCAN interfaces in examples

karlding:
The `socketcan_native` and `socketcan_ctypes` interfaces have been
deprecated in `python-can` as of `Version 2.2.0 (2018-06-30)`, and are
superseded by the `socketcan` interface. All the examples have been
updated to updated to use the `socketcan` interface instead.

**Note**: As of `2.2.0`, `python-can` dropped support for Python 3.3, so with
the update in `setup.py`, this implicitly drops support for that as well.

### DIFF
--- a/cantools/tester.py
+++ b/cantools/tester.py
@@ -223,7 +223,7 @@ class Tester(object):
 
     >>> import can
     >>> import cantools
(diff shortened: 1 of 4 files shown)


eerimoq:
Thank you!

PR #129, opened 2019-03-28
Fix test dbc files to include BU_ section

msalau:
BU_ section is mandatory. CANdb++ is not able to open a file without such section.


eerimoq:
Thanks!

Issue #127, opened 2019-03-28
cantools.database.can.bus.py only used for .kcd parsing (need .dbc support)


anxuae:
Hi,

Currently the database created from a .dbc file is always created sith an empty buses list.
See https://github.com/eerimoq/cantools/blob/4bde521fb35787776092d6ec31dc1356e4b3c35a/cantools/database/can/formats/dbc.py#L1352-L1356

However, information about the bus can be found in the DBC.  For instance in the one I'm using:

*Bus name:*

```
BA_DEF_  "DBName" STRING ;
BA_DEF_DEF_  "DBName" "";
BA_ "DBName" "Infotainment_AmplifierCAN";
```

*Bus baudrate:*

```
BA_DEF_  "Bauderate" INT 0 1000000;
BA_DEF_DEF_  "Bauderate" 125000;
BA_ "Bauderate" 125000;
```

Is it possible to add at least a `Bus()` in the list with the following info?

eerimoq:
Feel free to implement it. I'll likely merge it =)

anxuae:
Hi eerimoq,
Do you think the my pull request can be integrated?
Thanks
Anxuae

eerimoq:
Sorry, I've been working on another project lately and forgot about your PR. I'll have a look now.

eerimoq:
I added tests and refactored a bit.

There was a comment that the default baudrate was copied from kcd.py. I changed the default baudrate to None, or unavailable. Let me know if it should be 500000 instead and I'll change it back.

anxuae:
Thanks a lot.
No matter of the default value, I did it only to be consistent with KCD parser.

PR #132, opened 2019-04-02
Add support for Bus name and baudrate with .dbc files 

anxuae:
cantools.database.can.bus.py only used for .kcd parsing (need .dbc support) #127


anxuae:
Can not ensure the same coverage of:

``
if len(bus_names) == 0:
    if bus_name is not None: | !
        raise Error( "expected bus name None as there are no buses defined in "
                           "the database, but got '{}'".format(bus_name))
``

because now there is always at least one bus even with a dbc file

PR #141, opened 2019-05-07
Add support for pagination in monitor subparser

karlding:
Pagination is mapped to <kbd>PgDn</kbd> (Next Page) and <kbd>PgUp</kbd> (Prev. Page),
allowing content to overflow the visible screen and still be rendered on
a new page.

### DIFF
--- a/cantools/subparsers/monitor.py
+++ b/cantools/subparsers/monitor.py
@@ -42,7 +42,9 @@ class Monitor(can.Listener):
         self._received = 0
         self._discarded = 0
         self._basetime = None
+        self._page = 0
 
+        stdscr.keypad(True)
         stdscr.nodelay(True)
         curses.use_default_colors()
         curses.curs_set(False)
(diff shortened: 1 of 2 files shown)


cantools/subparsers/monitor.py
```diff
@@ -174,13 +184,23 @@ def process_user_input_menu(self, key):
...
             self._modified = True
             curses.curs_set(True)
+        elif key in ['KEY_PPAGE']:
+            # Decrement page
+            if self._page > 0:
+                self._page -= 1
+            self._modified = True
+        elif key in ['KEY_NPAGE']:
+            # Increment page
+            self._page += 1
```
eerimoq: Should we add an upper limit here?

karlding: I did some thinking about this, and if we want an upper limit, we need to know the maximum number of messages we could possibly decode with this DBC. This seems doable, since the DBC specifies all the messages it would display.

However, is there a way to get the number of multiplexer combinations that we have in a DBC?

If we don't have multiplexers, then we can simply use `cantools.database.can.Database.messages` and the length of that is the total possible lines that list is the maximum number of lines we would need to render.

But with multiplexing, we're rendering by taking each multiplexer and building out each possible combination via all the values the multiplexer can take on. As such, there isn't a 1:1 mapping between the number of messages in a DBC and the number of potential lines that need to be rendered. Is the only way to do this calculation by parsing the signal tree for each multiplexed message and building that result? Not sure if I missed some existing API that would save some work.

eerimoq: Let's leave it for now, just as you pointed out it's probably too much work in relation to the benefit for now.

PR #146, opened 2019-06-19
Fix crash when MUX index is not specified in DBC

karlding:
Fix a crash when the MUX index is not specified inside the DBC file. In
this case, we previously would crash with the following error:

    IndexError: list index out of range

This would happen because the MUX decoding would fail and raise an error
when an invalid MUX index was passed, but the DecodeError would be
caught and then serialized into a string, leading to the IndexError
when the code later attempted to index into it.

This change propagates the error to the monitor subparser's update
function. Here, we handle the invalid MUX index the same way as if we
received an unexpected CAN message that wasn't specified in the DBC, and
discard the message.

### DIFF
--- a/cantools/subparsers/monitor.py
+++ b/cantools/subparsers/monitor.py
@@ -253,7 +253,15 @@ class Monitor(can.Listener):
         name = message.name
 
         if message.is_multiplexed():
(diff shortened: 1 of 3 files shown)


eerimoq:
Part of version 32.19.1, available on PyPI.

PR #168, opened 2019-11-01
File ext ignore case

### LINKED ISSUE #165: load database requires lower case file extension
The following command throws an exception:
`dbc = cantools.database.load_file("dummy.DBC")`

...
`ValueError: expected database format 'arxml', 'dbc', 'kcd', 'sym', 'cdd' or None, but got 'DBC'`

When calling the method with lower case file extension (.dbc) everything is fine.
It seems to be irrelevant if the file's extension itself uses upper or lower case, it's just the "filter" in the load_file method.

Hint: For manual use it's no problem to use lower case. But searching for files with glob.glob(...) could return mixed results which then have to be enforced to lower case names.

geo4711:
Closes #165 --> accept database file extension both upper case/lower case/any combination

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,3 @@
 include LICENSE
 include Makefile
-recursive-include tests *.py *.arxml *.dbc *.cdd *.kcd *.sym *.h *.c *.mk
+recursive-include tests *.py *.arxml *.dbc *.cdd *.kcd *.sym *.h *.c *.mk *.DBC
--- a/cantools/database/__init__.py
+++ b/cantools/database/__init__.py
@@ -275,6 +275,8 @@ def load_string(string,
 
     """
 
+    # Accept both lower case/upper case/any combination for importing file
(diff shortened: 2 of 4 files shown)


eerimoq:
Lot's of tests are failing. Please investigate why and take action based on that.

eerimoq:
Please add a test case and I can merge.

geo4711:
thanks a lot for your patience and advice :-)

eerimoq:
The documentation here should be updated to match the current implementation. I suggest to write that both lower and upper case file extensions are allowed instead of adding upper case extensions to the table, but it's up to you. Mixing upper and lower case in the same extension is probably not worth mentioning.

https://github.com/eerimoq/cantools/blob/master/cantools/database/__init__.py#L106-L124

eerimoq:
Add *.DBC to MANIFEST.in. Otherwise the new file is not part of the release and the test using it will fail.

PR #169, opened 2019-11-01
Update dbc.py

### LINKED ISSUE #163: Double newline in dbc dump on Windows
I use cantools on a Windows machine.
I try to apply specific patches to given dbc files (which are created on Windows, too).

The input files are coded with CR+LF as a newline.
The output files, however, show kind of doubles newlines: CR+CR+LF

It's not a big issue, afaik other tools have no problems reading those files, but it doesn't look very nice in text mode.

geo4711:
Close #163: no more "doubled" newlines in generated dbc

### DIFF
--- a/cantools/database/can/formats/dbc.py
+++ b/cantools/database/can/formats/dbc.py
@@ -33,57 +33,57 @@ from .utils import num
 
 
 DBC_FMT = (
-    'VERSION "{version}"\r\n'
-    '\r\n'
-    '\r\n'
-    'NS_ : \r\n'
-    '\tNS_DESC_\r\n'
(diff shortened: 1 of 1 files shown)


geo4711:
I've had a look at the test results. The problem is, that the test cases expect `\r\n` newlines. To be in accourdance with the python recommendation I would suggest to remove the `\r` from the testcases. What do you think about it?
Or would you prefer a different solution? Maybe an extra (optional) parameter for the `fopen()` for the newline (default=None --> current behaviour)? But that would mean to add this parameter to some other functions, too.

geo4711:
Agreed on different solution, see comments on #163

PR #176, opened 2019-12-14
Extern c

nickolai-voyage:
Following on from my failed attempt to contribute the removal of useless casts, I have what is hopefully a more useful PR :D

This will make it more straightforward to use these files in C++ projects. Because of C++ compiler name mangling, trying to use these files could result in a bunch of `undefined reference` errors.

I tried adding a C++ test, but it gave a lot of warnings about conversion, which should probably be fixed, but are outside the scope of this PR and so I think for testing it's enough that C compilation is tested, and there are python tests comparing the output of the code generator to the files stored in the repo.


eerimoq:
It is already straight forward to use the generated code from C++, just wrap the include in an extern C block as below. I'll probably merge this anyway though, to make lazy C++-programmers happy. =)

``` c++
extern "C" {
#include "foo.h"
}
```

nickolai-voyage:
I know, but the reason I added this is because if someone is inexperienced with mixing C and C++, they won't know about this and they'll be staring down linker errors like `undefined reference to abs_bremse_33_pack` and scratching their heads as to how it could be undefined when the library is right there on the link line. Even for an experienced programmer like me, it took me a second to remember that I needed to add extern "C".

Plus it's nice for C++ support to work out of the box, you can add "C++ support" to the feature list!

eerimoq:
Ohh well, I'll merge =)

eerimoq:
Part of release 32.21.0, available on PyPI.

PR #180, opened 2020-01-20
Misprint fix in docs URL

manul7:


eerimoq:
I had the same issue in another project, and solved it by changing the link to the form `http://cantools.readthedocs.org`. I suggest that we do the same here.

manul7:
@eerimoq seems that the root cause of broken link issue is missing "/" at the end of URL, but as link to "readthedocs.org" will be redirected to "readthedocs.io" I decided to change top-level domain name too. If it's unacceptable, I'll leave ".org".

eerimoq:
Ok, let's change to `.io`.

eerimoq:
I updated the link, thanks!

PR #182, opened 2020-01-24
Bugfix for dumping dbc with objects whose names have been shortened

geo4711:
Objects (messages, signals, nodes) with an original name >32 chars and a modified name <= 32 chars are now dumped correctly.


eerimoq:
Released to PyPI in version 33.1.1.

PR #185, opened 2020-02-11
make it work with AUTOSAR 3 ARXML files

andlaus:
This patch implements support for the AUTOSAR 3 CAN bus specification. (This is the format used for older models within our company.) While I cannot guarantee that the spec is 100% satisfied, it seems work fine for the files which are floating around internally here.

---

For formal reasons, the following legalese needs to be included (sorry for that):

The changes contained in this pull request are contributed under the terms of the MIT License by Andreas Lauser <Andreas.Lauser@daimler.com>, Daimler AG acting on behalf of MBition GmbH.

https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md

The program was tested solely for our own use cases, which might differ from yours.

Daimler AG
Mercedesstraße 137
70327 Stuttgart, Germany
Phone.: +49711170
E-Mail: dialog@daimler.com

Sitz und Registergericht/Domicile and Court of Registry: Stuttgart, HRB-Nr./Commercial Register No.: 19360
Vorstand/Board of Management: Ola Källenius (Chairman), Martin Daum, Renata Jungo Brüngger, Wilfried Porth, Markus Schäfer, Britta Seeger, Hubertus Troska, Harald Wilhelm
Vorsitzender des Aufsichtsrats/Chairman of the Supervisory Board: Manfred Bischoff

https://www.daimler.com/provider/

### DIFF
--- a/cantools/database/can/formats/arxml.py
+++ b/cantools/database/can/formats/arxml.py
@@ -14,29 +14,45 @@ from ..internal_database import InternalDatabase
 LOGGER = logging.getLogger(__name__)
 
 # The ARXML XML namespace.
-NAMESPACE = 'http://autosar.org/schema/r4.0'
-NAMESPACES = {'ns': NAMESPACE}
+AUTOSAR3_NAMESPACE_CANDIDATES = [ 'http://autosar.org/schema/r3.0', 'http://autosar.org/3.1.4.DAI.4' ]
+AUTOSAR4_NAMESPACE_CANDIDATES = [ 'http://autosar.org/schema/r4.0' ]
 
-ROOT_TAG = '{{{}}}AUTOSAR'.format(NAMESPACE)
-
+def get_root_tag(namespace):
+    return '{{{}}}AUTOSAR'.format(namespace)
 
 def make_xpath(location):
     return './ns:' + '/ns:'.join(location)
 
 
 # ARXML XPATHs.
-CAN_FRAME_TRIGGERINGS_XPATH = make_xpath([
-    'AR-PACKAGES',
-    'AR-PACKAGE',
-    'ELEMENTS',
-    'CAN-CLUSTER',
-    'CAN-CLUSTER-VARIANTS',
-    'CAN-CLUSTER-CONDITIONAL',
-    'PHYSICAL-CHANNELS',
-    'CAN-PHYSICAL-CHANNEL',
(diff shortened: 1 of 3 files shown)


eerimoq:
Thanks, but you don't happen to have an ARXML 3 file that can be used to test the new functionality? Without a test the new functionality will break in the future, and I don't want that.

andlaus:
okay, I see what I can do to provide a publishable file. It will probably take a while and I'll likely need a hand to get it integrated into the test suite, though.

Another thing which I noticed while working on this is that the handling of XML namespaces in the ARXML loader currently lets a bit to be desired: It only works for AUTOSAR 4.0, but not for different minor or patch versions: e.g., it loads AUTOSAR 4.0 files but not AUTOSAR 4.0.1 ones. Is this a conscious decision? If not, I'll see what I can do on that front as well...

eerimoq:
There is no rush. I'll merge once a test is in place. =)

I was not aware of the minor or patch versions problem. Any help to fix it is very welcome.

cantools/database/can/formats/arxml.py
```diff
@@ -937,17 +976,36 @@ def load_string(string, strict=True):
     """Parse given ARXML format string.
 
     """
+    global NAMESPACE, NAMESPACES
```
eerimoq: Is there a way to do this without globals?

andlaus: Sorry for the belated answer, for whatever reason I did not see the notification.

> Is there a way to do this without globals?

Yes. I just wanted to do this with as few changes to the code as possible. I fully agree that these variables belong into the  `SystemLoader` class, though, so I'll move them into it in the next revision of the PR. (I'm still working on it, BTW. It's just a quite a bit more involved than I originally thought.)

eerimoq: Ok!

eerimoq: Any progress?

andlaus:
This took quite a bit longer than I expected (don't ask). I discovered that this PR was quite incomplete, so I refactored it considerably and rebased on top of the current master. the result is in [this](https://github.com/Daimler/cantools/tree/arxml_refactor) branch, which I intend to merge via a series of handy and easy to review PRs. the first of this series is #236.

PR #186, opened 2020-02-11
fix SIG_VALTYPE_ item on DBC dump for extended frame id messages

timheb:
The frame id in the SIG_VALTYPE_ value is wrong when writing the dbc as string if extended frame is valid


eerimoq:
Thanks! Please add a test and I will merge.

eerimoq:
Any progress?

PR #207, opened 2020-06-29
Dbc allow plus in sig def

geo4711:
Some DBC generators use an extra '+' for defining the ranges of signals. This PR allows/tolerates this extra char.


eerimoq:
Thanks. Looks good. Will merge once a test is in place.

geo4711:
Sure. Just added a test and a separate DBC for that. But I'm still wondering how that change leads to reduced coverage ...

eerimoq:
It's probably a bug in coveralls, so it's ok.

eerimoq:
Now available on PyPI.

PR #210, opened 2020-07-10
Fix: comment handling for database in dbc format

malneni:
**Updated dbc.py** to handle comment for Bus.
For comment Sequence after CM_ is choice. ex: CM_ "Special Release";

Test: 
**Added comment to vehicle.dbc.**,  Existing tests to compare InDBC = OutDBC is enough for testing.

### DIFF
--- a/cantools/database/can/formats/dbc.py
+++ b/cantools/database/can/formats/dbc.py
@@ -1382,7 +1382,7 @@ def _load_bus(attributes, comments):
     try:
         bus_name = attributes['database']['DBName'].value
     except KeyError:
-        bus_name = None
+        return None
 
     try:
         bus_baudrate = attributes['database']['Baudrate'].value
--- a/cantools/database/can/formats/dbc.py
+++ b/cantools/database/can/formats/dbc.py
@@ -350,8 +350,7 @@ class DbcSpecifics(object):
                  attributes=None,
                  attribute_definitions=None,
                  environment_variables=None,
-                 value_tables=None,
-                 comment=None):
+                 value_tables=None):
         if attributes is None:
             attributes = odict()
 
@@ -368,7 +367,6 @@ class DbcSpecifics(object):
         self._attribute_definitions = attribute_definitions
         self._environment_variables = environment_variables
         self._value_tables = value_tables
-        self._comment = comment
 
     @property
     def attributes(self):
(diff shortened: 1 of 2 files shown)


cantools/database/can/formats/dbc.py
```diff
@@ -381,6 +383,18 @@ def attributes(self):
...
+        """The DBC specific comment as a string.
+
+        """
+
+        return self._comment
+
+    @comment.setter
+    def comment(self, value):
+        self._comment = value
+
```
eerimoq, line 397: What's the reason for adding the comment attribute to the DbcSpecifics class? Isn't it better to use the comment attribute on the Bus class?

cantools/database/can/formats/dbc.py
```diff
@@ -558,6 +572,13 @@ def _dump_senders(database):
 def _dump_comments(database):
     cm = []
 
+    if database.dbc is not None:
+        if database.dbc.comment is not None:
+            cm.append(
+                'CM_ "{comment}";'.format(
+                    #comment=database.dbc.comment.replace('"', '\\"')))
+                    comment=database.dbc.comment) )
```
eerimoq, line 580: Remove commented code.

malneni: 1. Tests: OK, I will add more tests.
2. I do not have specific reason between DbcSpecifics class vs Bus class. It is tricky to choose
I tried to duplicate comments and see how DBC viewer tool handles like
CM_ "Special Release";
CM_ "Special Release2";
tool combines them like: CM_ "Special ReleaseSpecial Release2";
& Bus class handles only 3 attributes, if  there is no Bus name and baudrate attribute in dbc file. comment  is only attribute left 

DbcSpecifics class handles attributes dictionary so looks more appropriate for me personally

3.  I will remove commented code

eerimoq: Use the comment attribute in the Bus class. Do not add anything to the DbcSpecifiers class. The reason is that a comment seems generic over multiple database formats, and therefore belongs to the Bus class.

eerimoq: It looks like only the last found CM_ for signals and messages are stored. I suggest using the same logic here. That is, not concatenating comments.

malneni:
I am redoing the update on latest release

PR #212, opened 2020-07-21
Enforce python 3.6 in setup.py

superjax:
This PR adds some checks and an error message to `setup.py` to help users identify packaging problems if they are using Ubuntu Xenial (defaults to 3.5.2).

Without this error message, users on xenial will get syntax errors from `argparse-addons` with no clear indication that `cantools` is actually the problem.

The most elegant solution would be simply the addition of `python_requires`. Unfortunately, `python3-pip` also pins to version 8.1.1, so `python_requires` is also not supported by the xenial `pip`.  Therefore, I also had to add the somewhat ugly`sys` error mechanism.

I tested this with a `ubuntu:16.04` docker image and also on ubuntu 20.04.  Attempts to install `cantools` on 16.04 results in a much better error message, while the 20.04 installation succeeds as usual.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,11 @@
 
 from setuptools import setup
 from setuptools import find_packages
+import sys
 import re
 
+if sys.version_info < (3, 6):
+    sys.exit('Sorry, Python < 3.6 is not supported')
 
 def find_version():
     return re.search(r"^__version__ = '(.*)'$",
@@ -12,6 +15,7 @@ def find_version():
 
 
 setup(name='cantools',
+      python_requires='>3.6.0',
       version=find_version(),
       description='CAN BUS tools.',
       long_description=open('README.rst', 'r').read(),


setup.py
```diff
@@ -2,8 +2,12 @@
 
 from setuptools import setup
 from setuptools import find_packages
+import sys
 import re
 
+import sys
```
eerimoq, line 8: Remove this duplicated import.

setup.py
```diff
@@ -2,8 +2,12 @@
...
+if sys.version_info < (3,6):
```
eerimoq, line 9: Add a space after `,`.

superjax:
> What if a dependency suddenly requires Python 3.7? Then the >=3.6 version check in cantools will be incorrect, right?

That's right, if `cantools` ever requires 3.7, then you'll want to update the `setup.py` to reflect that.

> What did the error message look like before this patch was applied?

I got this error:

```

**Comment by eerimoq** [comment]
Sounds like you should file a bug report on pip/setuptools asking for better error reporting. =)

I can merge this PR, but the version will likely not be updated when a dependency suddenly needs a newer Python version.

File: setup.py [hunk 3]
```diff
@@ -12,6 +15,7 @@ def find_version():
 
 
 setup(name='cantools',
+      python_requires='>3.6.0',
```
eerimoq, line 18: Change to `>=3.6.0`?

altendky:
I dunno...  I get the interest in supporting Ubuntu LTS but keep in mind that means supporting 8 year old pip and 4 years past EOL python 3.5 by the time 16.04 dies in 2024.  That's a lot of painful compatibility.  To include an alternative policy, there is https://numpy.org/neps/nep-0029-deprecation_policy.html.  Also, people can use pyenv or deadsnakes as well as virtual environments.

altendky:
Also, the actual reported error here is a problem with argparse_addons not having python_requires so I guess we hop over there and add the 3.6+.

setup.py
```diff
@@ -2,8 +2,11 @@
 
 from setuptools import setup
 from setuptools import find_packages
+import sys
 import re
 
+if sys.version_info < (3, 6):
```
andlaus, line 8: in order to reduce the future maintainance hassle, please change this to specifying the minimum python version in a single place, i.e.:
```python
min_python_version = (3,6)
min_python_version_string = ".".join(map(str, min_python_version))
if sys.version_info < min_python_version:
  sys.exit(f'Sorry, Python < {min_python_version_string} is not supported')
# ...
setup(name='cantools',
      python_requires=f'>={min_python_version_string}'
      # ...
      )
```

andlaus (changes requested):
I am in favor of specifying the minimum python version explicitly here, but for maintainance reasons I want having to do this only once.

setup.py
andlaus: (interestingly enough, this even works with python 2: In the past, I occasionally ran `setup.py` using python 2.7 by mistake and got very non-intuitive error messages.)

altendky:
How long is it planned this manual reimplementation of a long-standing Python feature will need to be maintained?  End of standard support for Ubuntu 16.04 is April 2021, EOL is 2024.  If targeting the end of standard support then likely the next 'maintenance' is deletion of this code.  Also keep in mind that this is entirely lost in the wheel...  So, it's only useful for explicit VCS installs or `--no-binary` that'll get the sdist.  It doesn't do anything for a regular `pip install cantools` since it'll just grab the wheel (and this check will have been evaluated when \_building\_ the wheel, not installing it).

Looks like [pip 9 is sufficient](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) to respect `python_requires`.  So, the case of concern is pip<9 that is explicitly operating on source or sdist.

As to the concern that a dependency could suddenly require 3.7 while this still describes only a requirement of 3.6, it is a non-issue.  If we all properly describe our requirements then a 3.6 install of cantools will not get the new 3.7+ version of the dep, but rather the most recent 3.6 compatible one, and everything is happy.  If you update the minimum required version of a dependency to one that requires 3.7+ then presumably you will start failing your 3.6 tests.  Your Python version requirements and tested versions ought to agree.  So, you drop 3.6 testing and update your `python_requires` to 3.7.  Sure, there's some manual maintenance there but it is between your expressed compatibility and your tests, nothing directly about manually keeping tabs on the minimum Python version requirements of your dependencies.

If this code is tested against 3.6+, just express that.

andlaus:
cantools itself hasn't been supporting python 2.7 for a long time, but it is very nice to get an error message that is understandable by mere mortals when accidentally running `python setup.py` instead of `python3 setup.py`. Further, I'd estimate the required maintenance effort for this microfeature to be about 2 minutes per year. maximum.

altendky:
It relates to pip version, not so much Python 2.

altendky:
Sorry, depending which thing you were referring to.  `python_requires` is just right.  `if sys.version` is about dealing with pip<9 non-wheel installs.

andlaus:
> Sorry, depending which thing you were referring to. python_requires is just right. if sys.version is about dealing with pip<9 non-wheel installs.

yes, I also get the impression that we were talking about different things. I was *not* thinking about packaging at all, but about manually building and testing. The best approach would probably be to manually check if the setuptools module supports `python_requires` and only bail out manually manually if it doesn't.(i.e., let setuptools handle the other cases.) Unfortunately, `python_requires` does not seem to do anything to protect against accidentally running `setup.py` using the wrong python version in this scenario. I.e. if this protection is desired we need the manual check anyway :/

setup.py
altendky: Yes.

```suggestion
      python_requires='>=3.6',
```

altendky:
FWIW, you basically shouldn't run `setup.py` anymore.  `pip` installs, `build` builds sdists and wheels.  Other than a project where I have a really silly thing setup to run on any `setup.py` invocation, including just `--help`, I can't really remember the last time I ran it.  Note that it may just be pip, and not setuptools, that actually complains based on `python_requires`.

altendky:
(but sure, lots of people do still run `setup.py`)

andlaus:
Interesting, I did not know that: So far, I only used `pip` to install modules which I didn't clone manually. Are you saying that it is better to use it even in this case? If yes, I suppose that it would be a good idea to explicitly mention this in `README.rst` and also `setup.py` should probably be amended to refuse direct invocation... (also, the `Makefile` should be adapted not to do this anymore)

altendky:
I use `venv/bin/pip install -e .` all the time.  And when needed, `venv/bin/python -m build --sdist --wheel .`.  Without concluding that the readme here should not cover installation, I really do prefer it when there's a link to some place that describes installation of Python packages.  There are messy packages that need their own instructions and their are extra special good packages that need their own instructions, but in the middle everything else should get installed roughly the same way.  And, there are piles of considerations like not using your system-provided pip in Linux (especially not with sudo), generally working in an env (virtualenv or venv), using pip not `setup.py`, and whatever else I'm forgetting at the moment (oh, Docker isn't relevant to using an env or not, etc etc).  I suppose the doc would also cover decent ways to get whatever Python version you want.  It's hard to get this all described well and correctly, let alone do it in the context of the quick start people want in a readme.  On the flip side, `pip install thing` can easily result on many systems in a permission error which induces the hazardous knee jerk sudo response.

As to having `setup.py` actually check and keep people from running it?  I wouldn't do that.  With modern tooling it can simply not exist at all, but I wouldn't do some interrim thing out of the way.  Mostly I'm just encouraging letting the existing features and tools do their job.  If someone is going to use ancient tooling then...  well, it isn't that they 'deserve' a mess but it probably isn't worth your while to weasel out all the possibilities and ways to compensate for them without causing other trouble.

But, this discussion is completely blowing out any savings that could be had by not maintaining extra manual checks so...  no matter.  `:]`  Just be sure that `python_requires` is set correctly as it is the thing that hopefully most people will be affected by.

andlaus:
hm, seems like you are quite a bit more familiar with the python packaging system than anyone else here. It would be nice if you created a PR updating the build/install instructions for the modern tooling, because it seems that people are currently strongly steered towards the legacy ways?! I in particular, I'm interested in how the supposed development workflow is supposed to be without calling `setup.py` explicitly, e.g. running the test suite and the coverage tools locally. If there is good introductory documentation for this on the web, a tl;dr plus a link would suffice indeed.

altendky:
Note that the `.` passed to pip refers to the directory where `setup.py` or `pyproject.toml` are located.

| `setup.py` | other |
| --- | --- |
| `venv/bin/python setup.py install` | `venv/bin/pip install .` |
| `venv/bin/python setup.py develop` | `venv/bin/pip install --editable .` |
| `venv/bin/python setup.py sdist` | `venv/bin/python -m build --sdist .` |
| `venv/bin/python setup.py bdist_wheel` | `venv/bin/python -m build --wheel .` |

The test stuff from `setup.py` just isn't a thing I've seen used except in a very few projects.  I don't know the details of why it didn't catch on.  Is it `unittest`-only and as such the pytest popularity was it's death?  Certainly having a standardized way to launch tests could be a useful thing in various ways, but the ability to do so in `setup.py` isn't functionally standard across projects.  It has been [officially deprecated by setuptools](https://setuptools.readthedocs.io/en/stable/userguide/commands.html#test-build-package-and-run-a-unittest-suite).

image

Are there other things you have been doing with `setup.py`?  And yes, there are umpteen ways to do each and every thing it seems and the community is not being successfully steered, as a whole, towards the new ways.  Without meaning to poke too hard or use this to claim it is entirely bad, the focus on backwards compatibility with EOL versions of Python and their tools kind of plays into the confusion.

As a bit of a side note, `map()` and `filter()` are being significantly supplanted (at least for some) by comprehensions.

```python
min_python_version_string = ".".join(map(str, min_python_version))
```

```python
min_python_version_string = ".".join(str(n) for n in min_python_version)
```

Certainly doesn't particularly matter here, but they allow expressions to be evaluated without requiring functions (inclusive of lambdas) to be defined and also support filtering and multiple loops in the same comprehension.  Anyways, just a little tidbit that may or may not be news and may or may not be of interest.

Anyways, I would definitely recommend the `python_requires` gets in place either with or without the 'manual' check.

@superjax, there's been a lot of discussion here of adjacent stuff.  Are you comfortable with the requested changes?  Or should we take another pass at clarifying them?  I'm not an authority in this repo, just trying to make sure we get back to the actual PR to get it finished out.  `:]`

zariiii9003:
Closing this, we don't have a `setup.py` file anymore.

PR #214, opened 2020-07-29
Close #184 dbc dump with extended multiplexing

### LINKED ISSUE #184: Extended multiplexing dbc dump
I observed the following effect and would like to discuss whether it is intended or not:

1. import a dbc that uses extended multiplexing
2. dump it to a new file (after applying some changes, not relevant for the multiplexing aspect)
3. open the new file with a dbc editor (I use that one from a well known company in Stuttgart …)
-->  much of the multiplexing info is missing, and therefore many conflicts are reported (overlapping signals …).

I inspected the dumped dbc file and found out that there's no section ```SG_MUL_VAL_```.
That section seems to be mandatory when using extended multiplexing.
Extended multiplexing has two aspects:
* allows more than one multiplexer signal within the same message
* allows assigning more than one multiplexer value to the same signal
The section ```SG_MUL_VAL_``` has to be used even if only one of these aspects apply.

Extending a dump for that section is not a big effort, I already have prepared it for my porpuses.
But I'm interested in other users opinion, especially in the possibilities or restrictions. E. g.:
* How to decide if database allows extended multiplexing (the internal help of my dbc editor says 'for J1939' and some others protocols, but not for a "normal CAN extended" dbc.
* Are there any interferences with other database formats? ```cantools``` supports different database formats, but I've been using dbc only so far ...

geo4711:
This PR supports dbc export with extended multiplexing. Extended Multiplexing is only applied if necessary, i. e. if at least one of the following items is given:
- a message contains more than one multiplexer signal
- a signal is multiplexed by more than one multiplexer value


geo4711:
sorry, forgot to upload the new files ...

PR #215, opened 2020-07-29
Close #184: support extended multiplexing for dbc export

### LINKED ISSUE #184: Extended multiplexing dbc dump

geo4711:
This PR support export of dbc with extended multiplexing. This is applied only if necessary, i. e. if at least one of the following items is fulfilled:
- a message has more than one multiplexer signals
- a signal is multiplexed by more than one multiplexer values


geo4711:
Hi!
I'm not familiar with AppVeyor. It tells me that the build process with Python 3.8 failed due to communication problems with github:
image
I'm not sure if I can solve this problem by myself (or if it es caused by my PR).
Could you have a look at it, please?

eerimoq:
I re-ran it. Now it passed.

eerimoq:
Let's add the DBC dump assertion at the end of the already existing multiplex tests (test?multiplex, test_multiplex_choices and test_multiplex_2). If the new dump code needs additional testing we add more tests (I guess one or more of those you wrote in this PR). Add *_dumped.dbc if necessary as the dump function may reorder some lines.

geo4711:
Thanks for re-running the build!
I've just added your proposal on my local repo (adding dump assertion to existing tests). There are two issues:
* multiplex_2.dbc results in an invalid dump (same with the version before this PR) --> I suggest to solve this separately. I haven't investigated the route cause yet; maybe three sub issues regarding the following line:
```SIG_VALTYPE_ 201588478 S9 : 1;BA_DEF_ BO_  "TpJ1939VarDlc" ENUM  "No","Yes";```
  * #186 (SIG_VALTYPE with extended frame ids)
  * missing line break
  * section SIG_VALTYPE should be below BA sections, not before (at least the original file multiplex_2.dbc is coded like that, and when I manually adapt the dumped one it works again).
* The dump creates some redundant lines in the mux section. I. e. if one message needs lines in the mux section, ALL muxed signales are listed their. The specification I have for dbc format does not explicitly tell you how to deal with that. I have tolerated that because ...
  * the redundant info is tolerated (by cantools as well as Vector's CANdb++)
  * the previous version of cantools doesn't dump the mux section at all (i. e. it leads to loss of information)
  * I wanted to keep the code flat. But that's maybe a quite weak argument ;-) and maybe it's better to get that fixed with a couple of extra lines ...

Here's just a snapshot showing the differences in the mux section (left = multiplex_2.dbc, right = dumped version):
image

What do you think about these issues?

eerimoq:
I'll merge and probably make some modifications myself after the merge. Thanks for the PR =)

geo4711:
Sounds great. Thanks :-)

PR #216, opened 2020-08-05
Support SPN attribute for signals

approximatenumber:
Added SPN attribute to signal attributes.
Added SPN support to tests. But currently tests are broken in `master`, only python-tests are passed successfully.

### DIFF
--- a/cantools/database/can/signal.py
+++ b/cantools/database/can/signal.py
@@ -391,7 +391,7 @@ class Signal(object):
 
     @property
     def spn(self):
-        """The SPN (Suspect Parameter Number) value if the signal
+        """The J1939 Suspect Parameter Number (SPN) value if the signal
         has this attribute, ``None`` otherwise.
         
         """
--- a/cantools/database/can/signal.py
+++ b/cantools/database/can/signal.py
@@ -124,7 +124,8 @@ class Signal(object):
                  multiplexer_ids=None,
                  multiplexer_signal=None,
                  is_float=False,
-                 decimal=None):
+                 decimal=None,
+                 spn=None):
         self._name = name
         self._start = start
         self._length = length
@@ -145,6 +146,7 @@ class Signal(object):
         self._multiplexer_ids = multiplexer_ids
         self._multiplexer_signal = multiplexer_signal
         self._is_float = is_float
+        self._spn = spn
 
     @property
     def name(self):
@@ -387,6 +389,19 @@ class Signal(object):
     def multiplexer_signal(self, value):
         self._multiplexer_signal = value
 
+    @property
+    def spn(self):
+        """The SPN (Suspect Parameter Number) value if the signal
+        has this attribute, ``None`` otherwise.
+        
+        """
+
+        return self._spn
+
+    @spn.setter
+    def spn(self, value):
+        self._spn = value
+
     def choice_string_to_number(self, string):
         for choice_number, choice_string in self.choices.items():
             if choice_string == string:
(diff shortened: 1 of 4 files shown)


cantools/database/can/signal.py
```diff
@@ -387,6 +389,19 @@ def multiplexer_signal(self):
     def multiplexer_signal(self, value):
         self._multiplexer_signal = value
 
+    @property
+    def spn(self):
+        """The SPN (Suspect Parameter Number) value if the signal
+        has this attribute, ``None`` otherwise.
```
eerimoq, line 395: You should mention J1939.

approximatenumber: You may that I need to validate that DBC file is based on J1939 before SPN parsing?

eerimoq: Nothing like that. Just Mention J1939 in the docstring. SPN is part of the J1939 specification.

approximatenumber: Got it, mentioned.

eerimoq:
Just add a test case (and/or extend an existing test case) and I'll merge.

approximatenumber:
> Just add a test case (and/or extend an existing test case) and I'll merge.

Added SPN attribute test to existing `test_attributes`
Added SPN attribute to `attributes.dbc`

eerimoq:
I'm confused. You add the new attribute ``spn`` to the Signal class, but the test does not test that at all. Please test the new code, nothing else. =)

approximatenumber:
ok, should I leave attribute test and SPN values in DBC file?

eerimoq:
I suggest removing that. Let's just add a test of the new code now. Add SPN to a signal in `tests/files/dbc/j1939.dbc` and check it in the test `test_j1939`.

approximatenumber:
> I suggest removing that. Let's just add a test of the new code now. Add SPN to a signal in `tests/files/dbc/j1939.dbc` and check it in the test `test_j1939`.

Yep, done

geo4711:
Hi!
Sorry for barging in ...
I've seen that the SPN is read from the attribute list. But I haven't found a piece of code that writes back the SPN value to an attribute.
Of course, this only matters if you change the member variable spn via the setter methode ...

Did I miss anything about that? Perhaps you could double check that.
Thanks :-)

eerimoq:
I'm open to adding dump functionality later. Just reading is a good start.

tests/test_database.py
```diff
@@ -3880,6 +3881,20 @@ def test_j1939_dbc(self):
...
+        # test SPN signal attribute
+        attribute = attributes['SPN']
+        self.assertEqual(attribute.name, 'SPN')
+        self.assertEqual(attribute.value, 500)
+        self.assertEqual(attribute.definition,
+                         db.dbc.attribute_definitions['SPN'])
+        self.assertEqual(attribute.definition.kind, 'SG_')
+        self.assertEqual(attribute.definition.type_name, 'INT')
+        self.assertEqual(attribute.definition.minimum, 0)
+        self.assertEqual(attribute.definition.maximum, 524287)
```
eerimoq, line 3895: Remove all the above.

approximatenumber:
> I can't find any test verifying that the SPN loaded from the dbc-file can be accessed using the spn attribute on the signal class. That is the most relevant test you can add, . Please add it.

Sorry for misunderstanding. Please, check. Just added a simple accessability test.

RangelReale:
Shouldn't the SPN attribute use the default value if available? Currently it doesn't take it in account.

In the test file `dbc/j1939.dbc`, we have

```
BA_DEF_DEF_  "SPN" 0;
```

but on the test we check

```python
        signal = db.messages[1].signals[0]
        self.assertEqual(signal.spn, None)
```

shouldn't it be 0?

PR #218, opened 2020-08-13
message: compare ranges with Decimal values

mguentner:
since `value` is a `Decimal` and `signal.{minimum, maximum}` are both `float`s,
the check might fail due to floating point errors, even if `value` is
within the range but very close to / exactly `signal.{minimum, maximum}`.

Thank you for the great tools and library!

### DIFF
--- a/tests/test_database.py
+++ b/tests/test_database.py
@@ -4923,6 +4923,37 @@ class CanToolsDatabaseTest(unittest.TestCase):
             db,
             'tests/files/dbc/issue_184_extended_mux_cascaded_dumped.dbc')
 
+    def test_floating_point_limits(self):
+        """Test correct handling of limits in `can.Message.encode` if they
+           are floating point.
+        """
+
+        dbc_content = """
+VERSION "1337"
+
+BS_:
+
+BU_: XYY XYZ
+
+BO_ 123 Message_1: 6 XYY
+  SG_ Alpha : 7|8@0+ (1,0) [0|255] "" XYZ
+  SG_ Bravo : 9|2@0+ (1,0) [0|3] "" XYZ
+  SG_ Charlie : 15|4@0+ (1,0) [0|14] "" XYZ
+  SG_ Delta : 23|8@0+ (0.394,0) [0.394|100.076] "" XYZ
+        """
+        db = cantools.database.load_string(dbc_content, 'dbc')
+        message = db.get_message_by_name('Message_1')
+        data_min = { signal.name: signal.decimal.minimum for signal in message.signals }
+        message.encode(data_min)
+        data_max = { signal.name: signal.decimal.maximum for signal in message.signals }
+        message.encode(data_max)
+
+        # outside of lower bound for Delta
+        with self.assertRaises(cantools.database.EncodeError):
+            message.encode({ **data_min, 'Delta': data_min['Delta']-Decimal('0.001') })
+        # outside of high bound for Delta
+        with self.assertRaises(cantools.database.EncodeError):
+            message.encode({ **data_max, 'Delta': data_max['Delta']+Decimal('0.001') })
 
 # This file is not '__main__' when executed via 'python setup.py3
 # test'.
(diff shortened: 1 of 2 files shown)


eerimoq:
How do you know that `value` is `Decimal`? Looks like `value` is any number the user passes to the encode function.

mguentner:
Makes a lot of sense. I discovered this issue while writing a test that tries to encode `signal.decimal.maximum` / `signal.decimal.minimum`, then decodes the bytes again and verifies whether the values are the same.

You are completely right that `value` can be anything. But indeed, using decimal here is a bit of an improvement as at least one side of the comparison is safe(r).

> However, anyone using floating point numbers should expect rounding errors and such in many places in their code and this library. If you need exact representation of a number, use integers.

Absolutely! :100:

eerimoq:
Please add a test, just like you described, and I'll merge. Also, make the if-statements check `signal.decimal.minimum/maximum` instead of the floating point values.

There will be a slight performance hit comparing decimal instead of float I guess, but that it probably fine.

mguentner:
If you check out https://github.com/eerimoq/cantools/pull/218/commits/4414db3baaff5dca8fb34865273b45a10b224c99 run `make test`, it will fail with
```
cantools.database.errors.EncodeError: Expected signal 'Delta' value greater than or equal to 0.394 in message 'Message_1', but got 0.394.
```

If you comment out that line it will fail with
```
cantools.database.errors.EncodeError: Expected signal 'Delta' value less than or equal to 100.076 in message 'Message_1', but got 100.076.
```

I also adjusted the `not None` check to `decimal` :+1:

tests/test_database.py
```diff
@@ -4923,6 +4923,29 @@ def test_issue_184_cascaded_multiplexors(self):
             db,
             'tests/files/dbc/issue_184_extended_mux_cascaded_dumped.dbc')
 
+    # tests correct handling of limits in message.encode when they
+    # are floating point
```
eerimoq, line 4927: Make this a method docstring.

tests/test_database.py
```diff
@@ -4923,6 +4923,29 @@ def test_issue_184_cascaded_multiplexors(self):
...
+
+BO_ 123 Message_1: 6 XYY
+  SG_ Alpha : 7|8@0+ (1,0) [0|255] "" XYZ
+  SG_ Bravo : 9|2@0+ (1,0) [0|3] "" XYZ
+  SG_ Charlie : 15|4@0+ (1,0) [0|14] "" XYZ
+  SG_ Delta : 23|8@0+ (0.394,0) [0.394|100.076] "" XYZ
+        """
+        db = cantools.database.load_string(dbc_content, 'dbc')
+        message = db.get_message_by_name('Message_1')
+        self.assertIsNotNone(message)
```
eerimoq, line 4944: Remove this assert.

tests/test_database.py
```diff
@@ -4923,6 +4923,29 @@ def test_issue_184_cascaded_multiplexors(self):
...
+        data_min = { signal.name: signal.decimal.minimum for signal in message.signals }
+        message.encode(data_min)
+        data_max = { signal.name: signal.decimal.maximum for signal in message.signals }
+        message.encode(data_max)
```
eerimoq, line 4948: Also test values just outside the limits.

mguentner:
Implemented all comments of the review.

eerimoq:
Unfortunately this did not work for signed signals. My initial idea was to use float everywhere and just save the values as decimal to have the exact value available to the user and for dumping the database. I don't have any interest in this myself, so ideas are welcome.

mguentner:
Why does it not work for signed signals? Can you please elaborate.

eerimoq:
Runtime error dividing a float by a decimal (or maybe the other way around) when scaling the signal value.

mguentner:
So the code changed by this PR does not work for signed + scaled signals or your initial idea doesn't?
If the former, can you maybe provide an example DBC? I can have a look,

eerimoq:
I'm not sure if the current code can encode Decimal signals.

PR #226, opened 2020-10-30
Added missing cdd parser data object element path

tpierson-voxx:
While importing some CDD files, data elements are missing. Examining the CDD file content, there is an element path that is not considered in cdd.py. This PR adds the missing element path.

### DIFF
--- a/cantools/database/diagnostics/formats/cdd.py
+++ b/cantools/database/diagnostics/formats/cdd.py
@@ -158,6 +158,7 @@ def _load_did_element(did, data_types):
     offset = 0
     datas = []
     data_objs = did.findall('SIMPLECOMPCONT/DATAOBJ')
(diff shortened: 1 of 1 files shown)


eerimoq:
Please add a test and I can merge.

tpierson-voxx:
@eerimoq to properly test, data records will need to be added into the CDD file. Otherwise testing is accomplished with the current test cases by ensuring the added code did not break anything. Would you recommend adding the new test records to the test CDD file?

eerimoq:
You should test that the new functionality is correctly implemented. I'm a novice at CDD files so I can't say if adding new records to the existing CDD file is according to spec, or if a new CDD file is needed.

tpierson-voxx:
@eerimoq I'm going to abandon the PR. To make a proper CDD parser, a lot of work is required due to the different types of objects (CDD files are "object oriented", there can be DID's, routines, security services, etc). The entire CDD parser will need to be reworked.

PR #240, opened 2020-11-30
implement multi-lingual message and signal descriptions

andlaus:

ATTENTION: This patch includes some changes outside of the ARXML loader.

For now, the default is to assume that comments which don't explicitly specify a language are in English. An alternative would be to check if the dictionary containing the comments contains exactly a single element and use that instead. The problem with this approach is that comments for different languages might get mixed up more easily: Assume a supplier based in China who documents its signals only in Chinese and a OEM based in Brazil who likes to document its stuff using Portuguese...

### DIFF
--- a/cantools/database/can/message.py
+++ b/cantools/database/can/message.py
@@ -55,7 +55,19 @@ class Message(object):
         self._length = length
         self._signals = signals
         self._signals.sort(key=start_bit)
-        self._comment = comment
+
+        # if the 'comment' argument is a string, we assume that is an
+        # english comment. this is slightly hacky because the
+        # function's behavior depends on the type of the passed
+        # argument, but it is quite convenient...
+        if isinstance(comment, str):
+            # use the first comment in the dictionary as "The" comment
+            self._comments = { None: comment }
+        else:
+            # assume that we have either no comment at all or a
+            # multi-lingual dictionary
+            self._comments = comment
+
         self._senders = senders if senders else []
         self._send_type = send_type
         self._cycle_time = cycle_time
@@ -220,13 +232,31 @@ class Message(object):
     def comment(self):
         """The message comment, or ``None`` if unavailable.
 
-        """
+        Note that we implicitly try to return the comment's language
+        to be English comment if multiple languages were specified.
 
-        return self._comment
+        """
+        if self._comments is None:
+            return None
+        elif self._comments.get(None) is not None:
+            return self._comments.get(None)
+
+        return self._comments.get('EN', None)
+
+    @property
+    def comments(self):
+        """The dictionary with the descriptions of the message in multiple languages. ``None`` if unavailable.
+
+        """
+        return self._comments
 
     @comment.setter
     def comment(self, value):
-        self._comment = value
+        self._comments = { None: value }
+
+    @comments.setter
+    def comments(self, value):
+        self._comments = value
 
     @property
     def senders(self):
@@ -910,4 +940,4 @@ class Message(object):
             self._frame_id,
             self._is_extended_frame,
             self._length,
-            "'" + self._comment + "'" if self._comment is not None else None)
+            self._comments)
--- a/cantools/database/can/formats/arxml.py
+++ b/cantools/database/can/formats/arxml.py
@@ -132,7 +132,7 @@ class SystemLoader(object):
         frame_id = self._load_message_frame_id(can_frame_triggering)
         length = self._load_message_length(can_frame)
         is_extended_frame = self._load_message_is_extended_frame(can_frame_triggering)
-        comment = self._load_message_comment(can_frame)
+        comments = self._load_message_comments(can_frame)
 
         # ToDo: senders
 
@@ -181,13 +181,13 @@ class SystemLoader(object):
                        send_type=None,
                        cycle_time=cycle_time,
                        signals=signals,
-                       comment=comment,
+                       comment=comments,
                        bus_name=None,
                        strict=self._strict)
 
     def _load_message_name(self, can_frame_triggering):
         return self._get_unique_arxml_child(can_frame_triggering,
-                                           "SHORT-NAME").text
+                                            'SHORT-NAME').text
 
     def _load_message_frame_id(self, can_frame_triggering):
         return int(self._get_unique_arxml_child(can_frame_triggering,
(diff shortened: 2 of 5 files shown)


andlaus:
the test coverage should include the missing lines once the (ARXML-3)[https://github.com/andlaus/cantools/commit/c9df74e84daac12f41978892ce623c79fb24cfc6] patch is in. I will add it to the ARXML-4 files if you want, but this would slightly delay things for the rest of this series...

cantools/database/can/message.py
```diff
@@ -27,7 +27,7 @@ def __init__(self,
                  name,
                  length,
                  signals,
-                 comment=None,
+                 comments=None,
```
eerimoq, line 30: This will break scripts with `Message(..., comment="My comment.")`. I suggest that we leave the `comment` parameter unmodified and add `comments=None` at the end of the parameter list.

cantools/database/can/message.py
```diff
@@ -55,7 +55,19 @@ def __init__(self,
...
-        self._comment = comment
+
+        # if the 'comments' argument is a string, we assume that is an
+        # english comment. this is slightly hacky because the
+        # function's behavior depends on the type of the passed
+        # argument, but it is quite convenient...
+        self._comments = {}
+        if isinstance(comments, str):
+            # use the first comment in the dictionary as "The" comment
+            self._comments = { "EN": comments }
```
eerimoq, line 66: Hmm... As the language of the comment is unknown it's probably better to do `self._comments = {None: comments}`.

cantools/database/can/message.py
```diff
@@ -55,7 +55,19 @@ def __init__(self,
...
+        elif comments is not None:
+            # assume that we have a multi-lingual dictionary
+            self._comments = comments
```
eerimoq, line 69: Replace `elif comments is not None:` with `else:` as `self._comments` should be `None` if there are no comments.

cantools/database/can/message.py
```diff
@@ -220,13 +232,28 @@ def signal_groups(self, value):
...
 
-        return self._comment
+        """
+        return self._comments
 
     @comment.setter
     def comment(self, value):
-        self._comment = value
+        # we implicitly assume english as the language of "the" comment
+        self._comments = { "EN": value }
```
eerimoq, line 252: Must handle setting the comment to `None`.

eerimoq (commented):
Let me know what you think about my suggestions.

cantools/database/can/message.py
andlaus: okay, changed. thanks for the remark.

cantools/database/can/message.py
andlaus: changed. Thanks!

cantools/database/can/message.py
andlaus: IMO this would be a slightly bad idea  because the code could no longer rely on the fact that the `_comments` attribute always is a dictionary. Shall I still change it?

cantools/database/can/message.py
andlaus: Done in the latest revision. (I hope)

andlaus:
the last revions had some serious issues, sorry for that. The new one fixes this.

cantools/database/can/message.py
eerimoq: Well, the docstrings for both the `comment` and the `comments` properties says that it is `None` if there are no comments available. And `self._comments` is return from one of them.

The `comment` property should at least continue to return `None` if there is no comment. The `comments` property could return I dictionary I guess, and in that case the docstring should be updated.

andlaus: good point. I've changed it in the latest revision.

eerimoq:
It's not obvious to me how this should be designed. Suppose you have an ARXML-file with comments in german (no english) which you want to convert to a DBC-file. With the logic in the PR the DBC-file would not contain any comments as only `None` and `EN` are used. I would expect the DBC-file to have comments in german. Converting from DBC to ARXML (which is not currently supported) would have similar problems.

One alternative design is to add a language argument to all classes and functions that need to know the language in use. By default it would probably be english. All file formats without language information would gladly accept the language choice. Only file formats like ARXML would actually use the language. This would allow loading a DBC-file with comments in german to generate an ARXML with "DE" comments (given that the user passes `language='DE'` when loading the DBC-file.

`.comment` is in the selected language. `.comments` is a dictionary of comments in all available languages.

andlaus:
I suppose that the root of this problem is that the capabiities of DBC file format and ARXML are not identical for our purposes. I agree that converting all internal code to always use a dictionary when it comes to comments would do the trick, as far as I can see. For the ARXML-to-DBC converter, I also see this the same as you: One could have specify the language which ought to be used in the DBC externally and let the `.comment` attribute default to something reasonable (e.g., 'EN' if multiple languages are defined, and the only element in the dictionary if the dictionary size is one)? The question then becomes: What's a good default policy?

andlaus:
BTW: Do you happen know why coveralls again claims that the overall coverage is down, but if you inspect the results, coverage for all individual files is equal or increased? (maybe I just misinterpret its results)

andlaus:
ok, let's do it this way then. later?

eerimoq:
Sure, let's postpone it if it makes things easier for you.

andlaus:
could you please press The Green Button? (it probably is grey ATM, though ;)

andlaus:
thank :)

eerimoq:
I actually meant to postpone the whole PR, but let's just redesign it later (probably before a new release).

andlaus:
> I actually meant to postpone the whole PR, but let's just redesign it later (probably before a new release).

+1. For my current goal (i.e., proper support for Daimler-style ARXML-3 CAN description files), this is just a side concern, though. That said, I'll help with proper internationalization support where I can.

eerimoq:
What's special with Daimler ARXML-file? Will other companies/users still be able to use cantools? I certainly hope so.

andlaus:
> What's special with Daimler ARXML-files?

not much. They are basically just standard ARXML-3 files but they use a lot of features, and use a weird XML namespace.

So far all of my PRs equally applied to ARXML-4...

andlaus:
i.e., xmllint says they validate against the AUTOSAR 3.2 XSD which I found in the internet after changing the namespace.

eerimoq:
Great, you certainly add much value to the ARXML module, and if it's close enough to the standard it should be easy to support the general case as well if someone asks for it/implements.

PR #241, opened 2020-12-01
Miscellaneous fixes for the ARXML code

andlaus:

We are getting closer to the wrap up: These are small unrelated issues I've stumbled over while hacking on the ARXML parser: The first patch makes all floating point types signed (because IEEE754 types are always signed), the second patch recognizes more encodings for integers because ARXML also supports non-two complement encodings (actually decoding the CAN frames is left for another day, though) and in the last patch, I've added UUID attributes  to the `system-4.2.arxml` file to aid debugging.

### DIFF
--- a/cantools/database/can/formats/arxml.py
+++ b/cantools/database/can/formats/arxml.py
@@ -492,7 +492,15 @@ class SystemLoader(object):
 
             base_type_encoding = base_type_encoding.text
 
-            if base_type_encoding == '2C':
+            if base_type_encoding in ('2C', '1C', 'SM'):
+                # types which use two-complement, one-complement or
+                # sign+magnitude encodings are signed. TODO (?): The
+                # fact that if anything other than two complement
+                # notation is used for negative numbers is not
+                # reflected anywhere. In practice this should not
+                # matter, though, since two-complement notation is
+                # basically always used for systems build after
+                # ~1970...
                 is_signed = True
             elif base_type_encoding == 'IEEE754':
                 is_float = True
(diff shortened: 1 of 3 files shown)


andlaus:
the coveralls result seems to exhibit the usual weirdness: overall coverage decrease with no decrease in any individual file..

cantools/database/can/formats/arxml.py
```diff
@@ -492,9 +492,18 @@ def _load_signal_type(self, i_signal):
...
+                # sign+magnitude encodings are signed. TODO (?): The
+                # fact that if anything other than two complement
+                # notation is used for negative numbers is not
+                # reflected anywhere. In practice this should not
+                # matter, though, since two-complement notation is
+                # basically always used for systems build after
+                # ~1970...
                 is_signed = True
             elif base_type_encoding == 'IEEE754':
+                is_signed = True # all standardized IEEE-754 floating point types are signed
```
eerimoq, line 506: `is_signed` is only used for integers. It is never used in combination with `is_float`. At least that is how this package is designed atm.

andlaus: okay, thanks for the clarification; this wasn't clear to me. maybe `is_signed` should be renamed to `is_signed_int`?

eerimoq: It's fine as is. Maybe clarify in its property docstring.

andlaus: okay, I'll remove the patch...

andlaus: done

PR #242, opened 2020-12-01
ARXML: fix determining signal lengths their initial value and its range

andlaus:

This PR adds support for determining the length of a given signal via the system signal if the length is not specified by the i-signal itself (first patch),  it corrects determining the possible range of signals (second patch; the possible value range of the signal is the range over all COMPU-SCALEs, not just the first one)  and it implements loading the initial value of signals (third patch).

The next PR will contain the changes for the differences between ARXML-3 and ARXML-4.

### DIFF
--- a/cantools/database/can/formats/arxml.py
+++ b/cantools/database/can/formats/arxml.py
@@ -237,6 +237,7 @@ class SystemLoader(object):
             return None
 
         # Default values.
+        initial = None
         minimum = None
         maximum = None
         factor = 1
@@ -265,6 +266,20 @@ class SystemLoader(object):
             minimum, maximum, factor, offset, choices = \
                 self._load_system_signal(system_signal, decimal, is_float)
 
+        # loading constants is way too complicated, so it it is the job of a separate method
+        initial = self._load_arxml_init_value_string(i_signal)
+
+        if initial is not None:
+            if is_float:
+                initial = float(initial)
+            elif initial.strip().lower() == "true":
+                initial = True
+            elif initial.strip().lower() == "false":
+                initial = False
+            # TODO: strings?
+            else:
+                initial = int(initial)
+
         # ToDo: receivers
 
         return Signal(name=name,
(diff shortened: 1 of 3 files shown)


eerimoq:
No tests needed?

andlaus:
as usual, coverage results are strange, but this time in a different sense: there's a significant regression in `arxml.py'. I'll fix that tomorrow...

andlaus:
> No tests needed?

actually, there are, but I screwed up a bit: I put the these tests into the patch for autosar 3 (next PR) and surgically separated these patches later, while not adding proper tests for this to the existing AR4 infrastucture. sorry for the mess, I'll fix it ASAP.

eerimoq:
Great work overall!

eerimoq:
Btw, which features in cantools are Daimler using?

eerimoq:
@andlaus Any updates?

koppor:
I think, the fixing took longer. Because of other work, it will probably be mid-January until a follow-up. Sorry for the delay.

Oliver Kopp <oliver.kopp@mbition.io>, Mercedes-Benz AG on behalf of [MBition GmbH](https://mbition.io/).

eerimoq:
Ok!

andlaus:
ok, I finally managed update this. Note that I'm not 100% sure if determining the initial value of signals is correct (it is certainly not complete because it only includes the saner ways), but at least the `system-4.2.arxml` file still passes the XSD for autosar 4.3 (after changing the namespace).

andlaus:
@eerimoq : I guess this PR is ready to go in. I have no idea where the huge drop in coverage asserted by coveralls is supposed to come from: the PR only touches `arxml.py` when it comes to relevant files and that one has basically only a few error additional error cases uncovered which -- according to coveralls -- constitute 0.77% of the source code of that file.

andlaus:
> Btw, which features in cantools are Daimler using?

I can't speak for Daimler as a whole, but I personally use it to read the bus description files for various vehicles. Once the database object is read, it depends on the concrete application: sometimes I need to generate C++ code to decode frames, convert the descriptions to some simpler formats (like e.g. DBC) and sometimes I use the decoded data directly in python for some calculations. Overall I'm quite impressed by the overall code quality of cantools :)

PR #245, opened 2020-12-03
Adds support for handling j1939 and GM Extended frame ids

kdschlosser:
The current handling of the VFrameFormat I believe to be incorrect, This is not the "protocol", it is
a specifier on how the frame id has been formatted. The protocol lies in the ProtocolType attribute, this attribute defines the protocol used on top of CAN which would be the Transport layer (layer 4) of the OSI model. It is a way to know how to handle the data contained within the data section of the frame. An example would be setting the ProtocolType to OBD2. This instructs a user/application that the data contained is going to be as follows.

byte 0: number of data bytes (not including this byte). Because the OBD2 standard will always have a DLC of 8 any of the unused bytes will be populated with `0x0`
byte 1: mode or service identifier, a response frame will be the mode id + `0x40`
bytes 2-7: data

I have added the property `frame_format` to the Message class which now returns the set VFrameFormat for a message if this attribute has been set in the DBC file. I have changed the `protocol` property so it will no longer return the frame format and instead it will return the ProtocolType attribute if there is one in the DBC file.

GMLAN has a special way of handling the frame id just like the j1939 frame format does. in order to properly decode the frame id for GMLAN there is a special attribute that needs to get set to a 1, this attribute is `UseGMParameterIDs`. When using the `get_message_by_frame_id` I check to see if the supplied frame id is present in the stored message, if it is I return it. Then I check each message contained in the database and see if the message has the `UseGMParameterIDs` attribute and if it does I check to see if it is set to a 1, If it has been set to a 1 I then alter the supplied frame id so it is GMLAN compliant and check if it matches the message, if it does I return the message.py

for the j1939 frame format it functions almost identical to the GMLAN except it checks the message to see if the VFrameFormat attribute is set and if it starts with `J1936`

The handling of attributes has also been changed so that global attributes are available to access in a node, message and signal, If a global attribute is changed in a node, message or signal it does not change the global attribute, instead it creates an attribute specific to that object. If a global attribute needs to be changed it needs to be done at the database level.

I still need to work out changing a global attribute and having it also change in all of the objects in the database.


eerimoq:
I prefer the assert the other way around, even if actual and expected are swapped.

Why did you change the file path handling? Everything should run just fine when you run the tests from the project root folder.

kdschlosser:
If the tests get run from inside of an IDE like PyCharm the tests will not run because of the file path. Also having the asserts backwards produces output that is incorrect and when doing comparisons of the output it is backwards. Most developers will develop in an IDE like PyCharm and if things are not working properly inside of the IDE because of things like file paths and test output it makes it extremely difficult to produce code that is going to function properly and also test properly. It is much easier to be able to develop in a single environment then to have to keep jumping back and forth between a shell window and an IDE and copying and pasting output into the IDE to get comparisons. Doing that is an enormous headache.

eerimoq:
Developers prefer different tools. I like Emacs, many like Visual Code, many Vim, etc. It might be that your changes makes many of them happy, but right now these changes are not acceptable, sorry.

PR #246, opened 2020-12-08
Ability to generate pack functions (c-code) only for a specified sender node

dzid26:
Often for CANbus networks, each node in the network has assigned messages that it can transmit. 
That is indicated by a node name next to the message, eg in Motohawk.:
```
BU_: PCM1 FOO
BO_ 496 ExampleMessage: 8 PCM1
```
When generating code for such node, it is not needed to generate packing message code for messages other than assigned to the node.  

That's what this PR does - it adds ability to reduce c-code size (less dead code) based on the node name specified with an argument. 
Syntax is:
`cantools generate_c_source  <file>.dbc --sender_name <node_name>`

The way it was done was simple - I just split the code in `SIGNAL_DEFINITION_ENCODE_DECODE_FMT` to 
`SIGNAL_DEFINITION_DECODE_FMT` and `SIGNAL_DEFINITION_ENCODE_FMT`, etc. And added check if node name matches specified node via argument. 
I also created _motohawk_sender_ eaxamples and added feature description to the Readme.


cantools/subparsers/generate_c_source.py
```diff
@@ -84,4 +85,7 @@ def add_subparser(subparsers):
     generate_c_source_parser.add_argument(
         'infile',
         help='Input database file.')
+    generate_c_source_parser.add_argument(
+        '--sender_name',
```
eerimoq, line 89: I suggest that we rename this to `--node` as we want to generate sender and receiver code for given node. `--sender_name` is too specific to sending. Receiving is just as important.

cantools/database/can/c_source.py
```diff
@@ -1561,10 +1605,10 @@ def generate(database,
     structs = _generate_structs(database_name, messages, bit_fields)
     declarations = _generate_declarations(database_name,
                                           messages,
-                                          floating_point_numbers)
+                                          floating_point_numbers, sender_name)
     definitions, helper_kinds = _generate_definitions(database_name,
                                                       messages,
-                                                      floating_point_numbers)
+                                                      floating_point_numbers, sender_name)
```
eerimoq, line 1611: One parameter per line.

cantools/database/can/c_source.py
```diff
@@ -1350,11 +1363,18 @@ def _generate_declarations(database_name, messages, floating_point_numbers):
             signal_declaration = ''
 
             if floating_point_numbers:
-                signal_declaration = SIGNAL_DECLARATION_ENCODE_DECODE_FMT.format(
-                    database_name=database_name,
-                    message_name=message.snake_name,
-                    signal_name=signal.snake_name,
-                    type_name=signal.type_name)
+                if not sender_name or sender_name in message.senders:
```
eerimoq, line 1366: I suggest adding a helper function `is_sender(message, node_name)` for less duplicated code and improved readability.

cantools/database/can/c_source.py
```diff
@@ -1395,14 +1421,21 @@ def _generate_definitions(database_name, messages, floating_point_numbers):
...
-
+                if not sender_name or sender_name in message.senders:
+                    signal_definition += SIGNAL_DEFINITION_ENCODE_FMT.format(
+                        database_name=database_name,
+                        message_name=message.snake_name,
+                        signal_name=signal.snake_name,
+                        type_name=signal.type_name,
+                        encode=encode)
+                if not sender_name or not sender_name in message.senders:
+                    signal_definition += SIGNAL_DEFINITION_DENCODE_FMT.format(
```
eerimoq, line 1432: Typo `SIGNAL_DEFINITION_DENCODE_FMT`.

cantools/database/can/c_source.py
```diff
@@ -1428,16 +1461,23 @@ def _generate_definitions(database_name, messages, floating_point_numbers):
...
-                                               pack_unused=pack_unused,
-                                               unpack_unused=unpack_unused,
-                                               pack_variables=pack_variables,
-                                               pack_body=pack_body,
-                                               unpack_variables=unpack_variables,
-                                               unpack_body=unpack_body)
+            definition = ""
+            if not sender_name or sender_name in message.senders:
+                definition += DEFINITION_PACK_FMT.format(database_name=database_name,
+                                                database_message_name=message.name,
```
eerimoq, line 1467: Fix indentation.

eerimoq (commented):
Nice work! Have a look at my comments.

eerimoq:
I replaced Travis CI with Github Actions on the master branch. Please rebase on that branch so all tests can be run.

I'm out of credits on Travis CI...

cantools/subparsers/generate_c_source.py
dzid26: I actually had that at the beginning. Happy to change.

cantools/database/can/c_source.py
dzid26: Added with underscore as I think it shuold be internal:
_is_sender(message, node_name)

dzid26:
I still need to check what is causing the six failures.
I added the suggestions. I hope replacing the original commit is ok.

cantools/database/can/c_source.py
```diff
@@ -1350,11 +1365,18 @@ def _generate_declarations(database_name, messages, floating_point_numbers):
             signal_declaration = ''
 
             if floating_point_numbers:
-                signal_declaration = SIGNAL_DECLARATION_ENCODE_DECODE_FMT.format(
-                    database_name=database_name,
-                    message_name=message.snake_name,
-                    signal_name=signal.snake_name,
-                    type_name=signal.type_name)
+                if not node_name or _is_sender(message, node_name):
```
eerimoq, line 1368: Move `not node_name or` into the helper function to make this line `if _is_sender(message, node_name):`

cantools/database/can/c_source.py
```diff
@@ -1540,6 +1582,10 @@ def generate(database,
 
     Set `bit_fields` to ``True`` to generate bit fields in structs.
 
+    `node_name` specifies node for which message packers will be generated.
```
eerimoq, line 1585: Should be "... specifies the node ...", right?

cantools/database/can/c_source.py
```diff
@@ -1540,6 +1582,10 @@ def generate(database,
...
+    For all other messages, unpackers will be generated. If `node name` is not
```
eerimoq, line 1586: Typo, should be `node_name`.

cantools/database/can/c_source.py
```diff
@@ -1520,7 +1561,8 @@ def generate(database,
              source_name,
              fuzzer_source_name,
              floating_point_numbers=True,
-             bit_fields=False):
+             bit_fields=False,
+             node_name=''):
```
eerimoq, line 1565: I prefer `None` as default value to indicate not set.

cantools/subparsers/generate_c_source.py
```diff
@@ -84,4 +85,7 @@ def add_subparser(subparsers):
     generate_c_source_parser.add_argument(
         'infile',
         help='Input database file.')
+    generate_c_source_parser.add_argument(
+        '--node',
+        help='Generate pack functions only for messages from sender node and unpack functions for messages from all other nodes.')
```
eerimoq, line 90: Line too long, limit to around 85 characters.

cantools/database/can/c_source.py
dzid26: The problem is, currently:
1. packing functions  are generated if `_is_sender` 
2. unpacking functions  are generated if `not _is_sender` 
3. both packing and unpacking are generated if `node_name` name isn't provided (None).  

One function will not work to cover the 1,2  and 3. Also, it wouldn't make sense if node was not provided. 

Maybe it would be a bit mode clear if I checked against null instead of `not node_name`:
`_is_sender(message, node_name) or node_name == None` ?

cantools/subparsers/generate_c_source.py
dzid26: `Generate pack functions only for the sender node and unpack functions for other nodes`
I think this should be clear enough and is now within 85 characters.

cantools/database/can/c_source.py
eerimoq: Ahh, ok, thne I suggest having two helper functions, `_is_sender()` and `_is_receiver()`. Both returns true if `node_name` is `None`.

cantools/subparsers/generate_c_source.py
eerimoq: I suggest: `Generate pack functions for messages sent from given node, and unpack functions for other messages.`

eerimoq:
Use unix line ending please. =)

cantools/database/can/c_source.py
```diff
@@ -1339,8 +1350,13 @@ def _generate_structs(database_name, messages, bit_fields):
 
     return '\n'.join(structs)
 
+def _is_sender(message, node_name):
+    return node_name == None or node_name in message.senders
+
+def _is_receiver(message, node_name):
+    return node_name == None or not node_name in message.senders
```
eerimoq, line 1357: It's best practice to use `is` when comparing to `None`, as in `node_name is None`.

tests/files/c_source/motohawk_sender.c
```diff
@@ -0,0 +1,131 @@
+/**
```
eerimoq, line 1: Add a test that generates these (tests/files/c_source/motohawk_sender.h and  tests/files/c_source/motohawk_sender.c) and asserts that the generated code is equal to these files. Right now there is no test of the new functionality.

dzid26: Let me know if tests/files/c_source/motohawk_sender_node.* name is fine or perhaps would be better tests/files/c_source/motohawk_PCM1_node.*.

dzid26:
> Use unix line ending please. =)

Hopefully encoding will be good now. 
I was using WSL reading directly from Windows file system.

tests/files/c_source/motohawk_sender.c
eerimoq: The first one is fine.

eerimoq:
Hmmm, the motohawk DBC file only contains a single message. Would be nice to at least have one that given node sends and one that given node receives. It's probably easiest to create a new DBC file for this purpose.

dzid26:
I added a dbc from my other project to expand the sender_node testcase.

eerimoq:
I just had a look at the generated C source from motohawk.dbc. In that database the Temperature signal in the ExampleMessage message is actually both sent and received by the PCM1 node. The generated C code only contains pack, not unpack, which I guess it should.

I guess that it is rather unusual to both send and receive the same message, but either way we should do our best and generate unpack functions for all received messages.

dzid26:
@eerimoq I modified the behavior according to what you said. 
I think it is better now - we explicitly check signal receivers to determine whether given message belong to the given node.

If none of the signals has a matching receiver, there will be no `_decode` and no `_unpack` function created. 
If, say, only one signal has a matching receiver, the `_unpack` function will be created for _whole_ message and `_decode` function will be created _only_ for that signal. 
I think that should be fine. 

The only thing is that some people don't bother to specify receiver nodes for the signals, but I guess if they bother to use `--node` argument, then they should define the bus matrix fully.

dzid26:
There was a logical bug in c_source.py. 
When --node was specified with --no-floating-point, then corresponding for the node unpack functions would not be created.

dzid26:
I squashed the fix above. 

I am going to rebase to the latest.

eerimoq:
@andlaus Can you review this instead of me? I think you know more about the cantools code base than me now =)

dzid26:
@andlaus That's good. 
I noticed I need to add one more improvement that showed up with a `dbc` with more than two nodes.

dzid26:
Now I am quite happy with the --node_name argument behavior. 
I squashed extra changes into the last two commits. 

The --node_name will allow to only generate functions (encode, decode, range_check), header constants and signals in unpack
that are related to specified node_name.

The exception is the message struct. For that, even in the case when signal members are unused (not indicated as receivers in the communication matrix), all struct members will be generated.

andlaus (commented):
looks okay to me. I had a few issues using this on an internal ARXML file. Most of these where caused by some bugs/idiosyncrasies of the ARXML loader, but also because the current version of the code seems to have some issues with multiplexer messages.  PR coming.

besides this, I noticed that the C code generator does not currently support container messages and that it would probably be quite beneficial if it used a template engine like jinja2 (these are definitely topics for another day, though)

andlaus:
ok, the PR for your PR is ready: https://github.com/dzid26/cantools/pull/1

dzid26:
Should I rebase/squash these fixes reasonably into my existing commits?

andlaus:
both options are fine, so go with whatever you prefer...

dzid26:
I integrated readme and other fixes into these 5 commits.

andlaus:
nice, thanks!

@juleq: please merge this if you don't object.

juleq:
Great, thanks.

PR #248, opened 2020-12-18
decode: allow timestamp

erzoe:
recognize output from e.g. `candump -tA` and `candump -tz`


eerimoq:
Thanks. Please add a test or two.

PR #252, opened 2021-01-04
New subcommand plot

erzoe:
I have added a new subcommand "plot" to visualize the received data using matplotlib, see the additions to the readme and `cantools plot --help`.

I have added two test scripts for this subcommand called tests/test_plot.py and tests/test_plot_without_mock.py.
The first one contains the majority of the tests. It mocks matplotlib and checks that it's functions are called as expected.
The second one contains only a single test which checks for one example that matplotlib does not throw any exceptions and produces the expected output file.


eerimoq:
Nice work!

Please ass matplotlib as a dependency in setup.py as well.

eerimoq:
Part of release 36.1.0, available on PyPI once Github Actions finishes.

PR #255, opened 2021-01-18
Add a CLI flag to disable the snake_case conversion of message and signal names

tjhowse:
We have a use case in which we wish to preserve message and signal names in the original form in the generated C source code. This PR adds a CLI flag that disables the default behaviour of converting all message and signal names to snake_case.

It ads a guard around each use of `camel_to_snake_case()`. If `disable_snake_case_conversion` is true then we only perform the `_canonical()` sanitation. It also changes `self.snake_name` to `self.exported_name` for messages and signals.

A cleaner solution would be to create a `generator` class which stores the settings for the generation, and make all generator functions methods of the new class, rather than passing the new argument into every function in the call tree that includes `camel_to_snake_case`, but that would be a bigger architectural change.

If this is a change that other people are likely to find useful, let me know and I'll finish off this PR. As it stands we'll probably continue to use my fork internally.

- [ ] WIP
- [ ] Add tests
- [ ] Update readme

### DIFF
--- a/cantools/database/can/c_source.py
+++ b/cantools/database/can/c_source.py
@@ -536,9 +536,13 @@ SIGNAL_MEMBER_FMT = '''\
 
 class Signal(object):
 
-    def __init__(self, signal):
+    def __init__(self, signal, disable_snake_case_conversion):
         self._signal = signal
-        self.snake_name = camel_to_snake_case(self.name)
+
+        if disable_snake_case_conversion:
+            self.exported_name = _canonical(self.name)
+        else:
+            self.exported_name = camel_to_snake_case(self.name)
 
     def __getattr__(self, name):
         return getattr(self._signal, name)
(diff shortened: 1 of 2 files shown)


cantools/database/can/c_source.py
```diff
@@ -536,9 +536,13 @@
 
 class Signal(object):
 
-    def __init__(self, signal):
+    def __init__(self, signal, disable_snake_case_conversion):
```
andlaus, line 539: it is probably slightly better to use a "positive" name for the parameter, e.g. `use_raw_names`?

andlaus (commented):
it would probably be better move the stuff which concerns itself with C code generation out of the database part of the codebase into the relevant subparser? (side note: the class names here are quite confusing, IMO.)

besides this, the patch seems to be fine...

simoneruffini:
What's the status of this PR? I'm interested in it being merged

andlaus:
> What's the status of this PR? I'm interested in it being merged

it currently has some merge conflicts with master...

cantools/database/can/c_source.py
```diff
@@ -1540,13 +1569,17 @@ def generate(database,
 
     Set `bit_fields` to ``True`` to generate bit fields in structs.
 
+    Set `disable_snake_case_conversion` to ``True`` to preseve
+    message and signal names from the DBC databse to the extent
```
andlaus, line 1573:
```suggestion
    message and signal names from the database to the extent
```

andlaus:
... and it would be nice if the functionality was featured by a unit test. Besides this, I don't see an reason not to include this functionality. (It looks like this PR was left unmerged because the discussion does not seem to be finished, then it started to suffer from bitrot after a while...)

simoneruffini:
So we are missing unit test and it can be merged?

simoneruffini:
If the original author of the PR doesn't rebase, can someone else do it?

andlaus:
> If the original author of the PR doesn't rebase, can someone else do it?

yes. just make a new PR and mention that it is based on this one...

PR #261, opened 2021-02-01
Handle missing matplotlib more nicely.

tjhowse:
This should've been a part of #260 

My main unknown is if we want to print to console on the `plot` subparser being unavailable, or just silently leave it out?

### DIFF
--- a/cantools/__init__.py
+++ b/cantools/__init__.py
@@ -35,14 +35,18 @@ def _main():
     from .subparsers import dump
     from .subparsers import convert
     from .subparsers import generate_c_source
-    from .subparsers import plot
+
+    try:
+        from .subparsers import plot
+        plot.add_subparser(subparsers)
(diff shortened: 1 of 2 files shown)


eerimoq:
I suggest that we simply move the matplotlib import into `def _do_decode()` in plot.py. That way matplotlib is only imported if the plot command is used.

altendky:
I would tend not to hide the subcommand even when it won't work.  This leads to confusion.  Also, what is the upside of masking the original `ImportError` and raising another made up one instead?

I'll note I made up #265 before noticing this.  Sorry.

AxelVoitier:
@tjhowse, may I propose to consider this PR superseeded by #265 and close it?

tjhowse:
Absolutely!

AxelVoitier:
Thanks

PR #263, opened 2021-02-01
ARXML: Recursively check all AR-PACKAGES for CAN clusters

andlaus:
Currently just the top level packages are checked. I recently stumbled over a file in the wild which needs this atrocity.

This PR is a replacement for https://github.com/eerimoq/cantools/pull/174 and it hopefully completes the refactoring of the ARXML code.

### DIFF
--- a/cantools/database/can/formats/arxml.py
+++ b/cantools/database/can/formats/arxml.py
@@ -101,14 +101,46 @@ class SystemLoader(object):
         messages = []
         version = None
 
-        # This code inspects the top level packages. Packages in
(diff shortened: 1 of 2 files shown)


eerimoq:
This will probably break at some point as there is no test. Should I merge anyway?

andlaus:
> This will probably break at some point as there is no test.

I guess it's a bit unlikely because this code path is excercised even for flat files?

> Should I merge anyway?

Wait for now. I'll modify the ARXML3 test file.

andlaus:
ok. should be good to go now...

PR #268, opened 2021-02-03
Add --output-directory option to generage_c_source command

altendky:
Draft for:
- [x] Actually trying it out (a friend would like this addition)
- [x] A test
- [x] Handle output directory that doesn't exist
- [x] Make it actually work again
- [x] Checking on unwanted effects on the file contents

### DIFF
--- a/cantools/subparsers/generate_c_source.py
+++ b/cantools/subparsers/generate_c_source.py
@@ -31,23 +31,31 @@ def _do_generate_c_source(args):
         not args.no_floating_point_numbers,
         args.bit_fields)
 
-    with open(filename_h, 'w') as fout:
+    path_h = os.path.join(args.output_directory, filename_h)
+    
+    with open(path_h, 'w') as fout:
         fout.write(header)
 
-    with open(filename_c, 'w') as fout:
+    path_c = os.path.join(args.output_directory, filename_c)
+
+    with open(path_c, 'w') as fout:
         fout.write(source)
 
-    print('Successfully generated {} and {}.'.format(filename_h, filename_c))
+    print('Successfully generated {} and {}.'.format(path_h, path_c))
 
     if args.generate_fuzzer:
-        with open(fuzzer_filename_c, 'w') as fout:
+        fuzzer_path_c = os.path.join(args.output_directory, fuzzer_filename_c)
+
+        with open(fuzzer_path_c, 'w') as fout:
             fout.write(fuzzer_source)
+
+        fuzzer_path_mk = os.path.join(args.output_directory, fuzzer_filename_mk)
 
         with open(fuzzer_filename_mk, 'w') as fout:
             fout.write(fuzzer_makefile)
 
-        print('Successfully generated {} and {}.'.format(fuzzer_filename_c,
-                                                         fuzzer_filename_mk))
+        print('Successfully generated {} and {}.'.format(fuzzer_path_c,
+                                                         fuzzer_path_mk))
         print()
         print(
             'Run "make -f {}" to build and run the fuzzer. Requires a'.format(
@@ -81,6 +89,10 @@ def add_subparser(subparsers):
         '-f', '--generate-fuzzer',
         action='store_true',
         help='Also generate fuzzer source code.')
+    generate_c_source_parser.add_argument(
+        '-o', '--output-directory',
+        default='.',
+        help='Directory in which to write output files.')
     generate_c_source_parser.add_argument(
         'infile',
         help='Input database file.')
--- a/cantools/subparsers/generate_c_source.py
+++ b/cantools/subparsers/generate_c_source.py
@@ -1,4 +1,5 @@
 import os
+import os.path
 
 from .. import database
 from ..database.can.c_source import generate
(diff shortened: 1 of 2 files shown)


cantools/subparsers/generate_c_source.py
```diff
@@ -81,6 +92,10 @@ def add_subparser(subparsers):
         '-f', '--generate-fuzzer',
         action='store_true',
         help='Also generate fuzzer source code.')
+    generate_c_source_parser.add_argument(
+        '-o', '--output-directory',
+        default='.',
+        help='Directory in which to write output files.')
```
eerimoq, line 98: Change to `help='Directory in which to write output files (default: %(default)s).')`.

tests/test_command_line.py
```diff
@@ -1156,6 +1156,41 @@ def test_generate_c_source_database_name(self):
...
                                         'tests/files/c_source/' + database_c)
 
+    def test_generate_c_source_output_directory(self):
+        databases = [
+            'motohawk',
+        ]
+        
+        output_directory = 'some_dir'
+
+        for database in databases:
```
eerimoq, line 1166: I suggest that the loop is removed as the test tests a single file.

tests/test_command_line.py
```diff
@@ -1156,6 +1156,41 @@ def test_generate_c_source_database_name(self):
...
+            database_c = os.path.join(output_directory, f'{database}.c')
+
+            if os.path.exists(database_h):
+                os.remove(database_h)
+
+            if os.path.exists(database_c):
+                os.remove(database_c)
+            
+            if os.path.exists(output_directory):
+                os.rmdir(output_directory)
```
eerimoq, line 1184: I suggest replacing all three removes with a single `shutil.rmtree(output_directory, ...)`.

cantools/subparsers/generate_c_source.py
```diff
@@ -1,4 +1,5 @@
 import os
+import os.path
```
eerimoq, line 2: It should be just fine without this import.

eerimoq (commented):
Nice, just a few minor comments.

cantools/subparsers/generate_c_source.py
altendky: 'Fine', yes.  Good, no.  Without importing the module you use you are depending on other code elsewhere having imported it for you.  This means that this file can break without it having made any changes.  In this case, admittedly, `os` itself does import `os.path` for you.  Still not a good habit.

cantools/subparsers/generate_c_source.py
altendky: Sounds like you actually want to update to use the [`ArgumentDefaultsHelpFormatter`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentDefaultsHelpFormatter).

tests/test_command_line.py
altendky: Just copying another test that did the same, but sure.

tests/test_command_line.py
altendky: The actual solution is to use temporary directories everywhere as then you never have to think about removing files to begin with nor if where you are working is clean.

cantools/subparsers/generate_c_source.py
eerimoq: Not the best argument as it is a comment, but: https://github.com/python/cpython/blob/3.9/Lib/os.py#L3-L5

cantools/subparsers/generate_c_source.py
eerimoq: Not in this PR, maybe later.

tests/test_command_line.py
eerimoq: I want to keep the output to manually look at the generated files sometimes. But I agree, all created folders should not be created in the root, but somewhere else. That's for later, if ever.

cantools/subparsers/generate_c_source.py
altendky: Justifying exceptions to good habits is way more hassle than a consistent approach to imports.  You are concerned about the cost of updating a classifier when the unplanned Python 4 is released.  I expect it is much more effort to keep up with which packages promise to export their subpackages and modules not to mention tracking when they change those promises.

cantools/subparsers/generate_c_source.py
altendky: That is what I meant, yes.

altendky: See #275.

tests/test_command_line.py
```diff
@@ -1171,14 +1172,7 @@ def test_generate_c_source_output_directory(self):
...
 
-        if os.path.exists(database_h):
-            os.remove(database_h)
-
-        if os.path.exists(database_c):
-            os.remove(database_c)
-
-        if os.path.exists(output_directory):
-            os.rmdir(output_directory)
+        shutil.rmtree(output_directory)
```
andlaus: hm, this seems quite dangerous and unexpected behavior to me because I suppose most users won't expect the output directory to be deleted. I propose to simply bail out with an error message if the output directory already exists. `git clone` does the same.

altendky: This is in the tests, not the actual command.  Were your concerns related to the running of the tests?

andlaus: ah I see. I just didn't notice this. sorry.

PR #272, opened 2021-02-05
Monitor subcommand: interactivity improvements

andlaus:
This PR mainly improves the interactivity of the `monitor` subcommand to make it feel much more "less-like" and to make editing the filter regular expression much more convenient. Besides this, the last patch changes the instructions in `README.rst` to use `python3 -m cantools` instead of `cantools` on the CLI, because the latter does not always seem to work. Please yell at me if you want to have that patch reverted.


eerimoq:
I agree that it's a good idea to replace `cantools` with `python3 -m cantools`. We should replace `pip install cantools` with `python3 -m pip install cantools` as well. On windows it's usually `py -3 -m cantools`, but can't take all environments into account, so lets stick to `python3 -m ...`.

eerimoq:
I've actually not used cantools myself in quite a while and I can't recall how the monitor user experience was. I trust that your change is good for the majority of the users. I will not try it myself.

andlaus:
ok. fixing the tests was quite elaborate, mainly because I initially had serious issues understanding the error messages. I also did not extend the tests to check for every single new interactive behavior as this would IMO be a bit too excessive.

tests/test_monitor.py
```diff
@@ -705,148 +673,112 @@ def test_reset(self,
...
+                call(1, 0, '   TIMESTAMP  MESSAGE                                           ', 'green'),
+                call(29, 0, 'q: Quit, f: Filter, p: Play/Pause, r: Reset                     ', 'cyan'),
 
                 # Input after reset. 'f' pressed.
                 call(0, 0, 'Received: 1, Discarded: 0, Errors: 0'),
-                call(1,
-                     0,
-                     '   TIMESTAMP  MESSAGE                                           ',
-                     'green'),
+                call(1, 0, '   TIMESTAMP  MESSAGE                                           ', 'green'),
```
eerimoq, line 763: Line too long. Limit is about 85.

andlaus: ok, fixed. thanks. (be aware that this makes the tests a bit harder to grasp and to update.)

PR #275, opened 2021-02-09
Use ArgumentDefaultsHelpFormatter to show defaults in --help

altendky:
It is often nice to have CLI --help output include the default values for parameters that have a default.  `argparse` provides `argparse.ArgumentDefaultsHelpFormatter` to do this automatically.  This provides consistent presentation of the defaults without manual intervention for each parameter or continued maintenance keeping the help text inline with the actual operational code.


eerimoq:
Hmm, this might be good, but I prefer the flexibility of formatting the help text freely instead.

altendky:
Talk about maintenance hassle...

andlaus (commented):
Please resolve the merge conflict (due to the #283, sorry) and add a meaningful description of what the patch does. With that I am in favor of merging it.

altendky:
@andlaus, thanks for the interested.  If you meant you wanted a description added in the OP then that is done.  If you meant something else, let me know. `:]`

PR #265, opened 2021-02-03
lack of matplolib does not error until actually running subcommand

altendky:
Follow up for #259.

PR #280, opened 2021-02-17
Monitor: Avoid failing the entire CLI on missing curses module.

AxelVoitier:
Import-catch philosophy similar to the one used in #265 for matplotlib.

### DIFF
--- a/cantools/__init__.py
+++ b/cantools/__init__.py
@@ -28,8 +28,6 @@ def _main():
                                        dest='subcommand')
     subparsers.required = True
 
-    # Import when used for less dependencies. For example, curses is
-    # not part of all Python builds.
     from .subparsers import decode
     from .subparsers import monitor
     from .subparsers import dump
--- a/cantools/subparsers/monitor.py
+++ b/cantools/subparsers/monitor.py
@@ -1,14 +1,25 @@
 import re
 import time
-import curses
 import bisect
 import queue
+try:
+    # Curses is not necessarily available in all python installations
+    import curses
+except ImportError:
+    curses = None
 
 import can
 from argparse_addons import Integer
-from .. import database
+from .. import database, errors
 from .utils import format_message
 from .utils import format_multiplexed_name
+
+
+class NoCursesError(errors.Error):
+
+    def __init__(self):
+        super().__init__("Your python installation does not provide the curses"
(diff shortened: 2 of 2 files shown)


andlaus:
I agree that a subparser must not to render the others unusable. That said, I suppose the `ncurses` is not the only module which cannot be assumed to be unconditionally available, i.e., it's probably better to catch import errors for all subparsers [here](https://github.com/eerimoq/cantools/blob/master/cantools/__init__.py#L33)?

AxelVoitier:
I was initially going for such a thing. But it was rightly commented in [here](https://github.com/eerimoq/cantools/pull/261#issuecomment-772593175) that it could lead to user confusion if some subcommands were missing.

We could provide alternate subparsers in case of import error. But that seems more work than this solution.

In theory, you could indeed miss any module as someone can customise its python installation as it wishes (even beyond the provided config flags). That would be particularly the case for embedded systems where you may want to aim for a minimal footprint.

Here I think it's more pragmatic to see this on a case by case basis. Curses is one of the standard modules that typically gets disabled. Buildroot has an option for it for instance (that's how it bothered me). I have heard it might also be disabled by some Windows builds?

Another typically disabled standard module is sqlite3. Cantools does not seem to depend on it (as far as my testing about it went some times ago). But it might lurk in a dependency (in particular diskcache).

I gave a look at the other subparsers and could not spot obvious cases of standard modules I know to be typically disabled.

I would say that if someone goes as far as stripping down most of python standard libs but still want to use some cantools CLI subcommands, then maybe they could just patch it locally. Or open an issue/PR for it here.

andlaus:
I see. How much work would it be to implement an "Error subparser" which behaves just like the monitor subparser in this branch if ncurses is not available but with a customizable error message? (e.g., f'Module {subparser_name} not available: {exception}')

> Here I think it's more pragmatic to see this on a case by case basis.

I agree. I still prefer a generic solution if it is not much more work than the whack-a-mole one, though...

andlaus:
@AxelVoitier: Shall I give the generic solution a shot or do you want to try it yourself? If it turns out that the generic approach is too ugly, let's merge this PR as-is...

andlaus:
closing this PR: with the generic solution (#281) merged, this PR is (probably) obsolete. please report back if this is not the case for some reason.

PR #286, opened 2021-03-02
utils: do not decode a integer signal value to float if it exhibits non-fractional scaling and offset

andlaus:
Although using floating point values is probably technically correct for any decoded value, it is IMO nevertheless quite confusing if the signal in question is of integer type. In particular, this applies for the approximately 97% of all cases where no scaling or offset is specified...

### DIFF
--- a/cantools/database/utils.py
+++ b/cantools/database/utils.py
@@ -68,8 +68,15 @@ def _decode_field(field, value, decode_choices, scaling):
         except (KeyError, TypeError):
             pass
 
+    is_int = \
+        lambda x: x is isinstance(x, int) or (isinstance(x, float) and x.is_integer())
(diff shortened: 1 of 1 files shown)


andlaus:
once again, it seems like bean-counting is not a strong point of coveralls: it says "3.6% decreased coverage", but 1) that is unrealistic for such a tiny change, and 2) when inspecting the details of the run, coverage for *all* individual files either stayed the same or went up...

oomzay:
> once again, it seems like bean-counting is not a strong point of coveralls:

Do we (collaborators) have any control over coverals? (It is blocking one of my pushes because coverage dropped 0.3% due to improved test code detecting showstopping issues at the start of a test case,)

PR #293, opened 2021-03-03
CDD: Correct the endianness in the test case

oomzay:
Correcting the endianess assumptions in the CDD test case prior to fixing the implementation.

In CDD file bo='21' is BigEndian, bo='12' is LittleEndian

N.B. **This fix should cause the diagnostic database test to fail**

### DIFF
--- a/tests/test_diagnostics_database.py
+++ b/tests/test_diagnostics_database.py
@@ -9,6 +9,7 @@ class CanToolsDiagnosticsDatabaseTest(unittest.TestCase):
 
     maxDiff = None
 
+    @unittest.expectedFailure  # Current cdd.py implementation is flawed
     def test_example_cdd(self):
         db = cantools.db.load_file('tests/files/cdd/example.cdd',
                                    encoding='iso-8859-1')
@@ -148,6 +149,7 @@ class CanToolsDiagnosticsDatabaseTest(unittest.TestCase):
         decoded = did.decode(encoded)
         self.assertEqual(decoded, decoded_did)
 
+    @unittest.expectedFailure  # Current cdd.py implementation is flawed
     def test_example_cdd_repr(self):
         db = cantools.db.load_file('tests/files/cdd/example.cdd',
                                    encoding='iso-8859-1')
--- a/tests/test_diagnostics_database.py
+++ b/tests/test_diagnostics_database.py
@@ -48,9 +48,9 @@ class CanToolsDiagnosticsDatabaseTest(unittest.TestCase):
 
         data = did.get_data_by_name('Diagnostic_Identification')
         self.assertEqual(data.name, 'Diagnostic_Identification')
-        self.assertEqual(data.start, 64)
+        self.assertEqual(data.start, 71)
         self.assertEqual(data.length, 16)
-        self.assertEqual(data.byte_order, 'little_endian')
+        self.assertEqual(data.byte_order, 'big_endian')
         self.assertEqual(data.scale, 1)
         self.assertEqual(data.offset, 0)
         self.assertEqual(data.minimum, 0)
@@ -65,7 +65,7 @@ class CanToolsDiagnosticsDatabaseTest(unittest.TestCase):
             'Ident_Number_1_0': 0x3456,
             'Diagnostic_Identification': 0xabcd
         }
-        encoded_did = b'\x34\x12\x78\x56\x12\x90\x56\x34\xcd\xab'
+        encoded_did = b'\x12\x34\x56\x78\x90\x12\x34\x56\xab\xcd'
 
         encoded = did.encode(decoded_did)
         self.assertEqual(encoded, encoded_did)
@@ -141,7 +141,7 @@ class CanToolsDiagnosticsDatabaseTest(unittest.TestCase):
             'Vehicle_type': 'Sedan',
             'Special_setting': 3
         }
-        encoded_did = b'\x21\x03'
+        encoded_did = b'\x12\x03'
 
         encoded = did.encode(decoded_did)
         self.assertEqual(encoded, encoded_did)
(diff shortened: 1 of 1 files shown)


tests/test_diagnostics_database.py
```diff
@@ -48,9 +48,9 @@ def test_example_cdd(self):
 
         data = did.get_data_by_name('Diagnostic_Identification')
         self.assertEqual(data.name, 'Diagnostic_Identification')
-        self.assertEqual(data.start, 64)
+        self.assertEqual(data.start, 71)
```
oomzay, line 51: Note to other reviewers: For consistency with the DBC items, and compatability with the shared message codecs, using the same saw-tooth startbit numbering convention is necessary.

tests/test_diagnostics_database.py
```diff
@@ -48,9 +48,9 @@ def test_example_cdd(self):
...
         self.assertEqual(data.length, 16)
-        self.assertEqual(data.byte_order, 'little_endian')
+        self.assertEqual(data.byte_order, 'big_endian')
```
oomzay, line 53: Note to other reviewers: Diagnostic_Identification has datatype HexDump_2Byte. This has bo='21' which means big_endian.

oomzay:
As expected cantools now fails the diagnostic database test. 

Good, _This is the goal of this fix._

tests/test_diagnostics_database.py
```diff
@@ -141,7 +141,7 @@ def test_example_cdd(self):
             'Vehicle_type': 'Sedan',
             'Special_setting': 3
         }
-        encoded_did = b'\x21\x03'
+        encoded_did = b'\x12\x03'
```
andlaus, line 144: I can't judge the patch itself, but this line seems to be _nibble_ order, not _byte_ order?

andlaus (commented):
I don't have any expertise with the CDD stuff, so I can't say anything about the correctness of this patch. its aesthetics of it seem fine to me, but the test suite needs be adapted to make pass before merge.

tests/test_diagnostics_database.py
```diff
@@ -151,95 +151,95 @@ def test_example_cdd(self):
...
-            "did('InputOutput', 0x0080)\n"
-            "  data('Door_contact_front_left', 0, 1, 'little_endian', 1, 0, 1, 1, 'None', {0: 'closed', 1: 'open'})\n"
-            "  data('Door_contact_front_right', 1, 1, 'little_endian', 1, 0, 1, 1, 'None', {0: 'closed', 1: 'open'})\n"
-            "  data('Door_contact_rear_left', 2, 1, 'little_endian', 1, 0, 1, 1, 'None', {0: 'closed', 1: 'open'})\n"
-            "  data('Door_contact_rear_right', 3, 1, 'little_endian', 1, 0, 1, 1, 'None', {0: 'closed', 1: 'open'})\n"
-            "  data('_reserved', 4, 4, 'little_endian', 1, 0, 0, 255, 'None', None)\n")
+
+        self.assertEqual(repr(db), '''did('DEFAULT_SESSION', 0x0081)
+
+did('ProgrammingSession', 0x0085)
```
andlaus, line 157: this (and the following stuff) seems to be a bit fishy to me: shouldn't it stay part of a test case?

tests/test_diagnostics_database.py
```diff
@@ -151,95 +151,95 @@ def test_example_cdd(self):
...
-            "  data('Special_setting', 8, 8, 'little_endian', 1, 0, 1, 1, 'None', None)\n"
-            "\n"
-            "did('InputOutput', 0x0080)\n"
-            "  data('Door_contact_front_left', 0, 1, 'little_endian', 1, 0, 1, 1, 'None', {0: 'closed', 1: 'open'})\n"
-            "  data('Door_contact_front_right', 1, 1, 'little_endian', 1, 0, 1, 1, 'None', {0: 'closed', 1: 'open'})\n"
-            "  data('Door_contact_rear_left', 2, 1, 'little_endian', 1, 0, 1, 1, 'None', {0: 'closed', 1: 'open'})\n"
-            "  data('Door_contact_rear_right', 3, 1, 'little_endian', 1, 0, 1, 1, 'None', {0: 'closed', 1: 'open'})\n"
-            "  data('_reserved', 4, 4, 'little_endian', 1, 0, 0, 255, 'None', None)\n")
+
+        self.assertEqual(repr(db), '''did('DEFAULT_SESSION', 0x0081)
```
andlaus, line 155: leaving the `repr(db)` statement and the beginning of the string on separate lines probably makes the changes easier to compare on github because (hopefully) the lines will align then...

oomzay:
> I don't have any expertise with the CDD stuff, so I can't say anything about the correctness of this patch. its aesthetics seem to be fine to me, but the test suite needs be adapted to make it pass as a precondition to merging.

Please advice me how to deal with this. The existing CDD module under test is now failing this test **because this test for it has been fixed.**. It is my usual practice to write, or in this case fix, the test cases before fixing the code that they test.

I really want to make it very visible, and reviewed by CDD user/expert, that I consider the current CDD (and this Test for it) to be broken before I push my actual interface-changing fixes to the CDD module (which I do in fact have ready).

I could push my fix for the CDD module as well which would cause the new Test to pass - but that risks a superficial review.

tests/test_diagnostics_database.py
oomzay: This field position difference is a result of the different start-bit convention used by the codecs for fields that are declared big_endian vs little_endian regardless of their bit length. Because the current CDD loader mis-identifes the fields as little_endian the codecs are also misinterpreting the field start positions. (Compounded by the fact there is currently no mapping at all from the CDD start bit convention to the codec./DBC convention in the current CDD module)

tests/test_diagnostics_database.py
oomzay: Agreed. I'll match the formatting.

oomzay (commented):
Unless otherwise directed I will tag the tests as @unittest.expectedFailure.

simontegelid:
You can just push a commit that resolves the failing tests, then it can be merged. Not all commits need to pass the tests, just the immediate parent(s) to the merge commit later on.

oomzay:
> You can just push a commit that resolves the failing tests, then it can be merged. Not all commits need to pass the tests, just the immediate parent(s) to the merge commit later on.

On principal I would much rather submit and merge the improved test cases independantly of any fixes that I or someone else might do to address the failures that the improved test cases reveal.

Is their any objection to tagging this as an expectedFailure? This request will then pass CI checks and can can be submitted in its own right, independant of code fixes.

oomzay:
> Coverage decreased (-0.3%) to 96.296% when pulling **[a1423e7](https://github.com/eerimoq/cantools/commit/a1423e7fd65866c875af87a1d846713b60029a2b) on oomzay:master** into **[2f66e18](https://github.com/eerimoq/cantools/commit/2f66e185eee915ddd2bf9ef1672515c712e9171b) on eerimoq:master**.

Coveralls is vetoing this request. After a bit of digging it seems the coverage has dropped 0.3% because the test is now (correctly) bailing out early when it discovers an error condition.

Any advise on how to deal with this?

oomzay:
> Coveralls is vetoing this request. ... Any advise on how to deal with this?

So it seems we can choose to ignore coverals. But I'll wait for someone else to give the go ahead. Then I'll push the actual CDD fixes that I have ready in the wings.

oomzay:
Change of heart: I have marked this pull request down to draft status as I am having seconds thoughts about imposing the rather insane DBC-style start-bit convention on the DID objects at the time they are imported just to suit the codecs (implementation detail).

Although that specification "works" (I coded and tested it) I now think it cleaner to hide this mapping and just do it when creating the codecs and thereby preserve the more sensible CDD start-bit convention in the DID objects.

andlaus:
> Coveralls is vetoing this request. After a bit of digging it seems the coverage has dropped 0.3% because the test is now 
> (correctly) bailing out early when it discovers an error condition.
>
> Any advise on how to deal with this?

the test suite should cover the correct *as well* as the incorrect cases. My personal opinion with this regard is that some exotic error cases can be left untested if the amount of untested code is minimal (e.g. `if error_condition: return`)  and a test would require a disproportionate amount of work (e.g., addin an extra data file).

IMO small changes in coveralls coverage should (usually) not be an obstacle for merging: First, some of the error handling code is very inconvenient to test and second, coveralls seems to have issues doing statistics: it sometimes happens that it says the overall coverage is down while the coverage of every individual file stays at least identical.

In this particular case, it would be nice if the tests could be extended to increase coverage of the files `cantools/database/diagnostics/did.py` (in particular 'encode()`) and `cantools/database/diagnostics/data.py` (various new features?).

oomzay:
> In this particular case, it would be nice if the tests could be extended to increase coverage of the files `cantools/database/diagnostics/did.py` (in particular `encode()`) and `cantools/database/diagnostics/data.py` (various new features?).

@andlaus: Thankyou for your detailed advice. I am new here and a bit nervous about messing things up!

I limited myself to fixing tests that were incorrect initially. Test coverage should increase when I submit my changes to fix the module under test and some additional test cases I have created.  On that subject... 

I now have the fixes to this test case, and the cdd fixes under test and some more additional cases on [my fix-cdd-endianness branch](https://github.com/oomzay/cantools/tree/fix-cdd-endianness).  So I will ask for another bit of advice!...

Should I abandon this pull request and start another from that branch, or should I fast-forward my master branch which will presumably update this pull request and combine the changes to the test case and the cdd fixes?

(I would prefer one step at a time, in sequence, but I have to support local demands as well!)

andlaus:
> I am new here and a bit nervous about messing things up!

no worries, I had exactly the same problem for my initial contribution ;)

> Should I abandon this pull request and start another from that branch, or should I fast-forward my master branch which will presumably update this pull request and combine the changes to the test case and the cdd fixes?

I guess it does not really matter. Personally, I would rebase the new version on top of the upstream's master branch (`git rebase origin/master` if the remote of the upstream repo is `origin`) and once everything works as expected locally, force push to the branch of the PR that is already open (`git push -f oomzay master`, if the remote of your personal fork is called `oomzay`). I'd also advice that you use other branch names than 'master' for PRs. (`git checkout -b $BRANCH_NAME`)

oomzay:
I am going to abandon this request and open a new one which includes the proposed fixes as well as the fixed tests.

PR #296, opened 2021-03-09
Fix CDD endianness

oomzay:
This fixes the intepretation of byte order codes in .CDD files. By reference to files generated by CANdela. bo='21' is BigEndian, bo='12' is LittleEndian.

This fix also ensures that bigendian fields/datas are presented in network order to the codec generator. Previously this was just assumed, but was not the case for fields/datas parsed from CDD files.


cantools/database/diagnostics/formats/cdd.py
```diff
@@ -98,7 +99,11 @@ def _load_data_types(ecu_doc):
                 LOGGER.debug("Ignoring unsupported attribute '%s'.", key)
 
         if ctype.attrib['bo'] == '21':
+            byte_order = 'big_endian'
+        elif ctype.attrib['bo'] == '12':
```
andlaus, line 103: doesn't this switch the default behavior? if yes, do you know why it worked for some file before?

andlaus (approved):
(sorry for the delay, I missed that you requested a review when you opened the PR, and I only went over all open PRs today.)

The code looks good to me, but this is probably not worth too much: I never used the CDD file format so far, nor do I have any idea how it works...

cantools/database/diagnostics/formats/cdd.py
oomzay: >doesn't this switch the default behavior? 

Yes.

> if yes, do you know why it worked for some file before?

It didn't really work before. The test cases passed because they made the same false assumption about endianness as the parser. I have now added files and test cases for both BigEndian and LittleEndian properties.

cantools/database/utils.py
```diff
@@ -205,3 +209,26 @@ def create_little():
...
+
+def cdd_offset_to_dbc_start_bit(cdd_offset, bit_length, byte_order):
+    '''Convert CDD/c-style field bit offset to DBC field start bit convention.
+
+    BigEndian (BE) fields are located by their MSBit's sawtooth index.
+    LitteleEndian (LE) fields located by their LSBit's sawtooth index.
+    '''
+    if byte_order == "big_endian":
+        # Note: Allow for BE fields that are smaller or larger than 8 bits.
+        return (8 * (cdd_offset // 8)) + min(7, (cdd_offset % 8) + bit_length - 1)
```
tiantian1645, line 232: I use Vector dbc file and find the start bit is wrong with Motorola Big End.

Start bit in motorola signals from Vector dbc files

https://github.com/reinzor/libcan-encode-decode/issues/1

https://github.com/ebroecker/canmatrix/wiki/signal-Byteorder
```python
def itom_sb(b, length):
    b = b - (b % 8) + 7 - (b % 8)
    b = b + length - 1
    b = b - (b % 8) + 7 - (b % 8)
    return b

>>> itom_sb(7, 32) == 24
```

juleq: Hi there. You are posting to a PR that refers to CCD file format. Do you use CCD?

PR #308, opened 2021-05-17
Make parsers work with candump -a switch

ftab:
I found myself working on a CAN system that sends bits of ASCII every now and then, so it was helpful to use the `candump -a` switch.

This PR adds a few parsers compatible with the default (non-timestamped) and timestamped (-t a / -t z / -t A) outputs when using the -a switch. By compatible, I mean they don't do anything with it (since you can just look at it yourself in the data array), but they parse the rest of the line correctly.

That way, cantools decode can be used in conjunction with candump -a

### DIFF
--- a/cantools/logreader.py
+++ b/cantools/logreader.py
@@ -63,6 +63,24 @@ class CandumpDefaultPattern(BasePattern):
         return DataFrame(channel=channel, frame_id=frame_id, data=data, timestamp=timestamp, timestamp_format=timestamp_format)
 
 
+class CandumpDefaultAsciiPattern(BasePattern):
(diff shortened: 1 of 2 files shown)


simontegelid:
I would suggest that you just added `.*?` to the end of the existing regexes than adding duplicate classes (`CandumpDefaultPattern`, `CandumpTimestampedPattern` and `CandumpAbsoluteLogPattern`). See https://regex101.com/r/Fbc9vd/1

ftab:
I like that idea. I wish I'd thought of that. Sometimes my regex fu is lacking...glitch in the matrix or something. I will amend.

ftab:
Got it

PR #316, opened 2021-07-06
attempt to fix the recently appearing test failures of PRs

andlaus:
the assumption is of this PR that these errors are caused because github-CI recently removed gcc-8 and -- maybe -- clang-8.

While I see the point of using older compilers to ensure that they stay supported, the code to be compiled is based on the C99 standard which ought to be very seasoned for at least a decade, i.e., I would be *very* surprised if compilers introduced major incompatibilities with this code in the future. IOW, let's just use the default versions of the host system...

Signed-off-by: Andreas Lauser <andreas.lauser@mbition.io>
Signed-off-by: Oliver Kopp <oliver.kopp@mbition.io>

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -29,9 +29,9 @@ jobs:
         make test-c-clean
         make test-c
         make test-c-clean
-        CC=gcc-8 make test-c
+        CC=gcc make test-c
         make test-c-clean
-        CC=clang-8 make test-c-src
+        CC=clang make test-c-src
     - name: Coveralls
       uses: AndreMiras/coveralls-python-action@develop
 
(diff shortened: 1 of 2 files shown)


simontegelid:
+1 for merging this.

juleq:
I assume this concerns the tests that check generated c code output.

As far as I can see there is no CFLAG in test.mk that would insist on c99 input. If the generator code changes in the future, non c99 language elements could be introduced that the host compiler will not complain about. I think that is probably a minor concern but adding 
`CFLAGS += -std=c99`
to enforce c99 probably will not hurt and makes the selected standard more explicit. Opinions?

andlaus:
good point. Fixed, thanks for the catch!

PR #317, opened 2021-07-06
Initial value calculation according to Vector CANdb++

TheGreatChris:
Bugfix for issue described here: https://github.com/eerimoq/cantools/issues/251

### DIFF
--- a/cantools/tester.py
+++ b/cantools/tester.py
@@ -1,7 +1,6 @@
 # The tester module.
 
 import time
-import decimal
 
 try:
     from collections import UserDict
@@ -214,7 +213,7 @@ class Message(UserDict, object):
             maximum = 0 if not signal.maximum else signal.maximum
             if signal.initial:
                 # use initial signal value (if set)
-                initial_sig_values[signal.name] = (signal.initial * decimal.Decimal(signal.scale)) + decimal.Decimal(signal.offset)
+                initial_sig_values[signal.name] = (signal.initial * signal.decimal.scale) + signal.decimal.offset
             elif minimum <= 0 <= maximum:
                 # use 0 if in allowed range
                 initial_sig_values[signal.name] = 0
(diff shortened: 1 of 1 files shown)


andlaus (approved):
looks good to me. (I'm always a bit confused when the physical and when the "raw" value ought to be stored in the database...)

s3bastiand:
Hi, I have tried the fix and it only partially worked for me. I encountered a new issue after I have applied it:
TypeError: unsupported operand type(s) for *: 'decimal.Decimal' and 'float' "cantools"

Change the same line to:
initial_sig_values[signal.name] = (decimal.Decimal(signal.initial) * decimal.Decimal(signal.scale)) + decimal.Decimal(signal.offset)

solved my problem, but I'm not sure it's correct...

juleq:
I have not looked into the details but I noticed the following:

1.) The signal class has an (optional) copy of min/max/scale/offset as decimal type for lossless calculations.

2.) You could inspect utils.py to understand the handling there:
’’’
def _encode_field(field, data, scaling)
’’’

TheGreatChris:
> 
> 
> Hi, I have tried the fix and it only partially worked for me. I encountered a new issue after I have applied it:
> TypeError: unsupported operand type(s) for *: 'decimal.Decimal' and 'float' "cantools"
> 
> Change the same line to:
> initial_sig_values[signal.name] = (decimal.Decimal(signal.initial) * decimal.Decimal(signal.scale)) + decimal.Decimal(signal.offset)
> 
> solved my problem, but I'm not sure it's correct...

Please elaborate a bit more how the signal definition looks of the DBC that produced your described error. I can't reproduce the issue and neither can I fully comprehend the decimal classes supposed usage.

TheGreatChris:
I could do a bit of debugging and I was able reproduce the issue:
initial is always of type Decimal while scale and offset is of type float.
It seems that Decimal class only fails to multiply certain floats. I guess if the float can't be converted into a Decimal. For example Decimal('1') * 100 will work but Decimal('1') * 0.1 will fail.
I agree with s3bastiand's proposal to convert offset and scale to a Decimal before doing init value calculation.

TheGreatChris:
Using already present decimal values for lossless calculations as mentioned by @juleq.

juleq:
Looking good. Thank you for fixing this, @TheGreatChris.

PR #319, opened 2021-07-16
add 'list' subparser

andlaus:
in principle, this subparser does the same as 'dump', but it does not use ASCII art, so it is much easier to extend and the result can be grep'ed much more easily by shell scripts.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -255,6 +255,32 @@ Dump given database in a human readable format:
 
      ------------------------------------------------------------------------
 
+The list subcommand
+^^^^^^^^^^^^^^^^^^^
+
+Print all information of a given database in a human readable
+format. This is very similar to the "dump" subcommand, but the output
+is less pretty, slightly more comprehensive and easier to parse by
+shell scripts:
+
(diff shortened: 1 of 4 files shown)


juleq:
I assume since this is an addition and thus also ready to merge?

andlaus:
yes, it is IMO ready: this tool does not introduce any dependencies and should not break anything for anybody who is not using it...

juleq:
Ahh. Since the choices PR is in this collapsed to a more manageable change set :).

andlaus:
thanks for merging :)

juleq:
Do you release a new version? I cannot find the place where Eric explained how to do that.

andlaus:
I would like to, but I'm currently handcuffed by my emploer's own internal regulations. As far as I remember he said that you just need to create a new tag (cf. https://github.com/eerimoq/cantools/tags ) and everything else happens automatically. Since there are quite a few changes in the code since the last release I propose to make a slightly larger jump and create a `37.0.0` tag...

PR #320, opened 2021-07-16
implement comments for named signal values

andlaus:
Some file formats (err, ARXML) support this, and I occasionally consider these additional explanations to be useful.

This patch works by introducing a new class called `NamedSignalValue` and the signal choices dictionaries map to objects of it instead of raw strings. In most respects `NamedSignalValue` objects behave just like strings, but in addition they exhibit a `.comments` attribute which maps a language name to the comment für this language (just like the comments for `Signal` and `Message` objects).

### DIFF
--- a/cantools/database/can/c_source.py
+++ b/cantools/database/can/c_source.py
@@ -607,7 +607,7 @@ class Signal(object):
         """
 
         items = {
-            value: camel_to_snake_case(name).upper()
+            value: camel_to_snake_case(str(name)).upper()
             for value, name in self.choices.items()
         }
         names = list(items.values())
@@ -1162,7 +1162,7 @@ def _format_choices(signal, signal_name):
             fmt = '{signal_name}_{name}_CHOICE ({value}u)'
 
         choices.append(fmt.format(signal_name=signal_name.upper(),
-                                  name=name,
+                                  name=str(name),
                                   value=value))
 
     return choices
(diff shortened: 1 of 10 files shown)


andlaus:
note that #319 currently depends on this and thus the tests fail (besides the fact that #316 is not yet merged). Also, it is pretty trivial to adapt #320 to not print the comments of these enums, but then it would need to be added back later...

juleq:
As I understand it this offers the option to add more prose to a choice. A choice does already consists of a value and a value description (I think in case of dbc that is limited to 32 chars by default in Vector tools). This change adds a comment field where further information can be added. Correct so far?

I think its a good addition and it does not get in once way too much if not being used.

One question: You chose dict instead of previous odict. I assume that odict has been used before to retain order when dumping. I vaguely remember that Pyhon starts to change order of dicts if the key is a number. I assume you could change back to dict because that  is not true for objects?

andlaus:
> Correct so far?

perfect.

> I vaguely remember that Pyhon starts to change order of dicts if the key is a number. I assume you could change back to dict because that is not true for objects?

good point. IIRC I changed this back and forth during development a few times because I thought the odict caused some issues which in the end turned out to be unrelated. I just forgot to revert this part of the code. I'll hopefully fix this tomorrow, thanks for the catch...

andlaus:
alright: all dicts should be orderdered dicts again.

Issue #323, opened 2021-07-17
Incorrect DBC format


kartikarcot:
https://github.com/eerimoq/cantools/blob/1bf3e850ddd79eced42548722fe758e9f29ee675/cantools/database/can/formats/dbc.py#L852

I think this has to be just the signal name and not the entire object signature. i.e.
it has to be `signal.multiplexer_signal.name` and not `signal.multiplexer_signal`

juleq:
I think you are probably right.

If so, ext multiplexing seems to be lacking test coverage as it slipped through.

I have never used it myself before because Vector requires additional licensing for tool support and so I usually try to stay away.

Would you care to do a PR that fixes it and adds a test that exercises the extended multiplex IO?

andlaus:
ok, judging from the comments for #326, it looks like this issue is basically a misunderstanding. I'll close this issue for now, if I did get something wrong, feel free to re-open it...

PR #326, opened 2021-07-19
fixed minor bug in dbc file generation for multiplexing

kartikarcot:
This is in reference to issue #323. I do not know what kind of tests would be suitable for this since I am myself beginning to learn about CAN and DBC. I also do not have access to Vector Tools. Let me know how to go about it @eerimoq

### DIFF
--- a/cantools/database/can/formats/dbc.py
+++ b/cantools/database/can/formats/dbc.py
@@ -849,7 +849,7 @@ def _dump_signal_mux_values(database):
                 'SG_MUL_VAL_ {frame_id} {name} {multiplexer} {ranges};'.format(
                     frame_id=get_dbc_frame_id(message),
                     name=signal.name,
-                    multiplexer=signal.multiplexer_signal,
+                    multiplexer=signal.multiplexer_signal.name,
                     ranges=ranges))
 
     return sig_mux_values


juleq:
Have you checked that multiplexer_signal indeed does contain a signal object and not just a signal name?

I do not have the time to go into the details but the failing tests leave the impression that
1.) this is already covered by tests and
2.) these tests are now failing because it already was a signal name.

You could step through with the python debugger to see whats going on. I would propose just loading and dumping a dbc again and break on the line that you have changed to inspect the contents of multiplexer_signal.

If you find it to be a name already I'd say that changing the misleading name to "multiplexer_name" or "multiplexer_signal_name" might be the only thing left to propose.

kartikarcot:
Yea looks like this is due to a misunderstanding on my part. I interpreted your documentation about `multiplexer_signal` argument in the Signal constructor call to expect an object, whereas I see that you pass strings instead. Could be a better idea to pass the object since there would be consistency of name.

andlaus:
@kartikarcot: is this PR still relevant? i.e. is the current behavior wrong and we should wait for you to fix the tests or ought this PR to be closed?

juleq:
I´d say this was just a naming confusion. Will close for now. If anyone desires to rename to e.g. multiplexer_signal_name, this PR can be reopened.

PR #331, opened 2021-08-12
RFC (not ready to be merged) : c code gen: create convenience wrappers

v0lker:
the code generation is very useful but doesn't go far enough for my taste. for example, when sending a message, one still has to type in the invocations of the range check and encoding for each signal individually, assign the values to a struct, then call `*_pack()` to output that struct into the buffer that one actually needs.

proposal (assuming one message with two signals, sig1 and sig2):
`*_wrap_pack()` creates a message and writes it into a buffer, e. g.:
```
db_msg_wrap_pack(u8 *outbuf, size_t size, double sig1, double sig2)
{
        struct db_msg_t msg;
        msg.sig1 = msg_db_sig1_encode(sig1);
        msg.sig2 = msg_db_sig2_encode(sig2);

        int ret = db_msg_check_ranges(&msg);
        if(ret) return ret;

        ret = db_msg_pack(outbuf, &msg, size);
        if (<expected msg_length> != ret) return -1;

        return 0;
}
```
`*_check_ranges()` checks the ranges of all signals and returns the 1-based index of the first one that fails the check.

`*_wrap_unpack()` reads a message from a buffer and retrieves the contents like so:
```
int db_msg_wrap_unpack(u8 *inbuf, size_t size, double *sig1, double *sig2)
{
        struct db_msg_t msg;

        if (db_msg_unpack(&msg, inbuf, size) return -1;

        int ret = db_msg_check_ranges(&msg);
        if(ret) return ret;

        if(sig1) *sig1 = db_msg_sig1_decode(msg.sig1);
        if(sig2) *sig2 = db_msg_sig2_decode(msg.sig2);

        return 0;
}
```

range check in `wrap_unpack()` is still missing, i just noticed, and in some places the newlines are wrong, but that's basically the idea. i'm still testing this a bit more thoroughly and will address the open issues shortly...

any thoughts?


andlaus:
hm, I'm not opposed to this per-se, but I suppose the cantools ought to be split into a "core" and "fringe" tools in the future where the code generators would be "finge tools" along with the plotting, monitoring and convertion stuff. This would have the benefit that multiple template based code generation mechanisms could be easily added without potentially breaking stuff for somebody else...

v0lker:
ok, so - for now i'll just try and make the CI tests pass (they fail because the expected code gen output is compared as a string with the actual one, so the new functions are unexpected..), hope that's ok (?)

andlaus:
> hope that's ok (?)

no problem. I just wanted to inform you that the chances of getting something like this are rather slim with the current monolithic repository (IOW, don't spend an excessive amount of time on it)...

v0lker:
up to you (obviously). i don't really have time to split the repository up, though.

v0lker:
i think i've sorted out all the failing tests now, even the `test-c` which only seem to be run some times(?).

the generated multiplex-related code threw a warning (and thus test failure) because in the receive (unpack) path, the parts of the message struct were only initialised conditionally, while the range check that i'm generating always runs over all elements.

properly (conditionally) checking all the relevant signals seems the best way forward but i don't have any use for it right now and you already indicated that you're not very inclined taking the patch in the first place, so i'll work around it by memset()ting the whole struct to 0.

now - 0 might not actually be in the valid range for a given signal, but the range check on receive (`..._unpack()`) is not fatal, (the downside is that the false positive range check might mask out an interesting failed range check, though).

andlaus:
> now - 0 might not actually be in the valid range for a given signal,

You could add a [` VALGRIND_MAKE_MEM_UNDEFINED` valgrind client request](https://valgrind.org/docs/manual/mc-manual.html#mc-manual.clientreqs) to mark the object as not fit to be used after the memset. this way, accessing improperly initialized signals will at least get detected when running running the code via `valgrind`...

PR #335, opened 2021-08-18
Define is_int only when it is needed

JulianWgs:
is_int is only needed when scaling is True, otherwise it is defined and never used.

Also I deleted the `x is` which I believe to be a bug, because the function does not give the correct output for an int e.g 2.

I'm not sure whether `if scaling` should `elif scaling`

### DIFF
--- a/cantools/database/utils.py
+++ b/cantools/database/utils.py
@@ -68,9 +68,9 @@ def _decode_field(field, value, decode_choices, scaling):
         except (KeyError, TypeError):
             pass
 
-    is_int = \
-        lambda x: x is isinstance(x, int) or (isinstance(x, float) and x.is_integer())
     if scaling:
+        is_int = \
+            lambda x: isinstance(x, int) or (isinstance(x, float) and x.is_integer())
         if field.is_float \
            or not is_int(field.scale) \
            or not is_int(field.offset):


juleq:
I would agree on both accounts. Thanks.

JulianWgs:
What about the elif scaling?

juleq:
You mean elif to decode_choices, pass branch?

juleq:
I see no problem here.

1.) You fixed the lambda.
2.) You narrowed its scope.
3.) You did not change logic.

The pass above seems to be a deliberate decision to return using the regular code path if a choice cannot be decoded. Otherwise nothing would be returned.

JulianWgs:
Okay something is up with the decode_choices and scaling. In the example given in issue https://github.com/eerimoq/cantools/issues/271 both are set to True. This would produce different results when refactoring to elif (also replacing pass with return value).

JulianWgs:
Looking at cantools.database.can.database.Database.decode_message both decode_choices and scaling default values are True. At least in my code I keep it that way. Therefore both paths in _decode_field are always executed. @eerimoq Do you know why it is implemented this way? Couldn't both values not inferred by the message?

juleq:
I am not sure I get your point. If decode_choices is true, the function will return when the choice could be resolved. Could you rephrase your point?

juleq:
As a background: decode_choices and scaling are both options for the user to alter the decoding behavior. If decode_choices is set to false, cantools will not resolve choices to their description. If scaling is set to false, cantools will not apply the raw->physical translation but keep it raw instead.

Since both are true, the default behavior is: Return the choice if this field has choices associated with it and one matches. If not, do business as usual and return the physical value.

PR #336, opened 2021-08-18
update version to 36.3.1

andlaus:
this PR is very simple: it changes the in-code version to 36.3.1. The fact that this was forgotten (rather: unknown to anyone involved), leads PyPI to incorrectly announce the package version there to be 36.2.0 while in fact it is the new and shiny 36.3.0.

@juleq: Could you please create a "36.3.1" tag after merging this?


juleq:
I see. So there was a step two :).

andlaus:
> I see. So there was a step two :).

We all learn this by trial and failure ;)

juleq:
I have tagged 36.3.1 now but the version file seems to have 36.4.0. Care for a 36.3.2?

andlaus:
hm, looks like something went slightly wrong: besides you merging the PR definitely about a minute after I force-pushed the version change to be  36.3.1 instead of 36.4.0, github still merged the old version of the branch. This is why PyPI thinks its 36.4.0.

I see three solutions:

- do nothing, the new version on PyPI is higher than the old, so `pip3 update` will work
- leave the repo as-is but simply create a `36.4.0` tag
- make a new PR, this time I try harder not to screw up

I'm okay with any of these, but I think I have a tiny preference for the 36.4.0 tag...

juleq:
Ok, will do one.

juleq:
Looks better now. I did delete the 36.3.1, but trace elements seem to remain visible in the release timeline. We might have to lower our aesthetic expectations there :).

andlaus:
Nice! What is a release timeline anyway? (at least nothing went years behind schedule ;))

PR #338, opened 2021-08-19
Update control flow in _decode_field 

JulianWgs:
This is an update to the discussion at https://github.com/eerimoq/cantools/pull/335.

- before the update choices was always applied to the value, if it
succeeded, it would return, but if it failed scaling would be applied. I
believe this behavior to be inconsistent.
- This code passes all, but one test (cantools/tests/test_database.py", line 1266, in test_the_homer_encode_decode_choice_scaling self.assertEqual(decoded, decoded_message)).
- The code has different behavior from the previous code, when a signal
possesses choices, but also scaling. I dont know about the official CAN
spec, but this kind of message should not exist. Either choices or
scaling is used, but not both. To prevent this behavior one could add a
check to the database, which can be run mandatory or optionally.
- I also reformatted the code with black as I'm not a fan of the \ syntax
- This implementation should also provide a small performance boost as fewer lines of code will be executed

TODO:
- [ ] Compare to the CAN spec or to Vector software


juleq:
You are misunderstanding scaling. Scaling for a choice value is usually offset=0, factor=1 but _it does exist_.

JulianWgs:
I understand that they are the standard values, but is it a actual use case to have a not matching choice and then scaling the value (scaling != 1 or offset !=0)? I cant come up with a sensible use case for this.

I havent looked into the spec and dont have any vector software installed. How do they handle the case/dbc?

juleq:
With respect to the backslash formatting: @eerimoq as the BDFL did a good job demanding consistent formatting throughout the code base. I would argue that good citizenship would include adherence in that regard.

JulianWgs:
> With respect to the backslash formatting: @eerimoq as the BDFL did a good job demanding consistent formatting throughout the code base. I would argue that good citizenship would include adherence in that regard.

I'm fine also with that :)

juleq:
Did you get my other point? Scaling is almost always true. Only in a very specific circumstance, when the user wants to have the raw value, he can opt to set scaling to false.

So you are right in that for choices the translation raw->physical does make no sense, but that does not change control flow. It is taken care of by offset=0 and factor=1.

juleq:
Vector does it the same. The spec tells you that value tables deliver descriptions for raw values. But nothing would stop you from defining a factor of 2 even if that makes no sense.

andlaus:
> Did you get my other point? Scaling is almost always true. Only in a very specific circumstance, when the user wants to have the raw value, he can opt to set scaling to false.

actually, I've got quite a few files which exhibit this: For example the coolant temperature might be transmitted using an 8 bit integer, where the offset might be something like -20.0,  the scaling 0.5 and a transmitted value of 0 might indicate "lower than -20 Celsius" and 255 "higher than 107 degrees". I thus think that the current behavior is correct...

PR #345, opened 2021-09-22
generate functions to clamp values between min and max

v0lker:
technically, this does not depend on the first commit in this branch, which is a separate PR (https://github.com/eerimoq/cantools/pull/331), but it's a hassle to keep the changes in the tests separate. if someone likes this but not the first commit, i can pluck them apart - let me know.


drewr95:
I commented in the other PR [346](https://github.com/cantools/cantools/pull/346) but are doubles the only supported type for clamping?

v0lker:
> I commented in the other PR [346](https://github.com/cantools/cantools/pull/346) but are doubles the only supported type for clamping?

answered on that other PR ...

PR #346, opened 2021-09-22
c code gen: a function that returns whether a certain frame id is extended or not

v0lker:
this does not technically depend on the other PRs' (https://github.com/eerimoq/cantools/pull/331 and https://github.com/eerimoq/cantools/pull/345) commits, so if somebody likes this but not the other PRs, i could pluck it apart.


cantools/database/can/c_source.py
```diff
@@ -397,6 +437,16 @@
 bool {database_name}_{message_name}_{signal_name}_is_in_range({type_name} value);
 '''
 
+SIGNAL_DECLARATION_CLAMP_FMT = '''\
+/**
+ * clamp signal to allowed range.
+ * @param[in] val: requested value
+ * @returns   clamped value
+ */
+double {database_name}_{message_name}_{signal_name}_clamp(double val);
```
drewr95, line 446: Are there other clamping functions that support other types?

v0lker: apparently you can also choose to use floats instead with a command line flag since https://github.com/cantools/cantools/pull/394. it should still work (promoting the return value to double) or, if desired, one could also use a float to store the return value. that change should hopefully not be too tedious, but since this PR has apparently little chance of being accepted, it's low priority from my perspective right now..

drewr95: That's what I figured.  I like the idea of adding clamping and wrapping just wanted to make sure I didn't miss anything.  You were nice enough to provide 'CTOOLS_MIN/MAX' macros.  I was wondering if you could repurpose them into a CTOOLS_CLAMP.

Something like this that I quickly threw together: https://replit.com/@dr5801/CTOOLSCLAMP#main.c

v0lker: you mean that using a macro instead of the `_clamp()` functions? i would hope that the compiler might eliminate the overhead, but have never checked the generated code myself. istr that even integers are unpacked from the package as doubles, though (?) (in which case there will always be a cast back and forth?)

i could probably move things around a bit, but i don't have a project where i'm using cantools, atm - in fact i don't even have a test setup any longer. if my changes are useful to you, maybe fork the branch - that will probably give you results quicker than if i had to start working on this again.

PR #347, opened 2021-09-22
make loading subparsers modular

andlaus:
i.e., all python modules underneath the cantools module's 'subparsers' directory are loaded as subparsers automatically. This is a prerequisite to start modularizing cantools.


juleq (approved):
Good to see it being taken on :). Thought about it today because of the pending code generator PR: Since I am not a user of that subparser at present, I would rather not review those changes. Someone that is more invested in that functionality might be a better fit. Enhancing modularization will certainly help going forward. Maybe even dedicated subparser ownership in the future?.

I have read the code, but did not run it. But since CI did so, I'd be ready to merge.

andlaus:
> Thought about it today because of the pending code generator PR: Since I am not a user of that subparser at present, I would rather not review those changes. 

I share that sentiment. I would also like to see alternative code generators, e.g., ones based on template engines like mako (cf #151) or jinja (my personal favorite) as well as generators for more exotic languages like c++, go and rust. For all of those I am too cowardly to merge this functionality into the current monolitic repo but I think there would not be much in the way to merging such stuff into something like a `cantools-codegen` repo...

> Maybe even dedicated subparser ownership in the future?

yep, that's also my plan :)

> I'd be ready to merge.

please do.

PR #349, opened 2021-10-08
ensure that the result of decode() can be fed back into encode()

andlaus:
the named signal value changes broke this...


andlaus:
> Good catch. Think this justifies a release?

A point release probably yes. That said, it seems like nobody noticed during the last month or so, so it's unlikely to be an urgent issue...

PR #350, opened 2021-10-08
encode(): bail out when specifying unknown signals in strict mode

andlaus:
this ensures that encoding fails if the passed dictionary contains entries that do not correspond to a signal specified in the database. The previous behavior was to silently ignore the entry. This is very useful for preventing typos or copy-and-pastos...

### DIFF
--- a/cantools/database/can/message.py
+++ b/cantools/database/can/message.py
@@ -383,6 +383,14 @@ class Message(object):
         if scaling:
             self._check_signals_ranges_scaling(signals, data)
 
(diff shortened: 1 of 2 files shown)


andlaus:
ok: I rebased this on the latest master version (#349 caused some minor merge conflict), but now suddenly coveralls thinks that the test coverage decreased by 3.6% instead of a slight increase before the rebase. Since coveralls only shows green in the details, I suppose that its summary is spurious once again...

PR #361, opened 2021-11-18
ARXML: improve reference handling

andlaus:
This improves handling of references in ARXML insofar that arbitrary objects may now be referenced (instead of just elements of packages) and that the somewhat reasonable use cases for relative references are now supported.

### DIFF
--- a/cantools/database/can/formats/arxml.py
+++ b/cantools/database/can/formats/arxml.py
@@ -84,7 +84,7 @@ class SystemLoader(object):
             raise ValueError('This class only supports AUTOSAR '
                              'versions 3 and 4')
 
-        self._arxml_reference_cache = {}
+        self._create_arxml_references_dict()
 
     def autosar_version_newer(self, major, minor=None, patch=None):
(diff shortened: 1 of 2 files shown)


juleq:
I cannot do an in-depth review on this. But since new features are welcome, just ask for a merge when done. Done?

andlaus:
> Done?

regarding the ARXML functionality, yes: as far as I can see it works and it should not regress anything. You *might* have a few more general some objections, though. (e.g. I tend to mess up the coding style quite often...)

juleq:
I see :). Most notably, there are a few places where you have violated Erics rule concerning the nowline before and after compund statements (if, for). I like that rule ^^.

andlaus:
> I see :). Most notably, there are a few places where you have violated Erics rule concerning the newline before and after compund statements (if, for). I like that rule ^^.

okay, I hope to have fixed this by now. (note: given that this was rather trivial, it was quite a bit more work than I anticipated...)

PR #362, opened 2021-11-22
ARXML: always pretend to be version 4.0 for AR4

andlaus:
For some reason, all AR 4 revisions always use "http://autosar.org/schema/r4.0" as their XML namespace. To find out the exact AUTOSAR revision (i.e., 4.0, 4.1, 4.2, ...), the "xsi:schemaLocation" attribute of the root tag needs to be examined. Since this is pretty fragile (the used naming scheme has changed during the AR4 journey and with the latest naming scheme there seems to be no programmatic way to associate the value of schemaLocation with the AR revision), we pretend to always use AR 4.0. If incompatible changes between AR4 revisions are encountered in the future, it can be supported by implementing all posibilities and backtracking if necessary.

### DIFF
--- a/cantools/database/can/formats/arxml.py
+++ b/cantools/database/can/formats/arxml.py
@@ -522,10 +522,11 @@ class SystemLoader(object):
                 return None
 
             literal_spec = \
-                self._follow_arxml_reference(base_elem=signal_elem,
-                                             arxml_path=ref_elem.text,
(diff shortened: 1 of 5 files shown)


andlaus:
> Maybe their strategy is to only bump the namespace on major?

possibly. I somehow doubt that there will ever be an AR5, though...

> Should I "auto-merge" changes to arxml? I think you might be the most advanced user there :).

If you're happy with the code quality and coverage: sure...

PR #367, opened 2021-11-26
ARXML: extract the signal values for "invalid signal"

andlaus:
for many -- but not all -- signals, this is the same as initial value.


andlaus:
@juleq : merge at will...

PR #368, opened 2021-11-30
ARXML: extract bus information

andlaus:
This internalizes the information about the CAN buses which can be provided by ARXML files. most of the infrastructure was already present, the exception is the FD baudrate which was not around, supposedly because the base DBC file format (probably) does not support CAN-FD.

Besides this, I added the file format version of the input file to the autosar specifics of the database and cleaned up the message loading code slightly (to avoid using nested functions).


andlaus:
@juleq: as usual; merge at will

PR #372, opened 2021-12-07
Prune signal choices

andlaus:
This gets rid if redundant prefixes of the named values of signals thus reducing visual noise when decoding such signals considerably while still guaranteeing the uniqueness and validity of the new names.

On one hand, this requires minor modifications for some of the user scripts (sorry in advance to all affected users), but on the other hand, I think that it improves usability considerably enough to enable it by default.

### DIFF
--- a/cantools/database/__init__.py
+++ b/cantools/database/__init__.py
@@ -5,6 +5,7 @@ from .errors import Error
 from ..compat import fopen
 from . import can
 from . import diagnostics
+from . import utils
 import textparser
 import diskcache
 
@@ -96,6 +97,7 @@ def load_file(filename,
               database_format=None,
               encoding=None,
               frame_id_mask=None,
+              prune_choices=True,
               strict=True,
               cache_dir=None):
     """Open, read and parse given database file and return a
@@ -177,6 +179,7 @@ def load_file(filename,
             return load(fin,
                         database_format,
                         frame_id_mask,
+                        prune_choices,
                         strict)
     else:
         return _load_file_cache(filename,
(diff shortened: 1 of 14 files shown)


andlaus:
@juleq: I'd like to have your opinion here; shall the default for this be on or off when loading a database? I think that the results when turning it on much easier to read (in particular for large databases), but this constitutes an API change that might be undesireable...

juleq:
I am for reducing clutter when context is already provided.

And the option to output these "fully qualified choices" remains available. If users have parsers consuming the output, they can flick the switch to restore previous behaviour. Hopefully they just do split->last, that won't even break.

If we merge it i'd say a major bump is warranted.

andlaus:
> If we merge it i'd say a major bump is warranted.

I agree. I guess that it should also be mentioned in the release notes. (I noticed that you went through the hassle of creating some for the last release. nice!)

juleq:
Hold your horses. I just used Githubs shiny auto release notes feature for that. They are derived from your commit messages ;).

andlaus:
> They are derived from your commit messages ;).

another motivation for doing comprehensive commit messages ;)

juleq:
Sorry. Busy. I think I'll get to it tomorrow.

AxelVoitier:
Hello,

I came here after updating cantools broke several applications of ours...
Ok, you have a switch for it. But the sensible default was to keep behaviour unchanged. A major version bump is not an excuse for a reckless chanbe like that ...

juleq:
@AxelVoitier The default is set back to the old behaviour now. That said, I would recommend:

- Take part in discussions around the development of cantools if you depend on it
- Starting off a discussion without terms like "excuse" and "reckless" is probably advisable to make fast progress (and as a bonus it will lighten the mood)
- Have a test suite that tests for breaking changes
- Fix your dependencies to a version that you deem stable for your purposes

Best regards

Issue #375, opened 2021-12-20
arxml.py file issue. I can't load arxml database.


linpanusst:
I have some problem with load the arxml database file in the python object. The error message is :
`  File "C:\ProgramData\Anaconda3\lib\site-packages\cantools\database\can\formats\arxml.py", line 87, in parse_int_string
    return int(in_string, 0) # autodetect the base
ValueError: invalid literal for int() with base 0: '0.0'`

I have checked the input. The input is in_string variable is "0.0". It seams that int(in_string, 0) can't decode "0.0". However "0" works fine.
I checked that "21.0" also failed. "21" works fine.

andlaus:
it seems like the code expects an integer but a floating point value is specified by your file. I don't have any file which does this (e.g. fractional scaling works by providing integers for the numerator and the denominator), but if  you provide a test case which validates against the ARXML XSD, I'll fix this...

linpanusst:
Ok, I will prepare some simple test case via your e-mail which is open on your github home page before 2021.12.23. Thank you for  your time.

linpanusst:
Ok, I will prepare some simple test case via your e-mail which is open on your github home page before 2021.12.23. Thank you for  your time.

linpanusst:
> it seems like the code expects an integer but a floating point value is specified by your file. I don't have any file which does this (e.g. fractional scaling works by providing integers for the numerator and the denominator), but if you provide a test case which validates against the ARXML XSD, I'll fix this...

Hi, I have e-mail you some test data. Sorry for the delay due to some health problem. Thank you for your time.

PR #379, opened 2022-01-11
ARXML: be more lenient about accepting floating point values

andlaus:
Besides accepting fractional values for floating point signals, we now also accept floating point values whenever we expect integers provided that the specified value exhibits a decimal value of `.0`. This is a superset of the AUTOSAR mandated behavior, i.e. all valid ARXML files can be loaded correctly but not all files that can be loaded are valid. (We already do this in some places so we do not really get less standard-adherent...)

This hopefully fixes [#375](https://github.com/cantools/cantools/issues/375). thanks to [at]linpanusst for bringing up this issue and for providing a publishable ARXML file to reproduce it!

### DIFF
--- a/cantools/database/can/formats/arxml.py
+++ b/cantools/database/can/formats/arxml.py
@@ -71,7 +71,20 @@ class AutosarMessageSpecifics(object):
 
 
 
-def parse_int_string(in_string):
+def parse_number_string(in_string, allow_float=False):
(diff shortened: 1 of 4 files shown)


simontegelid:
Looks good to me!

andlaus:
> Looks good to me!

Feel free to press The Green Button :)

PR #381, opened 2022-01-13
Allow to specify the pattern for unused bits

andlaus:
The purpose of this is to avoid undefined behavior when encoding frames that contain bits which are not used by any signal. Currently, this pattern implicitly is 0xff if encoding with the `padding` argument set to `True`, and 0x00 otherwise. ARXML allows to specify something arbitrary, so let's also support this in cantools. IMO, this is quite a fringe feature and I'm not even completely sure if it is implemented fully correctly here: I assumed that a single byte gets specified as an integer (thus, endianness does not matter) and that the bit pattern is not aligned with the beginning of unspecified sections, i.e., the encoded message starts out as a sequence of the unspecified bit pattern bytes and then the parts which define signals get overwritten by actual data.


juleq:
Do you need to rebase to get the linter satisfied?

andlaus:
> Are you sure that Vector uses 0xff as a default? Will the behavior change when you do not enable padding?

if you set the padding argument of `encode()` to true, cantools will currently set all unused bits  to 1 if it is false, they are set to 0 (i.e. not touched by the codecs). If nothing else gets specified, this should hopefully stay the identical after the patch. I have no idea if the current behavior is what Vector does, though...

andlaus:
> Do you need to rebase to get the linter satisfied?

probably. I'll have a look...

cantools/database/can/message.py
```diff
@@ -65,6 +66,7 @@ def __init__(self,
         self._length = length
         self._signals = signals
         self._signals.sort(key=start_bit)
+        self.unused_bit_pattern = unused_bit_pattern
```
zariiii9003, line 69:
```suggestion
        self._unused_bit_pattern = unused_bit_pattern
```

andlaus: oops. fixed, thanks :)

andlaus:
yay! the linter is happy now :)

PR #382, opened 2022-01-14
3 Bugfixes related to dbc files

erzoe:
This pull request contains three fixes for what looked like bugs to me at first.

1) The `value` attribute of a `NamedSignalValue` instance is now an `int` instead of a `str`
   see commit 14362f9ca70bb27a12aed753806041aa799a5f3c
2) A choice named `"CmdRespOK"` was loaded as `"K"` and a choice named `"CmdRespErr"` was loaded as `"rr"`. While looking into this I have found that abbreviating the names is the intended default behavior and can be disabled with `prune_choices=False` and I have added that to the doc string of `load_file`. But the algorithm assumes the names contain underscores and if they don't it strips away too much of their name. Therefore I have changed the behavior of `prune_choices=True` to not abbreviate choice names which don't include an underscore.
   see commit 2fd5199baa744f60127a19de27b3ad7dd73a12af
3) Loading a dbc file and dumping it changed the order of multiplexed signals every time the dbc file was loaded and dumped. This is not a problem as long as only computers look at the file. But if a human wants to review certain changes this makes things difficult. While looking into this I have realized that not preserving the order of the signals is the intended behavior, signals are sorted by the start bit when loaded and reversed in order when written to the file. When using multipexed messages, however, it might make sense to not sort the signals by their start bit. I have therefore added a keyword argument `reverse_signals` which if explicitly set to `False` prevents the sorting and reversing of signals.
   see commits 749ef66b81a4b8b020a275f994fddd5cb93e505e and b643cd0eb89de78122aaca925a2db2f126658ddd


cantools/database/__init__.py
```diff
@@ -78,6 +78,7 @@ def _load_file_cache(filename: StringPathLike,
                      frame_id_mask: Optional[int],
                      strict: bool,
                      cache_dir: str,
+                     sort_signals : bool,
```
andlaus, line 81: possibly, it's better to specify a compare function (or `None` if the signals ought not to be sorted). Anyway I'm also fine with the proposed solution...

cantools/database/can/formats/dbc.py
```diff
@@ -556,7 +560,7 @@ def _dump_senders(database):
     return bo_tx_bu
 
 
-def _dump_comments(database):
+def _dump_comments(database, reverse_signals):
```
andlaus, line 563: as above: passing a `sort_fn` (or similar) function would be more general...

cantools/database/can/formats/dbc.py
```diff
@@ -1006,7 +1022,7 @@ def _load_value_tables(tokens):
 
     for value_table in tokens.get('VAL_TABLE_', []):
         name = value_table[1]
-        choices = {int(number): NamedSignalValue(number, text) for number, text in value_table[2]}
+        choices = {int(number): NamedSignalValue(int(number), text) for number, text in value_table[2]}
```
andlaus, line 1025: hm, I'm wondering why this worked at all before. Anyway, explitly casting this to int should not hurt...

cantools/database/utils.py
```diff
@@ -297,6 +297,11 @@ def prune_signal_choices(signal):
 
     if i >= 0:
         full_prefix = full_prefix[0:i]
+    else:
+        # full_prefix does not contain an underscore
+        # but the following algorithm assumes it does
+        # and would strip too much
+        return
```
andlaus, line 304: ops. that's a bug indeed. Thanks for the fix! Question: should we just remove the common prefix? this would imply that your examples get reduced to "Err" and "OK", but depending on the conventions used, it might lead to unexpected results as well...

tests/test_database.py
```diff
@@ -384,6 +384,26 @@ def test_dbc_load_empty_choice(self):
         self.assertEqual(message.signals[2].name, 'no_choice')
         self.assertEqual(message.signals[2].choices, None)
```
andlaus, line 386: is there no test for changing the order of the signals? (would it make sense to sort the messages as well, BTW?)

andlaus (commented):
Thanks for fixing my embarassing mistakes :). The PR looks good to me; besides the usual nitpicks, I'd just like to see a test about the signal ordering...

tests/test_database.py
juleq: I think dbc retains the order of messages chronologically.

cantools/database/__init__.py
erzoe: good idea, I have implemented that in commit a802630078421a75fd12d9ce9a931bc2c81f2379. And while implementing that I have found another bug, see commit bdc543c79af2e263f3f3bb906862b584bd5876ce. While I was at it I have also added a few more type hints in commit fedc1d7dc6e1a500c938d3770b5df43926ade892.

cantools/database/can/formats/dbc.py
erzoe: I have implemented that in commit 3f0916797c6e7fdb0199a18b24e2d191d5513775. Unfortunately I'm not very experienced with typing yet and mypy isn't happy with what I've done. Do you know how to fix those errors?

tests/test_database.py
erzoe: I'm not aware of any currently existing tests for the sorting of signals which happens by default. I have added a test for *not* sorting the signals in commit fc68778c9e3ec08ac6520b648f12e05f63e7babf, though. But I intend to add a test for passing a function to the new `sort_signals` parameter and also for leaving it out.

erzoe: I don't think it makes much sense to sort the messages. If a user wants them sorted they can easily add a `db.messages.sort(key=lambda msg: msg.frame_id)` to their script.

andlaus: good point

cantools/database/can/formats/dbc.py
erzoe: I have found the solution, mypy is happy now

cantools/database/can/database.py
```diff
@@ -17,6 +17,7 @@
 from .internal_database import InternalDatabase
 from .message import Message
 from .node import Node
+from ..utils import type_sort_signals, sort_signals_by_start_bit, DEFAULT
```
andlaus, line 20: I guess it is advisable to rename `DEFAULT` to `dont_sort_signals`, or `sort_signals_default`

cantools/database/utils.py
```diff
@@ -337,3 +343,12 @@ def prune_database_choices(database):
         for signal in message.signals:
             prune_signal_choices(signal)
 
+
+DEFAULT: Final = 'default'
```
andlaus, line 347: this might better be a function (which does nothing) just like the other sort functions? It would simplify the typing, but on the other hand, this would make implementing different behaviors for multiple file formats complicated. You decide.

andlaus (commented):
except for the unexpressive `utils.DEFAULT` name, I think this is ready to go in.

cantools/database/can/database.py
erzoe: I used a generic name because I thought this might be useful for other arguments, too. But I don't mind renaming it to `SORT_SIGNALS_DEFAULT`. (Done.) I am surprised that you are suggesting snake case names, though, as constants are [usually](https://www.python.org/dev/peps/pep-0008/#constants) written in upper case. Or does cantools follow a different coding style?

`dont_sort_signals` would not be a fitting name because the default action depends on the file format (when dumping files). Not sorting is the default for kcd files but for dbc files the default is to sort the signals by their start bit in *de*scending order. (I am not particularly happy with this default but I didn't want to change the default behaviour.) I have documented this in the [doc string of `dump_file`](https://github.com/erzoe/cantools/blob/bugfix-dbc/cantools/database/__init__.py#L217).

For loading files the default is to always sort the signals by their start bit in *a*scending order. I have documented this in the [doc string of `load_string`](https://github.com/erzoe/cantools/blob/bugfix-dbc/cantools/database/__init__.py#L314).

If a user does not want sorting they should pass `sort_signals=None`.

cantools/database/utils.py
erzoe: Yes, passing in a function would have simplified typing but I think passing a function which is not executed but compared whether it is a certain object would be somewhat confusing. And now that we have `type_sort_signals` it is easy to maintain.

juleq:
Thanks for the fixes. All checks passing. Tell me when you are ready to merge.

erzoe:
> Thanks for the fixes. All checks passing. Tell me when you are ready to merge.

I'm ready

cantools/database/can/database.py
andlaus: Thanks for the rename, all is golden now AFAICS...

> I am surprised that you are suggesting snake case names, though, as constants are usually written in upper case. Or does cantools follow a different coding style?

no, you're right. I suppose that this is just a bad old habit of mine ;)

PR #386, opened 2022-01-26
ARXML: implement extraction of sender and receiver nodes of messages and signals

andlaus:
This functionality is currently still somewhat incomplete: AUTOSAR allows to specify senders and receivers of individual PDUs which do not necessarily directly correspond to CAN frames. Currently, we simply ignore this fact and assume any frame which features a transmit-PDU is being send by the given ECU and in the case of a receive PDU, we assume the ECU to be a receiver of all signals defined by the frame which features the receive-PDU.


andlaus:
@juleq: as usual: please merge if you're happy with this...

PR #391, opened 2022-01-28
Handle Shared Signals With Multiplexed Message

### LINKED ISSUE #390: C Source Generation Duplicates Functions With Multiplexed Signals
`cantools generate_c_source` is generating an invalid C file when I use a multiplexed SYM file where each multiplexed signal shares a signal. For example, given the below .sym file:

```
FormatVersion=6.0 // Do not edit this line!
Title="Untitled"

{SIGNALS}
Sig=Shared_Signal unsigned 8 /max:1

{SENDRECEIVE}

[MuxedSymbol]
ID=234h
Len=8
Mux=FirstMux 0,1 0
Sig=Shared_Signal 8

[MuxedSymbol]
Len=8
Mux=SecondMux 0,1 1
Sig=Shared_Signal 8
```

This will generate a duplicate `_encode`, `_decode`, and `_is_in_range` function for the `shared_signal`, since `muxed_symbol` is multiplexed.

samc1213:
The root cause of this issue is that a new `Signal` object is instantiated for each multiplex instance, with different `multiplexer_ids`. Since the object gets duplicated, the C source generation is also duplicated. I chose to fix the C source generation, since the object structure is much more complex and disruptive to change.

Closes #390


juleq:
Great, thanks. Since you have already provided a test I could merge if you confirm that you are done.

samc1213:
> Great, thanks. Since you have already provided a test I could merge if you confirm that you are done.

Can you please wait to merge? I would like to test a little further, I just wanted to see how the automated tests run went

juleq:
Sure. Ready when you are.

samc1213:
@juleq This change should be good to merge - However, I feel that the representation of multiplex signals in the cantools database can cause some issues. For example, the duplication of values shared across multiplexed signals persists in `cantools decode`:

````
  can0  00FF3202   [8]  02 8E 01 00 00 00 00 00 ::
Query(
    Real: 2,
    ParameterNumber: 'ACTUAL_SPD',
    ParameterNumber: 'ACTUAL_SPD',
    ParameterNumber: 'ACTUAL_SPD',
    ParameterNumber: 'ACTUAL_SPD',
    ParameterNumber: 'ACTUAL_SPD',
    IntValue: 0
)
```

(note that ParameterNumber is repeated - I have a Mux Signal called "real" with 5 possible values, hence it shows up 5 times)

- Should we come up with a way to remove the duplication of Signal in the db for each multiplex id in the future?

Thanks for merging this promptly :)

juleq:
Hi. Could you elaborate a little bit on the purpose?

I am not familiar with SYM. But I have just checked with CANdb++: A dbc file will not let you add a signal multiple times to a message. If it is a multiplexed message, your only option to have it appear for multiple multiplexors is to make it a non multiplexed signal. That way, it will be present for each multiplexor.

I assume that this is also an option for SYM.

The issue otherwise would be: If a signal would be associated with multiple multiplexors, they also could have e.g. a different positioning within the message. The PCAN Symbol Editor lets you do that. You need multiple instances then.

However, I am not sure that such a feature is currently supported/considered throughout cantools. You might have surprises.

I would also argue that it might prove useful if your CAN definition is created in a way that does not rely on a feature that is unique to one vendor. Imagine a customer asks for a dbc because he uses the quite common Vector tooling... This is not representable in dbc.

samc1213:
I see, that does make sense to me. I didn't realize the positioning within the message does not need to be consistent across each multiplexor if you use the same signal across them. Then I think this change makes sense as-is, since it just dedups the `encode` and `decode` functions, and the autogen'd struct member which should not be duplicated regardless of the signal location within the multiplexor. Hopefully my unit test conveys the issue with the current `c_source` implementation when using a signal across multiplexors.

I now understand the reasoning for having a new `Signal` for each instance of a signal inside of a multiplexed message, so I really appreciate that explanation. What you are describing is allowed in PCAN, I just never have tried to use the same Signal in a different place across multiplexors.

Please let me know if you have other questions, and thanks for explaining that.

juleq:
Have you considered changing your CAN definition?

andlaus:
> What you are describing is allowed in PCAN, I just never have tried to use the same Signal in a different place across multiplexors.

This is slightly off topic: As far as I understand classical signal-based communication (AUTOSAR, DBC, and probably most other approaches), signal names must be globally unique for the whole system, i.e.if there is more than one signal with a given name, this is an error regardless of whether this signal duplicated within multiplexers, normal frames, container frames or even on a different bus. (I might be wrong here, though.)

juleq:
@andlaus I think the uniqueness is only required for the signals in one message in the case of dbc. The consistency check of CANdb++ also will not complain if you have multiple signals with the same name.

However, in the case at hand uniqueness of the signal name is no longer given and as far as I can see that is a feature that seems to be exclusive to SYM. The underlying format specification solves it like this: There are multiple definitions of each message, one for each multiplexor. Further, signals are globally defined and then referenced by these message definitions. That way you can represent a shared signal that exists for multiple (but not necessarily all) multiplexors and even position them differently (because the "sub-message" just references the signal and adds the positioning).

BUT: Since the message definitions in SYM reference a signal by name, the Peak tool will NOT allow you to create signals with the same name (globally unique collection). dbc can do that. There is a "catch all" message that contains all signals that have not been associated with a message. This collection has no uniqueness constraint. Signals become unique (within a message) as soon as they are put into a message.

andlaus:
ok. Do I understand this correctly: the issue at hand is probably a bug/ideosyncracy of the PCAN tool?

> @andlaus I think the uniqueness is only required for the signals in one message in the case of dbc. The consistency check of CANdb++ also will not complain if you have multiple signals with the same name.

I probably did not express what I meant very well: I meant that, in the context of signal based communication, if a signal with the same name gets specified by multiple different messages (multiplexed or otherwise), the global state of the system gets shoddy (because some ECUs might not extract a subset of these messages) and you'll be in trouble very quickly. Some file formats may support this to one extend or another, though...

juleq:
It appears to be a "feature" of SYM, but I would not recommend using it.

I agree that this feature is problematic w.r.t. consistency. The ECU could decide it is not interested in a certain mux value and might miss updates to the shared signal...

juleq:
And as always rule #1 pertains: Don't get clever. 😁 Other tools not supporting that is probably a hint that it is too clever.

samc1213:
I appreciate all the discussion and feedback here. To explain what I am doing, I'm trying to multiplex one CAN ID in order to communicate a wide variety of parameters to a device. The parameters are a variety of different data types. So, I'm creating a message that looks like this:

```
<data_type (MULTIPLEXOR)> <parameter_number > <parameter_value>
```

`parameter_number` is the Signal that exists in every multiplexor instance. 

For example, for a bit parameter, `data_type=BIT`, and `parameter_value` is a 1-bit signal. However, for an unsigned integer parameter, `data_type=UINT`, and `parameter_value` is a 32-bit signal. For each of these messages, though, `parameter_number` is the same, shared Signal. 

I think this level of abstraction in the SYM file is fairly uncommon - hence your concern about the ECU ignoring the Signal if it's not listening to the right mux value and missing the shared signal doesn't really apply to this setup. A simpler approach would be to explicitly list out each parameter_number as a separate multiplexor, and not multiplex on data type. Please let me know your thoughts on what might be most appropriate here.

I'm happy to close this PR if we feel it adds unnecessary complexity and I'm not really using the SYM file in a standard or straightforward way. This discussion of what is typical in other file formats also gives me some great context.

andlaus:
hm, I wonder if something like this is possible using SYM:

```
parameter_number,data_type_multiplexer{parameter_value}
```

i.e. make the parameter number a static signal outside of the multiplexer. I'm most familiar with ARXML and it allows stuff like this (and cantools supports it)

andlaus:
that should probably have been
```
parameter_number,data_type_multiplexer{parameter_value_$DATA_TYPE}
```

but I guess you get the idea...

andlaus:
come think about it, the `param_number` signal is redundant with `value_$PARAM_NUM`, i.e. your frame could look like this

```
value_multiplexer{value_$PARAM_NUMBER}
```

andlaus:
come think about it, the `param_number` signal is redundant with `value_$PARAM_NUM`, i.e. your frame could look like this

```
value_multiplexer{value_$PARAM_NUMBER}
```

andlaus:
come think about it, the `param_number` signal is redundant with `value_$PARAM_NUM`, i.e. your frame could look like this

```
value_multiplexer{value_$PARAM_NUMBER}
```

juleq:
Yes. You could either use the multiplexor as the parameter number or just have the parameter number outside of the multiplex -> it will be included in all variants of the message.

The advantage of using the multiplexor: You do not have kind of redundant numbering. And you can have each parameter as its own signal with its own type.

A disadvantage might be: I have seen tools that have constraints on the multiplexor. E.g. it has to be byte0 unsigned.

Alternatively you can drop the multiplex and solve this on a higher level. Use a parameter number signal, a 32bit unsigned signal and a type signal. You could drop the latter if the recipient knows the whole parameter list upfront and knows what type is to expect for which parameter. Depending on the scale of your system, it might even be advantageous not to have a signal for each parameter...

geo4711:
> But I have just checked with CANdb++: A dbc file will not let you add a signal multiple times to a message. If it is a multiplexed message, your only option to have it appear for multiple multiplexors is to make it a non multiplexed signal.

Hi!
Just a word to using same sig with different mux values: CANDB++ *does* support this. It is called "extended multiplexing". But there are restrictions, e. g. it is allowed for certain "types" of DBC only (e. g. for J1939 ones).
Cantools does support this, too. I think it was version 35.1 where we introduced/improved support for that.
I'm not familier with the other features mentioned in this discussion (C code generation, SYM import, ...), but as I've reade the "extended multiplexing" aspect, I tried to help :-)

juleq:
Yes, thats true if you choose to use J1939.

If you want to read up on extended multiplexing in dbcs:

https://cdn.vector.com/cms/content/know-how/_application-notes/AN-ION-1-0521_Extended_Signal_Multiplexing.pdf

I have never checked the cantools implementation in that regard.

samc1213:
> make the parameter number a static signal outside of the multiplexer

This is exactly what I would like to do, but looks like SYM does not support this. Once a Message has been multiplexed, all Signals must belong to a specific multiplexor value.

I will close this PR, hopefully this discussion will benefit someone trying to understand the limitations of SYM multiplexing. I will likely just multiplex on `parameter_number`. Thanks for all your feedback

PR #394, opened 2022-02-07
Add --use-float flag to generate_c_source

### LINKED ISSUE #393: Enable Floats in C Source Generation
Right now, `generate_c_source` only generates encode and decode functions with `double`s. Some floating point units may only have hardware support for single precision, so it might be nice to enable a flag like `generate_c_source --use-float`, which would use `float` instead of double.

@juleq Does this proposal make sense or do you forsee any issue? I'm happy to implement if you think this would be a worthwhile change.

samc1213:
* If you add --use-float, any floating point
  decoding/encoding will be done using the
  float type instead of the double type
* This is to be used on ECUs that only
  have a single-precision floating point
  unit
* NOTE: If a database file is using a type
  called Double (as in, the Signal in the DBC
  file specifies Double), then this autogenerated
  code will still use the double type, since there
  is no other valid way to handle this

closes #393


juleq:
Sorry, have been off to a business trip this week. @samc1213 can you already confirm that this would be ready to be merged?

To other users of the code generator: I you had the chance to try this, I would appreciate your feedback.

samc1213:
Yes this should be good to merge

jgui:
After review this pull request has exactly the same purpose as mine.
I compare the code => better than mine (I miss adding the 'f' suffix)
I try it => similar result.
The only thing that should be added is an example in README.rst so we know this option exists.

samc1213:
@jgui Thank you for taking a look. I updated README.rst in a similar fashion to your PR.

juleq:
Ok gentlemen, I'll merge now. Thank you both for your contribution.

You can expect a release later today.

PR #398, opened 2022-02-11
Add option to get single precision floating point (float) in the generated code, instead of double.

jgui:
For targets having single precision as native type for floating point numbers, ability to generate C code with float type instead double is welcome.
This pull request implements this possibility by adding the CLI option "--single-precision-floating-point":

- If specified, "float" is used in the generated C code.
- If not specified, "double" is used in the generated C code (the default behavior is then not modified).

A test has been added but not run due to difficulties with my Linux Ubuntu 16.04 (I work mainly under Windows). I'm not able to install Python 3.6  and lack time to setup another linux.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -298,6 +298,14 @@ file names are also changed by this option.
 See `my_database_name.h`_ and `my_database_name.c`_ for the contents
 of the generated files.
 
+In this example we use ``--single-precision-floating-point`` so floating point numbers in the generated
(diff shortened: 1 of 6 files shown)


juleq:
@jgui I think #394 tries to address  this already. Could you give that a try and add your confirmation/feedback to #394?

Thank you.

juleq:
Implemented by #394.

PR #399, opened 2022-02-16
Implement container messages

andlaus:
This PR is rather large (sorry). It implements AUTOSAR-style container messages which use the "short header" format. For this to make any sense in practice, the payload of such frames must be longer than eight bytes, i.e. CAN-FD frames is a precondition for container messages in practice (though in theory container frames also work using traditional CAN frames.) As a consequence, ARXML is currently the only supported file format for reading container messages. That said, for other file formats which support this feature, it will not be too hard to extend the code accordingly.

Be aware that -- since decoding container messages results in a list of (Message, SignalDict) tuples instead of a single SignalDict --, `decode()` must be explicitly instructed to handle them by setting the `decode_containers` parameter to `True`. The reason for this is that with this solution, user code which works before this PR will do so afterwards (if it handles `DecodeError` for `decode()`, which is basically mandatory anyway), but if container messages were decoded by default, such existing code would almost certainly raise unexpected exceptions if any container message were encountered.

Besides implementing the infrastructure, this commit adapts the `list`, `decode` and `monitor` subparsers to  handle containers. I think that at least adapting `dump` would be beneficial, but since the layout of containers is dynamic, I could not come up with a meaningful way of formatting the output.


andlaus:
@juleq: as usual: merge if you're happy. I went over this a few times, so I hope the biggest issues with this should be fixed, but it would not hurt if you skimmed over it because when reading my own stuff, I often read what I meant, not what I actually wrote.

juleq:
Hi there. Wow, that must have taken a few coffees 😀.

I am a little bit concerned with respect to maintainability. This is fairly specific to ARXML, but it adds to database and message, which I would consider "core API".

Our BDFL once asked me to contain the stuff that is specific to dbcs in a member so that I do not litter into the generic parts. That is where dbc_specifics came from. It was a far less complex matter, but I was still reminded by this proposal.

Do you see a way to contain it to a greater extend?

That being said, I am not here to stand in the way. If you feel that this is the way to go, I would not oppose.

Concerning a review: I have only scanned very shallowly. I cannot spent more time at present.

Best

andlaus:
>  Wow, that must have taken a few coffees

you bet ;)

> Our BDFL once asked me to contain the stuff that is specific to dbcs in a member so that I do not litter into the generic parts. [...] Do you see a way to contain it to a greater extend?

I am aware of this issue, the problem is that the only part that can be relatively easily encapsulated in the `autosar` objects that I could come up with is specifying the exact layout of the containers, i.e. the header format, header byte order etc. IMO the issue with this is that the encoding and decoding facilities must be aware of it anyway, so encapsulating it would make the encoding and decoding parts of cantools considerably more complicated and since so far the autosar-containers-with-short-headers mechanism is the only supported one, I decided to avoid unnecessary abstractions for the sake of readability, at least for now.

> Concerning a review: I have only scanned very shallowly. I cannot spent more time at present.

That's all I asked for: I don't fully trust myself not to deliver major screwups, so I'm grateful if somebody like you skims over my stuff and does not immediately notices any either ;)

juleq:
Hi Andreas,

the complexity of the type annotation on the decode_message return is the one place where I would suggest reconsideration. Wouldn't it be possible to use a class that contains that complexity and type check against that? So the return annotation would just be the Union of that class and the previous dict?

What do you think?

andlaus:
> the complexity of the type annotation on the decode_message return is the one place where I would suggest reconsideration. 

good point. I was annoyed by this as well, but not enough to spend time about thinking how this could be improved.

> Wouldn't it be possible to use a class that contains that complexity and type check against that? 

I think that this would solve the problem of the way too complex type annotations, but it would probably make *using* these objects in the calling code quite a bit more difficult because they would become custom objects instead of build-in ones. I suppose that type aliases are a better solution here. Unfortunately the PEP for explicit aliases (PEP-613) has only been implemented in python 3.10, so we should probably wait for a few more years until we start using it. Maybe it is possible to use old-style implicit aliases in a not too confusing way, I'll give it a shot...

cantools/database/can/message.py
```diff
@@ -23,6 +32,17 @@
     from .formats.arxml import AutosarMessageSpecifics
     from .formats.dbc import DbcSpecifics
 
+LOGGER = logging.getLogger(__name__)
+
+# type aliases
```
juleq, line 37: Comment suggestion: Type aliases. Introduced to reduce type annotation complexity while allowing for more complex encode/decode stategies like the one used for ARXML containers.

juleq (commented):
Looking good. I have added one comment suggestion to save future me and future you from head scratches :).

cantools/database/can/message.py
andlaus: done

juleq:
Hello Andreas,

something is wrong. All checks have passed but after I merged the two test jobs started failing. I re-ran them, outcome is the same. That´s odd. Could you have a look?

The whole log should be here: [Failing run](https://github.com/cantools/cantools/runs/5258944198).

The first failed test logs this:

`======================================================================
ERROR: test_bad_message_name (tests.test_tester.CanToolsTesterTest)
Try to send an unknown message.
----------------------------------------------------------------------
  File "/home/runner/work/cantools/cantools/tests/test_tester.py", line 271, in test_bad_message_name
    tester, _ = setup_tester('Node1')
  File "/home/runner/work/cantools/cantools/tests/test_tester.py", line 61, in setup_tester
    tester = cantools.tester.Tester(dut_name,
  File "/home/runner/work/cantools/cantools/cantools/tester.py", line 283, in __init__
    self._messages[message.name] = Message(message,
  File "/home/runner/work/cantools/cantools/cantools/tester.py", line 86, in __init__
    self.update(self._prepare_initial_signal_values())
  File "/home/runner/work/cantools/cantools/cantools/tester.py", line 101, in update
    self._update_can_message()
  File "/home/runner/work/cantools/cantools/cantools/tester.py", line 194, in _update_can_message
    self._can_message = can.Message(arbitration_id=arbitration_id,
TypeError: __init__() got an unexpected keyword argument 'extended_id'
`

andlaus:
it seems like python-can renamed the `extended_id` parameter of the constructor for their `Message` class to `is_extended_id`.  #400 should fix this...

andlaus:
alright, I digged a bit deeper: this seems to be caused by with python-can 4.0.0; it has removed the `extended_id` parameter in favour of `is_extended_id` and there seems to be an issue with virtual can busses and notifiers in that version. @juleq: shall we open an issue for python can and wait, or shall we restrict the python-can dependency to < 4.0 ?

juleq:
I would rather stick to < 4.0 for now. I assume there will be a little back and forth that we should let pass us by :).

andlaus:
alright; I've adapted the PR and will open an issue against python-can.

PR #400, opened 2022-02-19
fix the tester test for newer python can versions

andlaus:
also, specify the minimum required python-can version. this was motivated by https://github.com/cantools/cantools/pull/399#issuecomment-1046026905 but is probably unrelated to #399.

### DIFF
--- a/cantools/tester.py
+++ b/cantools/tester.py
@@ -192,7 +192,7 @@ class Message(UserDict, object):
                                     self.scaling,
                                     self.padding)
         self._can_message = can.Message(arbitration_id=arbitration_id,
-                                        extended_id=extended_id,
+                                        is_extended_id=extended_id,
                                         data=data)
 
         if self._periodic_task is not None:
--- a/tests/test_tester.py
+++ b/tests/test_tester.py
@@ -13,7 +13,7 @@ except ImportError:
 import cantools
 
 
-class CanBus(object):
+class CanBus(can.BusABC):
 
     def __init__(self):
         self.channel_info = None


andlaus:
it seems like the CI system for python 3.9 has an issue with python-can, or else I can't make any sense out of the following error:
```
  File "/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/can/notifier.py", line 69, in add_bus
    reader = bus.fileno()
AttributeError: 'CanBus' object has no attribute 'fileno'
```

is it possible to tell the CI-host to reinstall the requirements?

juleq:
Have you seen that python-can released 4.0.0 16 hours ago? I assume some breaking changes. Maybe fix to <4.0.0 for now to let it stabilize?

andlaus:
python-can issue: https://github.com/hardbyte/python-can/issues/1261 . Let's see what happens...

andlaus:
> Have you seen that python-can released 4.0.0 16 hours ago? I assume some breaking changes. Maybe fix to <4.0.0 for now to let it stabilize?

agreed. done.

andlaus:
as @zariiii9003 pointed out in https://github.com/hardbyte/python-can/issues/1261#issuecomment-1046146759, the custom `CanBus` class from `test_tester.py` did not derive from `can.BusABC`, which seens to have become mandatory in python-can 4.0.0. Fixed.

While at it, I also raised the minimum required python-can version to 4.0. Not completely sure if this a good idea...

andlaus:
> Not completely sure if this a good idea...

it isn't: python-can 4.0.0 does not work with python 3.6. reverted...

juleq:
Ok, was worth a shot. We should probably keep Python 3.6 support for now until we decide it is time to migrate to python-can 4.x.y.

We should probably do a major bump when the time has come to migrate. We can instruct Python 3.6 users to fix their tag to before if they can't update Python.

Since we currently break for new users because of the unfixed python-can requirement, I would propose to release.

Is this ready to merge?

andlaus:
> Is this ready to merge?

As far as I'm aware: Yes. It works with python-can 3.3 (and hence python 3.6) as well with python-can 4.0, so AFAICS it is a pretty nice solution...

andlaus:
> I would propose to release.

agreed, though my reason is rather to make the container message support easier to access. (which is pretty important for my employer internally...)

PR #404, opened 2022-02-24
ARXML: Use floats to represent the minimum and maximum of data types featuring linear segments

andlaus:
Conceptually, values that represent quantities from the physical world are always floating point values. In practice, this should not matter too much because such tables often exhibit a scaling factor of 1 and an offset of 0. It does not hurt to also handle the corner cases, though...


andlaus:
@juleq: AFAICS this is merge-ready...

Issue #358, opened 2021-11-11
EncodeError for multiplexed Message and strict=True


zariiii9003:
I have a dictionary that contains all signals of a multiplexed message. It does not contain any other signals.
When calling the `Message.encode` method with `strict=True` i get an EncodeError.

This breaking change was introduced in #350.

After [this line](https://github.com/cantools/cantools/blob/f52c7d657dd9c128a04594f1e18109aa43c53bd0/cantools/database/can/message.py#L445) the variable `all_signals ` only contains the signals that belong to the current value of the multiplexor signal.

Then in [this line](https://github.com/cantools/cantools/blob/f52c7d657dd9c128a04594f1e18109aa43c53bd0/cantools/database/can/message.py#L449) the `data` dictionary (that contains all signals of the message) is compared to `all_signals` (which only contains the signals for the current value of the multiplexor).

I would need to use a separate dictionary for each value of the multiplexor to avoid this error.

juleq:
What file format do you use?
Can you post a sample?
What is your use case for cantools?

zariiii9003:
It is a DBC file but the OEM wouldn't be happy if i show the example.
The use case is a restbus simulation.

Let's say i have a multiplexed message with a multiplexor signal named 'N_MUX' .
So depending on the value of the multiplexor i have 3 messages:

N_MUX=0, N_1, N_2, N_3
N_MUX=1, N_4, N_5, N_6
N_MUX=2, N_7, N_8, N_9

Now when i call `encode` with a `data` dict that contains N_MUX and all of N_1 up to N_9 i get an EncodeError. But these signals N_1 to N_9 are all contained in the same message in the DBC file. So in my opinion this should not fail.
This worked fine before and now it does not.

juleq:
One multiplex value has one collection that has three enries. So three collections in total.

This is the usual way to implement it. 

Imagine if you had 3x 10ms frames for the muxes. The encoder would have to determine the subset of signals for a message with a given mux on each encode if you do not partition in advance. Thats a lot of unnecessary operations.

andlaus:
> I would need to use a separate dictionary for each value of the multiplexor to avoid this error.

Is there anything which prevents you from encoding your message in non-strict mode? (Because strict mode is -- err -- strict...)

juleq:
The decision to make that fail in strict was taken because not checking it might leave simple and common mistakes (typos) fail silently. Thats not what I would expect from a software that knows about the error, even more so in strict.

[...]

PR #405, opened 2022-02-24
Modularize the 'strict' mode for encoding messages

andlaus:
Passing `strict=True` to `message.encode()` is now just a shortcut for calling `message.assert_{signals,container}_encodable()` before `message.encode()`. Thus, if `strict=False` is passed, the calling code can decide which checks (if any) to apply.

Besides this, `Message.gather_signals()` is added. Given a dictionary containing a superset of all required signals, this method produces a dictionary with only the signal values that are required to encode the message. (The `Message.gather_container()` method serves the same purpose for container messages.) This is quite useful if the application's state is kept in a global dictionary, but one wants to use strict mode when encoding messages during development to avoid mistakes.

This should close [#358](https://github.com/cantools/cantools/issues/358).

### DIFF
--- a/cantools/database/can/message.py
+++ b/cantools/database/can/message.py
@@ -5,6 +5,7 @@ import logging
 from copy import deepcopy
 from typing import (
     List,
+    Sequence,
     Tuple,
     Optional,
     Union,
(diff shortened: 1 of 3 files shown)


andlaus:
@juleq: this is ready to go in from my perspective...

PR #409, opened 2022-03-10
Add Send/Receive Information From SYM Into Message.senders

### LINKED ISSUE #408: Where to add SYM Direction Field?
In PCAN, you can define a message as send, receive, or bi-directional:

image

This is very useful to help generate I/O code, since you can know which messages are inputs vs outputs with respect to your ECU.

However, I don't see a great place to put this information in the cantools `Message` object. `send_type` and `senders` both don't seem exactly right, given `send_type` is more of a DBC concept. Can I please have some guidance on where this flag could go? Or if perhaps it's decoded somewhere already that I'm missing?

samc1213:
* Adds senders named `ECU` or `Peripherals` to the `Message.senders` property based on the Direction chosen in a SYM file. In a SYM file, a message can either be Send (sent by the ECU), Receive (send by Peripheral devices), or Send/Receive (both). Now this is accessible in the cantools database

closes #408


andlaus (commented):
Once the test failures are fixed, I'm happy to approve this :)

samc1213:
@andlaus I will fix the easy tests to fix, but the test converting SYM to DBC is failing. Instead of using an empty sender when converting to DBC, it's using "ECU"/"Peripheral". I feel this is an inherent problem in the sense that SYM only has Send/Receive flag, whereas DBC has Node sender name. Any thoughts on how to reconcile this?

andlaus:
> [...] the test converting SYM to DBC is failing

I guess this just means that you have to update the reference DBC which the test expects. (you should verify that the new reference data is correct using e.g. `cantools list -a $NEW_REFERENCE_DBC_FILE`.)

samc1213:
Ok thanks for the guidance, I've done that and the new generated file looks reasonable, only difference is is the Sending ECUs which is expected

andlaus (approved):
:+1: no complaints from my side anymore

samc1213:
Ok great, thanks for looking. I think it should be ok to merge

PR #414, opened 2022-04-06
Parse SPNs in SYM Files

### LINKED ISSUE #413: Add SPN to SYM Parser
Right now, a `Signal` will never have an `spn` attribute if it's loaded from a SYM file. Can we add this functionality please?

samc1213:
Closes #413

Add logic to parse the SPN from sym files and add them to signals


andlaus:
looks good to me, but I have zero knowledge of SYM. any thoughts @juleq?

juleq:
Straightforward addition that also gets tested, I would not require additional pinky swears for it.

@samc1213 Thanks. Assuming ready to merge?

samc1213:
Yes good to merge, thank you!

PR #418, opened 2022-04-20
Implement support for AUTOSAR secure on-board communication

andlaus:
The purpose of secure on-board communication (SecOC) is to ensure that only signals are accepted if they have been send by ECUs that are supposed to do so and that these signals are "fresh", i.e., that they are not a replay of a previous recording of the bus. Note that SecOC does *not* encrypt anything and that it also does nothing to prevent DOS attacks like a rouge ECU preventing somebody else from sending.

Also be aware that the AUTOSAR SecOC specification only specifies the scaffolding around verifying the authenticity and "freshness" of messages and does not specify any cryptographic primitives or how the freshness counter is determined (i.e., these are highly manufacturer specific). For this reason, this PR provides a `SnakeOilAuthenticator`, a set of "fake" primitives of highly dubious cryptographic quality, to demonstrate how SecOC can be implemented.

### DIFF
--- a/cantools/autosar/__init__.py
+++ b/cantools/autosar/__init__.py
@@ -1,3 +1,4 @@
 # a collection of AUTOSAR specific functionality
 
 from .end_to_end import *
+from .secoc import *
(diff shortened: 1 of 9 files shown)


andlaus:
@juleq: As usual, I'm not aware of any regressions due to this...

PR #422, opened 2022-05-06
Added code to handle Node Mapped attributes

malneni:
dbc: updated load & dump to handle REL attributes
Added dbc file with rel attributes
test_database: added function to test new feature


malneni:
Had to recreate with single commit

PR #378, opened 2022-01-10
Add type annotations to Database, Message and Signal

zariiii9003:
Add some type annotations to improve code completion in IDEs

PR #423, opened 2022-05-13
Distribute & expose type annotations per PEP 561

### LINKED ISSUE #421: Type annotations not available for code which calls cantools.
Type annotations were added to a few packages in !378 but these are not used by clients of cantools.

When I try to use cantools, mypy errors:
```
error: Skipping analyzing "cantools": module is installed, but missing library stubs or py.typed marker
```

This is because packages that include type annotations in the source code should include an empty file `py.typed` in the package (or a parent package). The mypy docs which show an example using setuptools:

https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

 See [PEP-561](https://peps.python.org/pep-0561/) for details.

Hnasar:
Type annotations were added in #378 but these are not exposed to clients of cantools:

	error: Skipping analyzing "cantools": module is installed, but
	missing library stubs or py.typed marker

Per [PEP 561], packages that include type annotations in the source code should include an empty file py.typed in the package (or a parent package). The [mypy docs] show an example using setuptools.

Diff from `pip install .; pip show -f cantools`:

	--- a/before.txt
	+++ b/after.txt
	@@ -98,6 +98,7 @@ Files:
	   cantools\logreader.py
	+  cantools\py.typed
	   cantools\subparsers\__init__.py

[PEP 561]: https://peps.python.org/pep-0561/
[mypy docs]: https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

Fixes #421

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,7 @@ setup(name='cantools',
       keywords=['can', 'can bus', 'dbc', 'kcd', 'automotive'],
       url='https://github.com/eerimoq/cantools',
       packages=find_packages(exclude=['tests']),
+      package_data={"cantools": ["py.typed"]},
       python_requires='>=3.6',
       install_requires=[
           'bitstruct>=6.0.0',


andlaus (approved):
Looks good to me. any thoughts, @juleq?

PR #427, opened 2022-05-19
Add more type annotations

zariiii9003:

PR #428, opened 2022-05-22
Improve encoding performance

zariiii9003:
Most of the perfomance benefits come from converting properties into normal attributes. Each property call results in a new stack frame which is quite expensive.

Benchmarks with some random message on my laptop:

`message.encode(data, strict=True, scaling=True)`: 49.19% improvement
`message.encode(data, strict=False, scaling=True)`: 22.98% improvement
`message.encode(data, strict=False, scaling=False)`: 60.74% improvement

This is the code i used for benchmarking:
```python3
import timeit

setup = """
import cantools
db = cantools.db.load_file("path_to_OEM_dbc")

message = db.get_message_by_name("OEM_message_name")
data = {signal.name: min(0x40, signal.maximum-1) for signal in message.signals}
"""

print(min(timeit.Timer('message.encode(data, strict=False, scaling=False)', setup=setup).repeat(100, 1000)))
```

Here is a cProfile dump for 10000 encode calls of the master branch (strict=False, scaling=True):
```
         2570002 function calls in 0.762 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    10000    0.013    0.000    0.758    0.000 message.py:863(encode)
    10000    0.009    0.000    0.740    0.000 message.py:759(_encode)
    10000    0.015    0.000    0.731    0.000 utils.py:88(encode_data)
    10000    0.077    0.000    0.706    0.000 utils.py:92(<dictcomp>)
   270000    0.493    0.000    0.614    0.000 utils.py:51(_encode_field)
   540000    0.032    0.000    0.032    0.000 {built-in method builtins.isinstance}
   540000    0.031    0.000    0.031    0.000 signal.py:249(name)
   270000    0.023    0.000    0.023    0.000 {method 'to_integral' of 'decimal.Decimal' objects}
   270000    0.018    0.000    0.018    0.000 signal.py:359(offset)
   270000    0.016    0.000    0.016    0.000 signal.py:347(scale)
   270000    0.016    0.000    0.016    0.000 signal.py:311(is_float)
    20000    0.006    0.000    0.006    0.000 {method 'pack' of 'bitstruct.c.CompiledFormatDict' objects}
    20000    0.003    0.000    0.003    0.000 {built-in method binascii.hexlify}
    10000    0.002    0.000    0.002    0.000 {built-in method builtins.hex}
    10000    0.001    0.000    0.001    0.000 {built-in method binascii.unhexlify}
    10000    0.001    0.000    0.001    0.000 {method 'rstrip' of 'str' objects}
    10000    0.001    0.000    0.001    0.000 message.py:323(is_container)
    10000    0.001    0.000    0.001    0.000 {built-in method builtins.len}
    10000    0.001    0.000    0.001    0.000 typing.py:898(cast)
```

And here is the same with my changes:
```
         660002 function calls in 0.412 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    10000    0.007    0.000    0.409    0.000 message.py:859(encode)
    10000    0.007    0.000    0.399    0.000 message.py:755(_encode)
    10000    0.009    0.000    0.392    0.000 utils.py:99(encode_data)
    10000    0.338    0.000    0.375    0.000 utils.py:51(_encode_fields)
   270000    0.020    0.000    0.020    0.000 {method 'to_integral_value' of 'decimal.Decimal' objects}
   270000    0.017    0.000    0.017    0.000 {built-in method builtins.isinstance}
    20000    0.005    0.000    0.005    0.000 {method 'pack' of 'bitstruct.c.CompiledFormatDict' objects}
    20000    0.002    0.000    0.002    0.000 {built-in method from_bytes}
    10000    0.001    0.000    0.001    0.000 {method 'to_bytes' of 'int' objects}
    10000    0.001    0.000    0.001    0.000 message.py:323(is_container)
    10000    0.001    0.000    0.001    0.000 {built-in method builtins.len}
    10000    0.000    0.000    0.000    0.000 typing.py:898(cast)
```

### DIFF
--- a/cantools/database/can/signal.py
+++ b/cantools/database/can/signal.py
@@ -210,15 +210,24 @@ class Signal(object):
                  decimal: Optional[Decimal] = None,
                  spn: Optional[int] = None
                  ) -> None:
-        self._name = name
+        #: The signal name as a string.
+        self.name: str = name
+
+        #: The scale factor of the signal value.
+        self.scale: float = scale
+
+        #: The offset of the signal value.
+        self.offset: float = offset
+
+        #: ``True`` if the signal is a float, ``False`` otherwise.
+        self.is_float: bool = is_float
+
         self._start = start
         self._length = length
         self._byte_order = byte_order
         self._is_signed = is_signed
         self._initial = initial
         self._invalid = invalid
-        self._scale = scale
-        self._offset = offset
         self._minimum = minimum
         self._maximum = maximum
         self._decimal = Decimal() if decimal is None else decimal
@@ -243,20 +252,7 @@ class Signal(object):
         self._is_multiplexer = is_multiplexer
         self._multiplexer_ids = multiplexer_ids
         self._multiplexer_signal = multiplexer_signal
-        self._is_float = is_float
         self._spn = spn
-
-    @property
-    def name(self) -> str:
-        """The signal name as a string.
-
-        """
-
-        return self._name
-
-    @name.setter
-    def name(self, value: str) -> None:
-        self._name = value
 
     @property
     def start(self) -> int:
@@ -309,18 +305,6 @@ class Signal(object):
         self._is_signed = value
 
     @property
-    def is_float(self) -> bool:
-        """``True`` if the signal is a float, ``False`` otherwise.
-
-        """
-
-        return self._is_float
-
-    @is_float.setter
-    def is_float(self, value: bool) -> None:
-        self._is_float = value
-
-    @property
     def initial(self) -> Optional[int]:
         """The initial value of the signal, or ``None`` if unavailable.
 
@@ -343,30 +327,6 @@ class Signal(object):
     @invalid.setter
     def invalid(self, value: int) -> None:
         self._invalid = value
-
-    @property
-    def scale(self) -> float:
-        """The scale factor of the signal value.
-
-        """
-
-        return self._scale
-
-    @scale.setter
-    def scale(self, value: float) -> None:
-        self._scale = value
-
-    @property
-    def offset(self) -> float:
-        """The offset of the signal value.
-
-        """
-
-        return self._offset
-
-    @offset.setter
-    def offset(self, value: float) -> None:
-        self._offset = value
 
     @property
     def minimum(self) -> Optional[float]:
@@ -555,14 +515,14 @@ class Signal(object):
                  for value, text in self._choices.items()]))
 
         return "signal('{}', {}, {}, '{}', {}, {}, {}, {}, {}, {}, '{}', {}, {}, {}, {}, {})".format(
-            self._name,
+            self.name,
             self._start,
             self._length,
             self._byte_order,
             self._is_signed,
             self._initial,
-            self._scale,
-            self._offset,
+            self.scale,
+            self.offset,
             self._minimum,
             self._maximum,
             self._unit,
--- a/cantools/database/can/signal.py
+++ b/cantools/database/can/signal.py
@@ -2,10 +2,9 @@
 import decimal
 from typing import Optional, Dict, TYPE_CHECKING, List, Any, Union
 
-from cantools.typechecking import Comments
+from cantools.typechecking import Comments, ByteOrder, Choices
 
 if TYPE_CHECKING:
-    from collections import OrderedDict
     from cantools.database.can.formats.dbc import DbcSpecifics
 
 
@@ -190,7 +189,7 @@ class Signal(object):
                  name: str,
                  start: int,
                  length: int,
-                 byte_order: str = 'little_endian',
+                 byte_order: ByteOrder = 'little_endian',
                  is_signed: bool = False,
                  initial: Optional[int] = None,
                  invalid: Optional[int] = None,
@@ -199,7 +198,7 @@ class Signal(object):
                  minimum: Optional[float] = None,
                  maximum: Optional[float] = None,
                  unit: Optional[str] = None,
-                 choices: Optional["OrderedDict[int, Union[str, NamedSignalValue]]"] = None,
+                 choices: Optional[Choices] = None,
                  dbc_specifics: Optional["DbcSpecifics"] = None,
                  comment: Optional[Union[str, Comments]] = None,
                  receivers: Optional[List[str]] = None,
@@ -210,6 +209,8 @@ class Signal(object):
                  decimal: Optional[Decimal] = None,
                  spn: Optional[int] = None
                  ) -> None:
+        # avoid using properties to improve encoding/decoding performance
+
         #: The signal name as a string.
         self.name: str = name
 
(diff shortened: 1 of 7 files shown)


zariiii9003:
There is another bottleneck in `_encode_fields` but changing this would lead to a few failing tests:
Replacing this
```python3
                    decimal = (Decimal(value) - Decimal(field.offset)) / Decimal(field.scale)
                    unpacked[field.name] = int(decimal.to_integral_value())
```
with
```python3
                    unpacked[field.name] = int((value - field.offset) / field.scale + 0.5)
```
results in
```
         390002 function calls in 0.139 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    10000    0.007    0.000    0.137    0.000 message.py:859(encode)
    10000    0.006    0.000    0.127    0.000 message.py:755(_encode)
    10000    0.009    0.000    0.122    0.000 utils.py:98(encode_data)
    10000    0.092    0.000    0.106    0.000 utils.py:51(_encode_fields)
   270000    0.015    0.000    0.015    0.000 {built-in method builtins.isinstance}
    20000    0.004    0.000    0.004    0.000 {method 'pack' of 'bitstruct.c.CompiledFormatDict' objects}
    20000    0.002    0.000    0.002    0.000 {built-in method from_bytes}
    10000    0.001    0.000    0.001    0.000 {method 'to_bytes' of 'int' objects}
    10000    0.001    0.000    0.001    0.000 message.py:323(is_container)
    10000    0.000    0.000    0.000    0.000 typing.py:898(cast)
    10000    0.000    0.000    0.000    0.000 {built-in method builtins.len}
```

For huge values the resulting raw value can be off by one. But i'd say that is acceptable.

cantools/database/can/signal.py
```diff
@@ -243,21 +256,8 @@ def __init__(self,
         self._is_multiplexer = is_multiplexer
         self._multiplexer_ids = multiplexer_ids
         self._multiplexer_signal = multiplexer_signal
-        self._is_float = is_float
         self._spn = spn
 
-    @property
-    def name(self) -> str:
```
andlaus, line 250: the performance improvement is great, but I'd like to keep the "officially supported" attributes somehow documented in the class' docstring. is it somehow possible to add a dummy member in the class for this, something like:

```python
class Message:
  # [...]
  name = "<undefined>"
  """The signal name as a string.
  """
```

(unfortunately the docstring will be ignored, but at least  the `name` attribute will be visible without instantiating an object...)

a slightly more verbose but IMO better alternative would be to keep the property for the docstring and overriding it in the constructor:

```python
class Message:
  def __init__(self, name : str):
    self.name = name

  @property
  def name(self) -> str:
        """The signal name as a string.
        """
        raise RuntimeError("Undefined property") 
```

the main drawback of this is that `help(messageObject.name)` will produce the documentation of the build-in string type instead of the docstring of the property. (`help(messageObject)` and `help(Message.name)` are fine, though.)

cantools/database/diagnostics/data.py
```diff
@@ -79,30 +73,6 @@ def byte_order(self):
     def byte_order(self, value):
         self._byte_order = value
```
andlaus, line 74: IMO it would be good to handle all "simple" properties uniformly, not just the performance-critical ones...

andlaus (commented):
nice, better performance is always a good thing! I'd prefer to continue having the properties properly documented, but if it turns out that this is infeasible, this PR should IMO be merged regardless...

cantools/database/can/signal.py
zariiii9003: The attributes with `#:`
```
        #: The signal name as a string.
        self.name: str = name

        #: The scale factor of the signal value.
        self.scale: float = scale

        #: The offset of the signal value.
        self.offset: float = offset
```
will continue to be described in the docs. But `help()` would not work though since the attributes don't have a `__doc__` method.
Alternatively we could document the attributes in the class docstring like [this](https://stackoverflow.com/a/8659919/13170690)

cantools/database/diagnostics/data.py
zariiii9003: Yes i agree. I didnt want to change more than necessary to make the review easier. Once we agree on the change i will update the other properties.

cantools/database/can/signal.py
andlaus: > The attributes with `#:`

neat. I didn't know that so far :)

> will continue to be described in the docs. But help() would not work though since the attributes don't have a __doc__ method.

While obviously not 100% optimal, I think that this is an acceptable trade-off for improved performance.

> Alternatively we could document the attributes in the class docstring like [this](https://stackoverflow.com/a/8659919/13170690)

I'd prefer to keep the documentation as close to the code as possible. That's because I am of the opinions that it is next to impossible to keep it complete and consistent otherwise. (besides, putting this into the docstring of the class would still not change the "documentation of object attributes" issue...)

cantools/database/diagnostics/data.py
zariiii9003: I converted the properties of Signal and Data. Tell me if i missed something

zariiii9003:
I also edited the decoder code a little.
Here is the profile from **before** #427 (i introduced a major performance regression by using `TypeGuard` inside `_decode_field`)
```
         1510000 function calls in 0.316 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    10000    0.004    0.000    0.313    0.000 message.py:1054(decode)
    10000    0.005    0.000    0.307    0.000 message.py:943(_decode)
    10000    0.011    0.000    0.303    0.000 utils.py:101(decode_data)
    10000    0.105    0.000    0.278    0.000 utils.py:110(<dictcomp>)
   270000    0.101    0.000    0.145    0.000 utils.py:65(_decode_field)
   540000    0.029    0.000    0.029    0.000 signal.py:249(name)
    70000    0.011    0.000    0.016    0.000 utils.py:74(<lambda>)
   270000    0.015    0.000    0.015    0.000 signal.py:422(choices)
    20000    0.009    0.000    0.009    0.000 {method 'unpack' of 'bitstruct.c.CompiledFormatDict' objects}
    80000    0.005    0.000    0.005    0.000 signal.py:347(scale)
    10000    0.005    0.000    0.005    0.000 {method 'update' of 'dict' objects}
    70000    0.004    0.000    0.004    0.000 signal.py:359(offset)
    80000    0.004    0.000    0.004    0.000 {built-in method builtins.isinstance}
    40000    0.003    0.000    0.003    0.000 signal.py:311(is_float)
    10000    0.001    0.000    0.001    0.000 {method 'is_integer' of 'float' objects}
    10000    0.001    0.000    0.001    0.000 message.py:329(is_container)
```

And here is the profile of this PR:
```
         60000 function calls in 0.070 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    10000    0.004    0.000    0.067    0.000 message.py:1040(decode)
    10000    0.003    0.000    0.062    0.000 message.py:929(_decode)
    10000    0.051    0.000    0.059    0.000 utils.py:96(decode_data)
    20000    0.008    0.000    0.008    0.000 {method 'unpack' of 'bitstruct.c.CompiledFormatDict' objects}
    10000    0.001    0.000    0.001    0.000 message.py:323(is_container)
```

cantools/database/diagnostics/data.py
andlaus: looks good. It seems like cpython does not pick it up for the class' docstring, but sphinx does. While not optimal, I think this is an acceptable trade-off...

andlaus:
I'm happy with this now. You too, @juleq?

zariiii9003:
What do you think about replacing `(Decimal(value) - Decimal(field.offset)) / Decimal(field.scale)`?
I'm wondering why this is the default. What is the use case for this? AFAIK no ECU would use decimals for encoding CAN messages.

andlaus:
> What do you think about replacing (Decimal(value) - Decimal(field.offset)) / Decimal(field.scale)?

As long as the differences due to replacing it are minor, I'm happy to sacrifice it on the altar of performance. (Since this construct predates my time with cantools, I can also only speculate why it was done that way. My money is on not even theoretically running into precision issues...)

juleq:
Yes, I think this was it. It was not only this line, but internal storage of non-ints and other code parts, too. Because there is no guarantee for an exact representation in DP, your alternative option would be to be very consistent in your rounding/formatting/comparison. Can of worms :).

andlaus:
> your alternative option would be to still use DP but be very consistent in your rounding/formatting/comparisons. Can of worms :).

I see. do you think that adding a set of floating point attributes for de-/encoding besides the Decimal ones would be feasible? i.e., the  decimal ones would be used for dumping the files, while the floating point ones for everything else?

juleq:
> But the encode function does not influence a database dump. 
Yes, thats true. Either the contributor did that for consistency with changes related to load/dump or maybe encode/decode roundtrip was a concern. Or inference of underlying number type. It is nice to retain the underlying integer type when scaling and offset would in principle allow for that. But as @andlaus mentioned, considering the signal type and some rounding strategy would also do that. Maybe the contributor had something to say in his commit for that line?
> 
> 
> I think some rounding error is to be expected when the user sets `scaling=True`, at least for anyone who is familiar with floating point data types. The decode function also just uses regular floats.
Also true. It is kind of being the unicorn when probably every other tool will probably top out at DP precision and call it a day. And that decode does not stick to this strategy speaks against the encode/decode roundtrip hypothesis.

Not sure how to proceed. Whats the performance gain? Would it be worth a shot?

andlaus:
> Not sure how to proceed.

I recommend merging this PR (#428) and leave the better decode performance for a later one. (If I understood this correctly, that's an orthogonal issue...)

andlaus:
> Removing decimal scaling reduces encoding time by 70% compared to this PR

does that mean that decoding gets faster by a factor of more than 3, or "only" by factor ~1.7?

> The decoding changes are actually already in this PR

okay, I missed that somehow *g*. If @juleq considers this change to be too risky to do in one swoop, would do mind factoring it out and make a separate pull request for it?

zariiii9003:
> does that mean that decoding gets faster by a factor of more than 3, or "only" by factor ~1.7?

For **encoding** it is factor 3.3 compared to this PR (factor 4.6 compared to master) with `strict=False, scaling=True`. Decimal scaling is only used for encoding, so removing it has no impact on decoding.
But i measured this on one message only. Numbers may vary for other messages.

> okay, I missed that somehow _g_. If @juleq considers this change to be too risky to do in one swoop, would do mind factoring it out and make a separate pull request for it?

Yeah i could.

andlaus:
@zariiii9003: we're always looking to expand the number of maintainers for cantools. If you're interested, drop me an email. (I couldn't figure out how to get in contact with you directly...)

PR #430, opened 2022-05-23
Added code to handle Node Mapped attributes

malneni:
dbc: updated load & dump to handle REL attributes
Added dbc file with rel attributes
test_database: added function to test new feature

### DIFF
--- a/cantools/database/can/formats/dbc.py
+++ b/cantools/database/can/formats/dbc.py
@@ -83,8 +83,11 @@ DBC_FMT = (
     '\r\n'
     '{cm}\r\n'
     '{ba_def}\r\n'
+    '{ba_def_rel}'
     '{ba_def_def}\r\n'
+    '{ba_def_def_rel}'
     '{ba}\r\n'
+    '{ba_rel}'
     '{val}\r\n'
     '{signal_types}\r\n'
     '{sig_group}\r\n'
@@ -357,7 +360,9 @@ class DbcSpecifics(object):
                  attributes=None,
                  attribute_definitions=None,
                  environment_variables=None,
-                 value_tables=None):
+                 value_tables=None,
+                 attributes_rel=None,
+                 attribute_definitions_rel=None):
         if attributes is None:
             attributes = OrderedDict()
 
@@ -370,10 +375,18 @@ class DbcSpecifics(object):
         if value_tables is None:
             value_tables = OrderedDict()
 
+        if attributes_rel is None:
+            attributes_rel = OrderedDict()
+
+        if attribute_definitions_rel is None:
+            attribute_definitions_rel = OrderedDict()
+
         self._attributes = attributes
         self._attribute_definitions = attribute_definitions
         self._environment_variables = environment_variables
         self._value_tables = value_tables
+        self._attributes_rel = attributes_rel
+        self._attribute_definitions_rel = attribute_definitions_rel
 
     @property
     def attributes(self):
@@ -413,6 +426,22 @@ class DbcSpecifics(object):
         """
 
         return self._environment_variables
+
+    @property
+    def attributes_rel(self):
+        """The DBC specific attribute rel as dictionary..
+
+        """
+
+        return self._attributes_rel
+
+    @property
+    def attribute_definitions_rel(self):
+        """The DBC specific attribute definitions rel as dictionary.
+
+        """
+
+        return self._attribute_definitions_rel
 
 
 class LongNamesConverter(object):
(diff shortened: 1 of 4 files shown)


juleq:
Hi @malneni,

you need to update your pull to the master head and reapply your changes. Configure your editor and your git to get the line endings right. I'll have a look then.

I would start with configuring line endings and, if you do not know how, google how to sync/update your local checkout to the head of origin master.

Best

tests/files/dbc/BU_BO_REL_.dbc
```diff
@@ -47,7 +47,6 @@ BA_DEF_ BO_  "GenMsgCycleTime" INT 0 65535;
 BA_DEF_DEF_  "GenMsgCycleTime" 0;
 BA_DEF_REL_ BU_BO_REL_  "MsgProject" ENUM  "A","B","C";
 BA_DEF_DEF_REL_ "MsgProject" "A";
-BA_REL_ "MsgProject" BU_BO_REL_ COM 1 2;
```
juleq, line 50: What is the reason for removal?

cantools/database/can/formats/dbc.py
```diff
@@ -83,8 +83,11 @@
     '\r\n'
     '{cm}\r\n'
     '{ba_def}\r\n'
+    '{ba_def_rel}'
```
juleq, line 86: Correct that line endings are handled differently here? I would prefer keeping it consistent if possible.

cantools/database/can/formats/dbc.py
```diff
@@ -1705,8 +1889,11 @@ def dump_string(database: InternalDatabase, sort_signals: type_sort_signals = SO
                           cm='\r\n'.join(cm),
                           signal_types='\r\n'.join(signal_types),
                           ba_def='\r\n'.join(ba_def),
+                          ba_def_rel="".join([elem+"\r\n" for elem in ba_def_rel]),
```
juleq, line 1892: Correct that line endings are handled differently here? I would prefer keeping it consistent if possible.

juleq (changes requested):
Please check my review comments.

juleq:
Good job, thanks. Some minor cleanup, maybe?

tests/files/dbc/BU_BO_REL_.dbc
malneni: This is not valid line. I checked with vector CANdb++. Since updated code handles this type of attributes, I need to remove it or update it with correct syntax

cantools/database/can/formats/dbc.py
malneni: Yes, Line endings are handled differently otherwise when there is no attribute related to ba_def_rel, it will print a empty newline in output dbc. All the previous dbc files does not have this dbc file so testcases will fail at  "self.assert_dbc_dump(db, filename)".

& All the dbc file sin tests are having empty spaces that matches output generated(either intentionally or unknowingly). this is no longer possible as ba_def_rel not considered before. if you open these dbc files in tools like Vector DBC++ & save the file, most of the the test cases will fail.

cantools/database/can/formats/dbc.py
malneni: yes, \r\n  should only be added if at least one element present in ba_def_rel. otherwise extra empty spaces will exists in output dbc even though there is no empty space in input. I thought about changing old implementation but I stick to feature that I thought for now. 

It is not possible to implement line endings as done previously.

tests/files/dbc/BU_BO_REL_.dbc
jak72: Move the line BA_DEF_DEF_  "GenMsgCycleTime" 0; 
to come after BA_DEF_REL_ BU_BO_REL_  "MsgProject" ENUM  "A","B","C";
keep BA_REL_ "MsgProject" BU_BO_REL_ COM 1 2;
And file opens correct in Candb++.
Candb++ is sensitive for order of DEF_ and DEF_DEF_ s in dbc file.

andlaus:
@juleq: do you consider this PR ready to be merged? If yes, please do so. (I still don't know what node mapped attributes are, but that's just my ignorance regarding DBC...)

tests/files/dbc/BU_BO_REL_.dbc
malneni: > @jak72 Yes, that is true. @malneni can you reorder to fix and use your CANdb++ to check that it loads properly?
I have handled Node-mapped signals but apparently there is Node-mapped Messages also. I have not handled this so I get test fail in BU_BO_REL_.dbc. I need more time to complete second feature. 
So, Can you consider these 2 as separate feature? & two separate pull request?

juleq:
@malneni I did not check if the inconsistency in the line endings I have mentioned in the review is really required. Could you make a statement if this is required because of different formatting of these attributes or if it is just a simple inconsistency that you will fix?

Afterwards I can merge. And we would naturally also appreciate the mapped message support. I have not used those either, but more complete dbc support -> more good. Thank you for contributing.

@andlaus That answer was pending. I think there is nothing else preventing a merge.

malneni:
> @malneni I did not check if the inconsistency in the line endings I have mentioned in the review is really required. Could you make a statement if this is required because of different formatting of these attributes or if it is just a simple inconsistency that you will fix?
> 
> Afterwards I can merge. And we would naturally also appreciate the mapped message support. I have not used those either, but more complete dbc support -> more good. Thank you for contributing.
> 
> @andlaus That answer was pending. I think there is nothing else preventing a merge.

I will fix the inconistancies (for line endings) in the next feature update. New way of lne endings is really required for these attibutes.

PR #431, opened 2022-05-25
Add Support for Dumping Database as SYM File

samc1213:
* Logic to convert cantools database to sym file
* Basic unit tests written - may be limitations converting from specific formats to SYM, but general functionality is implemented

### DIFF
--- a/cantools/database/__init__.py
+++ b/cantools/database/__init__.py
@@ -248,6 +248,8 @@ def dump_file(database,
         newline = ''
     elif database_format == 'kcd':
         output = database.as_kcd_string(sort_signals=sort_signals)
+    elif database_format == 'sym':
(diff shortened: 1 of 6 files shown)


andlaus (approved):
looks good. @samc1213: To me it seems like this does not affect the "core"of cantools, but only the SYM file format support, right?  If so, I recommend merging this @juleq...

andlaus:
>  If so, I recommend merging this ...

... after the mypy issues have been fixed (obviously)

samc1213:
@andlaus there are no core changes, only affects SYM files in an additive manner. I want to add more tests and fix mypy before merging. Thanks for your prompt review

samc1213:
@andlaus testing with the Jopp 6.0 file helped find some bugs. I think this is ready for merge

PR #432, opened 2022-05-25
Use floating point scaling in encoding

zariiii9003:
See discussion in #428

### DIFF
--- a/cantools/database/utils.py
+++ b/cantools/database/utils.py
@@ -59,20 +59,17 @@ def _encode_fields(fields: Sequence[Union["Signal", "Data"]],
         value = data[field.name]
 
         if isinstance(value, (float, int)):
+            _transform = float if field.is_float else round
             if scaling:
-                if field.is_float:
-                    unpacked[field.name] = (value - field.offset) / field.scale
-                    continue
-
                 if field.offset == 0 and field.scale == 1:
                     # treat special case to avoid introduction of unnecessary rounding error
-                    unpacked[field.name] = round(value)
+                    unpacked[field.name] = _transform(value)
                     continue
 
-                unpacked[field.name] = round((value - field.offset) / field.scale)
+                unpacked[field.name] = _transform((value - field.offset) / field.scale)
                 continue
 
-            unpacked[field.name] = value if field.is_float else round(value)
+            unpacked[field.name] = _transform(value)
             continue
 
         unpacked[field.name] = field.choice_string_to_number(str(value))
--- a/cantools/database/utils.py
+++ b/cantools/database/utils.py
@@ -59,17 +59,20 @@ def _encode_fields(fields: Sequence[Union["Signal", "Data"]],
         value = data[field.name]
 
         if isinstance(value, (float, int)):
-            dtype = float if field.is_float else int
             if scaling:
-                if field.offset == 0 and field.scale == 1:
-                    # treat special case to avoid introduction of unnecessary rounding error
-                    unpacked[field.name] = dtype(value)
+                if field.is_float:
+                    unpacked[field.name] = (value - field.offset) / field.scale
                     continue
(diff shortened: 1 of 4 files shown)


tests/test_database.py
```diff
@@ -547,24 +547,24 @@ def test_motohawk_encode_decode(self):
                     'AverageRadius': 0,
                     'Enable': 0
                 },
-                b'\x01\x00\x20\x00\x00\x00\x00\x00'
+                b'\x01\x00\x40\x00\x00\x00\x00\x00'
```
andlaus, line 550: any idea what the reason for this is?

cantools/database/utils.py
```diff
@@ -60,16 +59,17 @@ def _encode_fields(fields: Sequence[Union["Signal", "Data"]],
...
-                else:
-                    decimal = (Decimal(value) - Decimal(field.offset)) / Decimal(field.scale)
-                    unpacked[field.name] = int(decimal.to_integral_value())
+                if field.offset == 0 and field.scale == 1:
+                    # treat special case to avoid introduction of unnecessary rounding error
+                    unpacked[field.name] = dtype(value)
                     continue
 
-            unpacked[field.name] = value
+                unpacked[field.name] = dtype((value - field.offset) / field.scale)
```
andlaus, line 69: I'm wondering: if `dtype` is `int`,  should this be rounded instead of truncated? (I'm eternally confused by the signal scaling code...)

tests/test_database.py
andlaus, line 550: any idea what the reason for this is?

andlaus (commented):
From the code quality perspective I don't see any regression or improvement coming with this change. Given that the performance improvements seem to be substantial, I'd like to see this merged once there is an explanation for the change of the expected result of the second `test_motohawk_encode_decode` test...

cantools/database/utils.py
zariiii9003: You're right, fixed.

tests/test_database.py
zariiii9003: It's unnecessary with rounding. I reversed the changes

cantools/database/utils.py
```diff
@@ -64,12 +63,16 @@ def _encode_fields(fields: Sequence[Union["Signal", "Data"]],
                 if field.is_float:
                     unpacked[field.name] = (value - field.offset) / field.scale
                     continue
-                else:
-                    decimal = (Decimal(value) - Decimal(field.offset)) / Decimal(field.scale)
-                    unpacked[field.name] = int(decimal.to_integral_value())
+
+                if field.offset == 0 and field.scale == 1:
+                    # treat special case to avoid introduction of unnecessary rounding error
+                    unpacked[field.name] = round(value)
```
andlaus, line 69: now this is (probably) only correct if `dtype` == `int`? I guess the simplest fix is to introduce 
```python
convert = float if field.is_float else round
```
possibly you can come up with a better name?

zariiii9003: fixed

andlaus (approved):
no complaints from my side anymore. (finally.) thanks for finding and fixing that bottleneck!

juleq:
I have just one question remaining: The help still informs the user about lossless decimal options for min/max/scale/offset. If the user chooses decimal there, it would now do the lossy encode/decode, right? If so, I think that this is probably appropriate, but maybe hinting at that in the docs will prevent a surprise.

There is still a benefit to decimal min/max/scale/offset because numbers that are not representable in DP will be retained there, e.g. for the loading/dumping roundtrip?

zariiii9003:
Which part of the docs do you mean?

juleq:
Agreed. Merge and release?

juleq:
I'll just wait :).

Issue #434, opened 2022-05-30
IO Control DIDs not properly parsed from cdd


1atabey1:
# Issue
It seems like the cdd parser currently ignores any *Control States* parameters for diagnostic instances of type *IO Control* and only parses values from the *ControlEnableMaskRecord* into data objects. 

# Steps to Reproduce
The attached archive contains a cdd with a few IOControl DIDs, I parsed them with the following code
```python
relevant_dids = [0x0300, 0x0301, 0x0302]
db_data = cantools.db.load_file("SBS_Test.cdd")
relevant_dids = [did for did in db_data.dids if did.identifier in relevant_dids]
```
## CANdelaStudio Reference
Looking at e.g. DID 0x0302 shows no data with cantools but does have the *Control State* `SafeState` of type `off/on/invalid` when checked in CANdelaStudio Ver. 10RP4
![image](https://user-images.githubusercontent.com/55498083/170988455-2c780ad0-c2b0-4c00-88a9-7d4f0681ef44.png)
![image](https://user-images.githubusercontent.com/55498083/170988464-473487bb-d681-4a68-b00c-e58449f17184.png)

# Files
[SBS_Test.zip](https://github.com/cantools/cantools/files/8798023/SBS_Test.zip)

1atabey1:
Nevermind, this is not a bug.
Just had a quick look at the ISO14229-1 again, which clearly states that _"The
controlEnableMaskRecord shall not be supported when the dataIdentifier to be controlled consists of only a single
parameter."_
Adding a second value to the EnableMask fixes the issue and the file is also correctly parsed by the python cantools - closing this issue again.

1atabey1:
I take everything back, the Control States always need to be parsed and the mask is optional, I created a small pull request with a proposed fix: https://github.com/cantools/cantools/pull/436

Still not perfect since it seems to have issues with parsing proper bitfields in the enableMask but I can already work with the proposed changes. IMHO the whole diagnostics handling may need some significant refactoring, if anyone is already planning on doing this, I'd be happy to help out. Otherwise I would also gladly propose and discuss new concepts.

PR #436, opened 2022-05-30
Fix parsing of referenced data in CDDs

1atabey1:
Some data objects are not directly part of the variant specific DID XML elements but are instead referencing to another location, this pr helps with that.
All diagnostics unit tests are still passing and a new test for this functionality has been added.

### DIFF
--- a/cantools/database/diagnostics/formats/cdd.py
+++ b/cantools/database/diagnostics/formats/cdd.py
@@ -1,5 +1,6 @@
 # Load and dump a diagnostics database in CDD format.
 import logging
+from typing import Dict
 
 from xml.etree import ElementTree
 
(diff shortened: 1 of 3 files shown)


1atabey1:
This addresses issue https://github.com/cantools/cantools/issues/434

andlaus (commented):
The PR looks fine to me. That said, I've got zero experience with non-UDS diagnostics.

juleq:
Could you check why the cdd diff is so massive. Github will not render...

1atabey1:
> Could you check why the cdd diff is so massive. Github will not render...
> 
> And linter is unhappy. Maybe have a look there, too.

Just pushed a change that should make the linter happy.

The cdd was generated using Vector's CANdela Studio, I already deleted all diagnostic instances that I could.. unfortunately vector is not really known for their easy readable and lightweight formats (: 
I can try and modify one of the existing cdds but that would probably also cause a huge diff (and a lot of unit tests that I would have to fix) so not sure if that's worth it

juleq:
Thats fine. Just wanted to make sure the diff size is not due to some formatting issue. Merging, thank you.

PR #417, opened 2022-04-19
implement decoding of partial messages

andlaus:
This (probably) fixes [#407](https://github.com/cantools/cantools/issues/407). After this PR is merged, cantools depends on bitstruct >= 8.14.0...

Andreas Lauser &lt;<andreas.lauser@mercedes-benz.com>&gt;, on behalf of [MBition GmbH](https://mbition.io/).
[Provider Information](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md)

PR #437, opened 2022-05-31
Small improvements after #417

zariiii9003:
A type annotation was overwritten by #417.

Also: Shouldn't `allow_truncated` be False by default?

### DIFF
--- a/cantools/database/utils.py
+++ b/cantools/database/utils.py
@@ -104,21 +104,17 @@ def decode_data(data: bytes,
 
     unpacked = formats.big_endian.unpack(bytes(data), allow_truncated=allow_truncated)
 
-    if len(data) < expected_length and allow_truncated:
+    if allow_truncated and len(data) < expected_length:
         # to deal with truncated little-endian signals, we have to pad
         # the raw data and remove the spurious signals after unpacking
         # (i.e., before the signal values themselfs get decoded)
-        le_data = bytes(bytearray([0x00]*(expected_length - len(data)))
-                        + data[::-1])
-
+        le_data = data[::-1].rjust(expected_length, b"\xFF")
         le_unpacked = formats.little_endian.unpack(le_data)
 
         # remove spurious little endian signals
+        valid_bit_count = len(data) * 8
         for signal in fields:
-            if signal.byte_order != "little_endian":
-                continue
-
-            if signal.start + signal.length > len(data)*8:
+            if signal.start + signal.length > valid_bit_count:
                 del le_unpacked[signal.name]
 
         unpacked.update(le_unpacked)
@@ -127,27 +123,30 @@ def decode_data(data: bytes,
 
     decoded = {}
     for field in fields:
-        if field.name not in unpacked:
-            continue
+        try:
+            value = unpacked[field.name]
 
-        value = unpacked[field.name]
+            if decode_choices:
+                try:
+                    decoded[field.name] = field.choices[value]  # type: ignore[index]
+                    continue
+                except (KeyError, TypeError):
+                    pass
 
-        if decode_choices:
-            try:
-                decoded[field.name] = field.choices[value]  # type: ignore[index]
+            if scaling:
+                decoded[field.name] = field.scale * value + field.offset
                 continue
-            except (KeyError, TypeError):
-                pass
+            else:
+                decoded[field.name] = value
 
-        if scaling:
-            decoded[field.name] = field.scale * value + field.offset
-            continue
-        else:
-            decoded[field.name] = value
+        except KeyError:
+            if not allow_truncated:
+                raise
 
     return decoded
 
-def create_encode_decode_formats(datas, number_of_bytes):
+
+def create_encode_decode_formats(datas: Sequence[Union["Data", "Signal"]], number_of_bytes: int) -> Formats:
     format_length = (8 * number_of_bytes)
 
     def get_format_string_type(data: Union["Data", "Signal"]) -> str:


cantools/database/utils.py
```diff
@@ -104,21 +104,17 @@ def decode_data(data: bytes,
...
-        le_data = bytes(bytearray([0x00]*(expected_length - len(data)))
-                        + data[::-1])
-
+        le_data = data[::-1].rjust(expected_length, b"\xFF")
         le_unpacked = formats.little_endian.unpack(le_data)
 
         # remove spurious little endian signals
+        valid_bit_count = len(data) * 8
         for signal in fields:
-            if signal.byte_order != "little_endian":
```
zariiii9003, line 118: Is this check necessary? All signals in the padded range should be removed, right?

zariiii9003:
I looked again and it is indeed False by default. I am a little confused now, sorry 😄

cantools/database/utils.py
zariiii9003: So removing the check is fine? le_unpacked cannot contain big endian fields anyway, right?

andlaus: yes and yes. the reason for that if statement might also be due to the history of that patch: if using padding for both little and big endian signals and decoding them into the same dictionary, the condition does not work for BE signals... (I don't remember anymore if that was the case, though.)

be aware that for systems with lots of truncated frames exhibiting little endian signals, reducing the number of required python instructions to be executed as outlined above probably yields a sizeable performance improvement. That said, I think this is an edge case and I consider this solution about as clunky as the current code but slightly harder to read. I'm fine with whatever you consider to be better...

zariiii9003:
I pushed a change and now this test fails:

> FAIL: test_motohawk_decode_truncated (tests.test_database.CanToolsDatabaseTest)
Decoding truncated frames.
  File "/home/runner/work/cantools/cantools/tests/test_database.py", line 638, in test_motohawk_decode_truncated
    self.assertEqual(decoded, {'AverageRadius': 0.0, 'Enable': 'Disabled'})
AssertionError: {'Enable': 'Disabled', 'AverageRadius': 0.0, 'Temperature': 251.43} != {'AverageRadius': 0.0, 'Enable': 'Disabled'}

However, the temperature field should be in the valid range? 

> BO_ 496 ExampleMessage: 8 PCM1
 SG_ Temperature : 0|12@0- (0.01,250) [229.52|270.47] "degK"  PCM1,FOO
 SG_ AverageRadius : 6|6@0+ (0.1,0) [0|5] "m" Vector__XXX
 SG_ Enable : 7|1@0+ (1,0) [0|0] "-" Vector__XXX

All fields of the message are inside the available two bytes.

andlaus:
hm, I vaguely remember that there was an issue, but I forgot what it was. While all signals have *some* bits in the first two bytes, the `Temperature` signal is not fully contained in it:
```
$ cantools dump tests/files/dbc/motohawk.dbc 
================================= Messages =================================

  ------------------------------------------------------------------------

  Name:           ExampleMessage
  Id:             0x1f0
  Length:         8 bytes
  Cycle time:     - ms
  Senders:        PCM1
  Layout:

                          Bit

             7   6   5   4   3   2   1   0
           +---+---+---+---+---+---+---+---+
         0 |<-x|<---------------------x|<--|
           +---+---+---+---+---+---+---+---+
             |                       +-- AverageRadius
             +-- Enable
           +---+---+---+---+---+---+---+---+
         1 |-------------------------------|
           +---+---+---+---+---+---+---+---+
         2 |----------x|   |   |   |   |   |
     B     +---+---+---+---+---+---+---+---+
     y               +-- Temperature
     t     +---+---+---+---+---+---+---+---+
     e   3 |   |   |   |   |   |   |   |   |
           +---+---+---+---+---+---+---+---+
         4 |   |   |   |   |   |   |   |   |
           +---+---+---+---+---+---+---+---+
         5 |   |   |   |   |   |   |   |   |
           +---+---+---+---+---+---+---+---+
         6 |   |   |   |   |   |   |   |   |
           +---+---+---+---+---+---+---+---+
         7 |   |   |   |   |   |   |   |   |
           +---+---+---+---+---+---+---+---+

  Signal tree:

    -- {root}
       +-- Enable
       +-- AverageRadius
       +-- Temperature

  Signal choices:

    Enable
        0 Disabled
        1 Enabled

  ------------------------------------------------------------------------
```

I suppose its caused by either a incorrect LE codec, or our old friend sawtooth bit ordering is to blame. I think it is the latter...

cantools/database/utils.py
```diff
@@ -102,52 +102,48 @@ def decode_data(data: bytes,
...
-
-        le_unpacked = formats.little_endian.unpack(le_data)
-
-        # remove spurious little endian signals
+    actual_length = len(data)
+    if allow_truncated and actual_length < expected_length:
+        data = data.ljust(expected_length, b"\xFF")
+
+    unpacked = {
+        **formats.big_endian.unpack(bytes(data)),
```
andlaus, line 110: this causes the test failure you've stumbled over: big endian signals use sawtooth bit ordering, so it is really awkward to find out which bit ranges they occupy using a sane ordering. that was probably the reason why I separated the handling of BE and LE signals in the first place...

andlaus: (truncated BE can done directly by `bitstruct`)

zariiii9003: CANdb++ agrees with you. Interestingly it shows the startbit as 21 which to me makes it easier to reason about
image

juleq: @zariiii9003 Go to Options->Settings if this is not sufficiently confusing already :). I wonder why they left out the option to use roman numbering...

zariiii9003: @juleq that's true 😄 
But i fixed it and learned something new. Compared to master branch it is 4% slower with `decode(b'\x00\x11', decode_choices=False, scaling=False, allow_truncated=True)`. With scaling enabled the performance is equal to the master branch.
I can revert this though if you want to keep using the bitstruct parameter.

cantools/database/utils.py
```diff
@@ -115,7 +115,13 @@ def decode_data(data: bytes,
         # remove signal that are outside available data bytes
         valid_bit_count = actual_length * 8
         for signal in fields:
-            if signal.start + signal.length > valid_bit_count:
+            if signal.byte_order == "little_endian":
+                sequential_startbit = signal.start
+            else:
+                # calculate startbit with inverted indices
+                sequential_startbit = 7 - signal.start % 8 + (signal.start // 8) * 8
```
andlaus: just wondering: this seems to be the same as [`sawtooth_to_network_bitnum()`](https://github.com/cantools/cantools/blob/master/cantools/database/utils.py#L258)?

cantools/database/utils.py
andlaus: > I can revert this though if you want to keep using the bitstruct parameter.

I vote for keeping your version because it uses the same mechanism for everything. (and it can use an older `bitstruct` version...)

cantools/database/utils.py
zariiii9003: So that's what that function is for... yes, it's the same. The function call increases the performance penalty from ~4% to ~8% but improves readability. Should i change it?

andlaus: ok, I did not expect that consequence. Leave it then. (in C++, the performance penalty of such a function call would be zero as it would be inlined. I suppose that writing well performing code with python comes with quite a few specific pitfalls...)

andlaus: I agree. Maybe you can extend the comment with a reference to the function and why it was not used. (Only if you want.)

zariiii9003: done

andlaus:
@juleq: this should be merged IMO...

PR #438, opened 2022-06-01
minor bug fixes and improvements

andlaus:
the most significant change is to fix decoding of truncated multiplexer messages if the multiplexor signal itself is missing. (mea culpa! if just "content" signals were missing, it worked fine...)

besides this, I added an officially sanctioned way to calculate the authenticator value for a message to the AUTOSAR SecOC code and extended the decode subparser to print the raw data of the individual chunks of container messages...

### DIFF
--- a/cantools/autosar/secoc.py
+++ b/cantools/autosar/secoc.py
@@ -15,6 +15,36 @@ class SecOCError(Error):
     """
     pass
 
+def compute_authenticator(raw_payload: Union[bytes, bytearray],
+                          dbmsg: Message,
+                          authenticator_fn: Callable[[Message,
+                                                      bytearray,
+                                                      int],
+                                                     bytearray],
+                          freshness_value: int) \
+                          -> bytearray:
+    """Given a byte-like object that contains the encoded signals to be
+    send, compute the full authenticator SecOC value.
+    """
+
+    if dbmsg.autosar is None or dbmsg.autosar.secoc is None:
+        raise SecOCError(f'Message "{dbmsg.name}" is not secured')
+
+    secoc_props = dbmsg.autosar.secoc
+    n_fresh = secoc_props.freshness_bit_length
+    payload_len = secoc_props.payload_length
+
+    # build the data that needs to be passed to authentificator function
+    auth_data = bitstruct.pack(f'u16' # data ID
+                               f'r{payload_len*8}' # payload to be secured
+                               f'u{n_fresh}', # freshness value
+                               secoc_props.data_id,
+                               raw_payload[:payload_len],
+                               freshness_value)
+
+    # compute authenticator value
+    return authenticator_fn(dbmsg, auth_data, freshness_value)
+
 def apply_authenticator(raw_payload: Union[bytes, bytearray],
                         dbmsg: Message,
                         authenticator_fn: Callable[[Message,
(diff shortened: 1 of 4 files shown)


andlaus:
@juleq: I think this is merge-ready. as always, reviewes are highly appreciated anyway...

andlaus:
alright. now even mypy is happy. @zariiii9003: I had to resort to `# type: ignore` comments twice. if you know a better way of conveying to the type checker that a given attribute is of a certain type (I'm looking for something akin to `assert isinstance(decoded_message, ContainerDecodeResultType)` which works for Parameterized generics), please let me know...

zariiii9003:
> alright. now even mypy is happy. @zariiii9003: I had to resort to `# type: ignore` comments twice. if you know a better way of conveying to the type checker that a given attribute is of a certain type (I'm looking for something akin to `assert isinstance(decoded_message, ContainerDecodeResultType)` which works for Parameterized generics), please let me know...

The return type of `decode()` is indeed difficult to handle. I tried to use `typing.overload` to make the return type dependent on the function arguments but gave up in the end.
If you use ignore, you should put the error type in square brackets so other errors are not silenced.

andlaus:
thanks for the suggestion. While also not very elegant, I went with adding specific decode functions for containers and normal messages, so the ignores are no longer required in this instance...

andlaus:
@zariiii9003: can you have a quick look at this?

cantools/autosar/secoc.py
```diff
@@ -15,6 +15,36 @@ class SecOCError(Error):
     """
     pass
 
+def compute_authenticator(raw_payload: Union[bytes, bytearray],
+                          dbmsg: Message,
+                          authenticator_fn: Callable[[Message,
```
zariiii9003, line 20: You could create a type alias for the authenticator_fn

cantools/autosar/secoc.py
```diff
@@ -15,6 +15,36 @@ class SecOCError(Error):
...
+                                                      bytearray,
+                                                      int],
+                                                     bytearray],
+                          freshness_value: int) \
+                          -> bytearray:
+    """Given a byte-like object that contains the encoded signals to be
+    send, compute the full authenticator SecOC value.
```
zariiii9003, line 27:
```suggestion
    sent, compute the full authenticator SecOC value.
```

cantools/autosar/secoc.py
```diff
@@ -15,6 +15,36 @@ class SecOCError(Error):
     """
     pass
 
+def compute_authenticator(raw_payload: Union[bytes, bytearray],
```
zariiii9003, line 18: You can just use `bytes`. The [docs](https://docs.python.org/3/library/typing.html#typing.ByteString) say:

> As a shorthand for this type, [bytes](https://docs.python.org/3/library/stdtypes.html#bytes) can be used to annotate arguments of any of the types mentioned above.

cantools/database/can/message.py
```diff
@@ -1085,19 +1064,33 @@ def decode(self,
...
-                return self._decode_contained(data,
-                                              decode_choices,
-                                              scaling,
-                                              allow_truncated)
-            else:
-                raise DecodeError(f'Message "{self.name}" is a container '
-                                  f'message, but decoding such messages has '
-                                  f'not been enabled by setting the '
-                                  f'decode_containers parameter to True')
+        if self.is_container and decode_containers:
```
zariiii9003, line 1067: Evaluate the computationally inexpensive boolean first.

```suggestion
        if decode_containers and self.is_container:
```

cantools/autosar/secoc.py
andlaus: ah, cool I didn't know that...

PR #442, opened 2022-06-17
Fix test_plot to pass it in large time difference zone and fix misspelling 

rshirono-c:
I fixed test to pass it in master branch in large time difference zones and fix misspelling.

1.
Master branch doesn't pass make test in Japan or other large time difference zones from GMT.
This is caused by hard coding dates in test_plot.py. 
2.
There are some misspellings and codespelling outputs error.


juleq:
Thanks

PR #388, opened 2022-01-27
change the default for pruning the names of signal choices to False

andlaus:
Although the results exhibit quite a bit less visual clutter when pruning, defaulting to True caused too much trouble for too many existing users...

Andreas Lauser &lt;<andreas.lauser@daimler.com>&gt;, on behalf of [MBition GmbH](https://mbition.io/).
[Provider Information](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md)

PR #451, opened 2022-07-07
Change other prune_choices defaults to False

mon:
#388 missed a couple


juleq:
Thanks.

andlaus:
thanks for the catch. I wonder how this fell through the crack when the remaining occurences of this where changes back, but I suppose that none of the tests calls these functions directly...

PR #456, opened 2022-07-18
Extended formatting options for DBC serialisation

mon:
In this commit I've added to `Database.as_dbc_string`:
- `sort_attribute_signals`: for sorting non-message signal definitions in comments, attributes, value tables etc
- `sort_attributes`: for sorting `BA_` entries, given a list of tuples for each attribute type (dbc/node/message/signal)
- `sort_choices`: For sorting the values inside choices. Convention in existing files seems to be by value, descending, but I've left the default to not sort
- `shorten_long_names`: To disable the long name truncation/conversion

I also added some additional sorters, and examples for the new sorters to utils.py, as well as tests.

Hopefully the style is OK and you're happy with the general architecture! `sort_attributes` required some big changes.
`sort_attribute_signals` might need a better name, considering it affects several sections of the generated .dbc.

### DIFF
--- a/cantools/database/can/database.py
+++ b/cantools/database/can/database.py
@@ -19,7 +19,13 @@ from .internal_database import InternalDatabase
 from .message import Message, EncodeInputType, DecodeResultType
 from .node import Node
 from ..errors import DecodeError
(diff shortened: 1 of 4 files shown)


andlaus (approved):
looks good

PR #461, opened 2022-08-01
ARXML: refactor the compu method handling code

andlaus:
The goal is to be more consistent and having less redundant code. The biggest functional change is that enumeration values do no longer count to the range of signals. This is because enumeration values are specified for internal values, whereas `signal.minimum` and `signal.maximum` represent physical ones.
    
Besides this, the parser no longer accepts ARXML files that specify more than a single value for given single enumeration and it produces a warning if more than one linear scale is defined for a signal (if this happens, most likely the computation method is a piecewise linear function instead of a linear one. Adding support for piecewise linear functions is probably not worth the effort because it seems to be an rather uncommon feature and it would change major assumptions of cantools' underlying data model.)
    
Finally, the string to number parsing function now interprets the strings "true" and "false" as 1 and 0. While IMO this is conceptionally questionable (such values should be texttables), it is a pragmatic solution because I've got quite a few (older) files which do this...


andlaus:
@juleq: As usual, I'm not aware of any showstoppers here, but a review would be highly appreciated anyway...

juleq:
Have just scanned through shallowly. Not really competent to comment on ARXML though. F-string cleanup looks fine. Merging.

PR #471, opened 2022-08-21
Set canfd flag by vframeformat

ytwpg:
set "is_fd" flag if VFrameFormat is defined as StandardCAN_FD or ExtendedCAN_FD

For unit test, use existing foobar.dbc that includes a FD message and add test criteria for that in tests/test_database.py


cantools/database/can/formats/dbc.py
```diff
@@ -1639,7 +1639,14 @@ def get_protocol(frame_id_dbc):
             except (KeyError, TypeError):
                 frame_format = None
 
-        if frame_format == 'J1939PG':
+        return frame_format
+
+    def get_protocol(frame_id_dbc):
```
andlaus, line 1644:
```suggestion
    def get_protocol(frame_format_dbc):
```

(in my mind, frame id is more or less synonymous with the arbitration ID...)

andlaus (commented):
looks good from a code quality perspective. @juleq is the expert about the nitty-gritty DBC details, though...

andlaus:
can you add write support for this? (a.k.a. the DBC module's [`dump_string()`](https://github.com/cantools/cantools/blob/master/cantools/database/can/formats/dbc.py#L1823) function)

ytwpg:
>

ytwpg:
@andlaus 
sorry my late response...
I will !!

PR #477, opened 2022-08-29
Bugfix: BU_BO_REL causes 'index out of range' error

malneni:


juleq:
Ahh. A subtle one. Could you check why the test is failing?

`test_relation_message_attributes (tests.test_database.CanToolsDatabaseTest) ... FAIL `

malneni:
I know the problem: Line endings in new dbc file BU_BO_REL_Message.dbc  (LF)
when uploaded to github, Line ending changes from windows (CR LF) to
unix(LF) type. I do not know the best solution for it but I have a
workaround.
I will upload the file in Mac instead of windows & set the line ending
style to windows in a text editor (may be sublime text)

andlaus:
> Line endings in new dbc file BU_BO_REL_Message.dbc  (LF)
when uploaded to github, Line ending changes from windows (CR LF) to
unix(LF) type.

sounds like we should instruct git to leave the line endings of `*.dbc` files alone or maybe even tell it to go for \r\n: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#example

juleq:
@andlaus Yes, going for \r\n for dbc is appropriate because the proprietary tool for it is Windows exclusive. IDK if we have any configuration in place in this regard.

@malneni Do you think the line endings were changed by your git client (local repo) or by the GH server (remote origin)?

malneni:
> 

I do not know what causes. I use github desktop. For now I fixed it using Mac

juleq:
Great, thank you.

PR #479, opened 2022-09-06
PCAN trace (.trc) format to logreader

tuna-f1sh:
Using https://www.peak-system.com/produktcd/Pdf/English/PEAK_CAN_TRC_File_Format.pdf as a reference, adds all PCAN trace formats to the logreader and auto-detector.

There is also a minor change to the `CandumpDefaultLogPattern` to ensure valid 2 char byte and adds lazy operators in attempt to speed it up a bit.

Only doctests added for now. File tests can be added if deemed appropriate. I wanted to see if there was interest for this PR before committing more time.


juleq:
Nice. All for it 👍. File test would be good to ensure we retain that functionality.

tuna-f1sh:
Cool I'll add a test and rebase to remove the extra commits. I work for a company where most people use the PCAN tools. Myself I use gs_usb/SocketCAN and this module, which is what lead me to add the support so I can do stuff with their files!

juleq:
Looking good. If you confirm that it is done, I'll merge. Will also do a release.

tuna-f1sh:
It’s good to go 👍

Issue #469, opened 2022-08-19
dump re-direct to file is raising an error


Ahelion:
Hello,
While dump a dbc file(with the nice format) to file, i get this:
LINES value must be >= 2 and <= 609: got 1
initscr(): Unable to create SP

command: cantools dump d:\mydbc.dbc >file.txt.

BTW: I tried to call the layout _tring function from a script and python said it was not part of the message class..

andlaus:
weird: I do this all the time and so far it never failed. I'm on Linux, though, so it might be a Windows specific issue?

does `cantools --debug dump d:\mydbc.dbc > foo.txt` produce a more meaningful error?

Ahelion:
> weird: I do this all the time and so far it never failed. I'm on Linux, though, so it might be a Windows specific issue?
> 
> does `cantools --debug dump d:\mydbc.dbc > foo.txt` produce a more meaningful error?

Hello, sorry for taking so long to reply. A picture says more than 1000 words, right?
![image](https://user-images.githubusercontent.com/672974/190873899-efe1fe6f-62bc-41a3-930e-3b7ca883fb30.png)

andlaus:
ok. this seems to be an issue of the ncurses bindings on windows (it seems to terminate the process instead of raising a python exception). luckily the `dump` subparser does not really need `ncurses`, cf #482. can you check if it does the trick for you?

Ahelion:
> ok. this seems to be an issue of the ncurses bindings on windows (it seems to terminate the process instead of raising a python exception). luckily the `dump` subparser does not really need `ncurses`, cf #482. can you check if it does the trick for you?

Hello,
I had some issues installing for some reason, I changed the version because I was confused what i was installing, that is why the picture shows 36.4.1.
I cannot confirm that it works.
![image](https://user-images.githubusercontent.com/672974/191077741-f74377ba-c4fd-40ac-858f-e4889e0fcd81.png)

Ahelion:
> > ok. this seems to be an issue of the ncurses bindings on windows (it seems to terminate the process instead of raising a python exception). luckily the `dump` subparser does not really need `ncurses`, cf #482. can you check if it does the trick for you?
> 
> Hello, I had some issues installing for some reason, I changed the version because I was confused what i was installing, that is why the picture shows 36.4.1. I cannot confirm that it works. ![image](https://user-images.githubusercontent.com/672974/191077741-f74377ba-c4fd-40ac-858f-e4889e0fcd81.png)

Again me.

[...]

PR #482, opened 2022-09-19
do no longer use the ncurses module in the dump subparser

andlaus:
it seems to cause issues when trying to redirect the output of the subparser on windows. Since it is only used to determine the width of the terminal anyway, let's determine that using `os.get_terminal_size()`.


andlaus:
this hopefully fixes #469. @Ahelion: can you verify that your issue is gone with this?

PR #484, opened 2022-09-20
Fix 'Assertation' NameError typo in test_database

Hnasar:
Found with flake8


juleq:
Ha. That is the second exception inception I come across today :).

Thanks for fixing.

PR #485, opened 2022-09-20
Add missing crccheck dependency to setup.py

Hnasar:
crccheck is actually used in the cantools package, so it should be in install_requires.

https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/


juleq:
Yepp, thanks.

PR #488, opened 2022-10-07
Bugfix in handling more than one node mapped attribute per signal

malneni:
dbc: corrected indentation error that occurred in [https://github.com/cantools/cantools/pull/477](477)


juleq:
Thanks.

PR #492, opened 2022-10-13
A few bug fixes

zariiii9003:
@andlaus

### DIFF
--- a/cantools/autosar/end_to_end.py
+++ b/cantools/autosar/end_to_end.py
@@ -37,7 +37,7 @@ def compute_profile2_crc(payload : Union[bytes, bytearray],
         assert msg.autosar.e2e.data_ids is not None
 
         protected_len = msg.autosar.e2e.payload_length
-        seq_counter = (payload[1]&0xf0) >> 4
+        seq_counter = payload[1] & 0xf
         data_id = msg.autosar.e2e.data_ids[seq_counter]
     else:
         protected_len = len(payload)
--- a/cantools/database/utils.py
+++ b/cantools/database/utils.py
@@ -250,12 +250,12 @@ def create_encode_decode_formats(datas: Sequence[Union["Data", "Signal"]], numbe
     little_fmt, little_padding_mask, little_names = create_little()
 
     try:
-        big_compiled = bitstruct.compile(big_fmt, big_names)
+        big_compiled = bitstruct.c.compile(big_fmt, big_names)
     except Exception as e:
         big_compiled = bitstruct.compile(big_fmt, big_names)
 
     try:
-        little_compiled = bitstruct.compile(little_fmt, little_names)
+        little_compiled = bitstruct.c.compile(little_fmt, little_names)
     except Exception as e:
         little_compiled = bitstruct.compile(little_fmt, little_names)
 
--- a/tests/test_autosar.py
+++ b/tests/test_autosar.py
@@ -26,7 +26,7 @@ class CanToolsAutosarTest(unittest.TestCase):
                          173)
         self.assertEqual(cantools.autosar.compute_profile2_crc(b'\x00\x51\x22',
                                                                msg),
-                         173)
+                         17)
 
 
         # test the apply function
@@ -34,14 +34,14 @@ class CanToolsAutosarTest(unittest.TestCase):
                                                               msg))
         self.assertEqual(cantools.autosar.apply_profile2_crc(b'\xff\x51\x22',
                                                              msg),
-                         b'\xad\x51\x22')
+                         b'\x11\x51\x22')
 
         # test the check function
         self.assertFalse(cantools.autosar.check_profile2_crc(b'\xff',
                                                              msg))
         self.assertFalse(cantools.autosar.check_profile2_crc(b'\x00\x51\x22',
                                                              msg))
-        self.assertTrue(cantools.autosar.check_profile2_crc(b'\xad\x51\x22',
+        self.assertTrue(cantools.autosar.check_profile2_crc(b'\x11\x51\x22',
                                                             msg))
 
         # make sure that profile 5 protected messages cannot be used


cantools/autosar/end_to_end.py
```diff
@@ -37,7 +37,7 @@ def compute_profile2_crc(payload : Union[bytes, bytearray],
         assert msg.autosar.e2e.data_ids is not None
 
         protected_len = msg.autosar.e2e.payload_length
-        seq_counter = (payload[1]&0xf0) >> 4
+        seq_counter = payload[1] & 0xf
```
andlaus, line 40: are you sure that the other nibble is used  for the sequence counter? (I thought I tested this, but maybe I made a stupid mistake...)

cantools/database/utils.py
```diff
@@ -250,12 +250,12 @@ def create_little() -> Tuple[str, int, List[str]]:
     little_fmt, little_padding_mask, little_names = create_little()
 
     try:
-        big_compiled = bitstruct.compile(big_fmt, big_names)
+        big_compiled = bitstruct.c.compile(big_fmt, big_names)
```
andlaus, line 253: it must be assumed that `bitstruct.c` might not be available, so this should either be surrounded by a `try..except` statement, or something like 

```python
try:
 import bitstruct.c as bitstruct
except:
 import bitstruct
```

should be done. (the latter must be used with caution as the `bitstruct.c` module misses quite a few features compared to `bitstruct`.)

cantools/autosar/end_to_end.py
zariiii9003: Yep, look [here](https://www.autosar.org/fileadmin/user_upload/standards/foundation/21-11/AUTOSAR_PRS_E2EProtocol.pdf) under `6.4.1 Header Layout`

cantools/database/utils.py
zariiii9003: Both the import as well as the function call are inside a try...except statement. It's not pretty, but it works.

If we use `import bitstruct.c as bitstruct` then we can't use the python fallback in the function call, right?

cantools/autosar/end_to_end.py
zariiii9003: The high nibble is data:
image

andlaus: right. I suppose I only tested the code with data for profile 5 (which uses 16 bit sequence counters). for profile 2 I guess I became confused by the fact that `cantools dump` prints the highest significant bit on the left...

cantools/database/utils.py
andlaus: yes, at least not easily (and it would not really make sense because these import shenangians would only be done in order to avoid adding all these try-catch blocks). That said, if there's not enormous performance drawback because of this, I'd argue for just continuing to use plain-ol-`bitstruct` unconditionally!?

zariiii9003: I'll test it later, but i assume the performance impact will be huge.

andlaus: cool. could you pleas not test the `compute_profile2_crc` function in isolation (my uneducated guess there is that the performance impact for this will be at least 20%), but encoding a message followed by calculating and applying the E2E bits... (since there's not much point in these functions without encoding a message first, right?)

zariiii9003: Running  `print(min(timeit.Timer('message.encode(unpacked, strict=False, scaling=False)', setup=setup).repeat(10, 1000)))` results in 0.008279s with the C extension and 0.02554s without the C extension. -> 208% increase for simple encoding.

Running  `print(min(timeit.Timer('message.decode(packed, decode_choices=False, scaling=False)', setup=setup).repeat(10, 1000)))` results in 0.004291s with the C extension and 0.019378s without the C extension. -> 351% increase for simple decoding.

andlaus: thanks for the result. That's not exactly what I meant, though: I was wondering how much faster the snippet `raw_msg = message.encode(unpacked); apply_profile2_crc(raw_msg, 123)`  would be in the case that `bitstruct.c` was used for `apply_profile2_crc()` versus it not being not used, but with `bitstruct.c` being used for `message.encode()` in both cases. (I think that this is the real question here...)

zariiii9003: I tested with a message length of 48 bytes.

```python3
print(min(timeit.Timer('data = message.encode(unpacked, strict=False, scaling=False)', setup=setup).repeat(10, 1000)))
print(min(timeit.Timer('data = message.encode(unpacked, strict=False, scaling=False);apply_profile2_crc(data, 123)', setup=setup).repeat(10, 1000)))
print(min(timeit.Timer('data = bytearray(message.encode(unpacked, strict=False, scaling=False));e2e.p02.e2e_p02_protect(data=data, length=47,data_id_list=data_id_list, increment_counter=False)', setup=setup).repeat(10, 1000)))
```

This is the result with bitstruct.c:
```
0.008037800000238349
0.034645600000658305
0.00881640000079642
```

This is the result with just bitstruct:
```
0.02558330000101705
0.05244979999952193
0.026652200000171433
```

The third value is with my own [lib](https://github.com/zariiii9003/autosar-e2e), but it only supports p02 and the CRC functions as of now. We get the same CRC though, so that's nice.

Even when you encode and apply E2E the difference between bitstruct.c and bitstruct is significant. If you just encode with no E2E, the difference is huge.

So: let's keep bitstruct.c. I don't see why we should remove it.

andlaus: Ok, I guess I got confused by the fact that this PR conflates two unrelated issues into the same commit: the fix for the E2E packing bug and the preferred usage of `bitstruct.c` for any messages, not just E2E protected ones. (right?)

I agree that we should try to use `bitstruct.c` for encode/decode because it is a performance critical path, but the current patch unfortunately does not yet cut it fully: the [Formats](https://github.com/cantools/cantools/blob/master/cantools/typechecking.py#L24) class currently only allows `CompiledFormatDict` objects from 'bitstruct', not the ones from `bitstruct.c` (this should be rather easy to fix, though...)

zariiii9003: Yes, the two issues are not really related. 

`bitstruct.c.CompiledFormatDict` and `bitstruct.CompiledFormatDict` are pretty much the same from a typing point of view. mypy ignores bitstruct anyway since does not contain a `py.typed` file.

andlaus: right. I think this PR should be merged once the test failure stemming from the change of the sequence counter nibble is fixed. Ideally, the issue with the Formats class should be fixed as well because the bitstruct module might include `py.typed` in the future...

zariiii9003: I fixed the tests, but i'm not sure how to handle `bitstruct.c.CompiledFormatDict` in type annotations since it might not be available on all platforms. I assume, that `bitstruct.CompiledFormatDict` and `bitstruct.c.CompiledFormatDict` have the same methods and attributes, so imo it is acceptable to ignore the c implementation in type hints.

andlaus: fair enough. @juleq: merge?

PR #493, opened 2022-10-15
Make dbc format more closely match Vector CANdb++ 3.1.18

Wetmelon:
- Indicate a value is signed if it's a float
- Reverse sort signals (`SIG_VALTYPE_`) by start bit

Newest Vector CANdb++ tool indicates a signal is signed (`-`) if it's a floating point value and has a different sorting for signals in .dbc files.

### DIFF
--- a/cantools/database/can/formats/dbc.py
+++ b/cantools/database/can/formats/dbc.py
@@ -487,7 +487,7 @@ def _dump_messages(database, sort_signals):
                 length=signal.length,
                 receivers=format_receivers(signal),
                 byte_order=(0 if signal.byte_order == 'big_endian' else 1),
-                sign=('-' if signal.is_signed else '+'),
+                sign=('-' if signal.is_signed or signal.is_float else '+'),
                 scale=signal.scale,
                 offset=signal.offset,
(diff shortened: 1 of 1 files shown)


andlaus:
besides the the fact that the unit tests need to be fixed and that I'm a bit skittish about changing the order of the signals here (this should possibly be done by specifying the `sort_signals` parameter of the database's [constructor](https://github.com/cantools/cantools/blob/master/cantools/database/can/database.py#L62), this LGTM. @juleq calls the shots with the DBC code, though...

juleq:
I was thinking that it might not be desirable to let cantools make changes to an input file (of whatever origin) for no other reason as to make it appear to have been produced by CANdb++.

The issue with that are:
- User expectation: User loads a valid dbc with strange signal ordering, makes a minor change using cantools and dumps again. He will probably be surprised about the massive diff his minor change has produced.
- Testing: The easiest testing strategy is to load/dump and check for equality. That does not work if cantools decides to reorder instead of retain.

So I would say: Ordering cantools internal data without API support is straightforward. Providing optional API to do so is good. Changing the default behavior is probably not so good.

PR #494, opened 2022-10-22
Use tox

zariiii9003:


andlaus (commented):
seems fine to me. That said, I don't have any experience with tox. any opinions, @juleq?

PR #495, opened 2022-10-24
Typing improvements

erzoe:
- Use `Mapping` instead of `Dict` where possible, see the commit message of commit 7f853c660768d6142a4c027acdeb567e827d42d0
- Return type of `Database.decode_message` is not considered to be `Any` with `mypy --strict`, see the commit message of commit 72e1642b590513001f235e20b5378d4fd29b806d


zariiii9003:
Why should the return value of `decode` be immutable?

erzoe:
@zariiii9003 good point. I have changed the return type back to Dict. Thank you for checking my code.

andlaus (approved):
looks good. any vetos, @juleq ?

PR #508, opened 2022-12-06
Ensure that only Decimal types are stored in the Signal.decimal attribute

jonmon6691:
Corrects a bug where floats can leak into the Signal.decimal attribute and cause operator errors elsewhere when minimums and maximums are operated on in conjunction with offsets and factors


andlaus (approved):
looks good. The unit tests could possibly adapted to ensure this more uniformly, but IMO this is not strictly required...

andlaus:
@jonmon6691: can you rebase this on top of the latest master? this should fix the CI (for some reason, just re-running the workflows does not do the trick...)

jonmon6691:
@andlaus Done.

PR #509, opened 2022-12-08
Fix CI

andlaus:
this PR tries to fix the github actions which recently started failing:

- The first issue is that on windows, cantools depends on the `windows-curses` package which was an implicit dependency of previous versions `python-can`. (Note that I'm not sure if the fix is correct: we check if we are running on windows in `setup.py` and if we are, add `windows-curses` to the requirements. This approach will fail if dependencies get "hard baked" into PyPi packages and one tries to install a package prepared on a non-Windows system on Windows or vice-versa.)
- the second issue is that python-3.6 does not seem to be available on the latest ubuntu plattform anymore. This is fixed by disabling the python 3.6 tests on ubuntu-latest (to compensate, we now also test on ubuntu 20.04)

I guess it is time to think about dropping support for python 3.6. do you have a stance here @juleq?

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -41,22 +41,34 @@ jobs:
 
   test:
     runs-on: ${{ matrix.os }}
-    if: ! ${{ matrix.skip_job }}
     strategy:
       matrix:
-        os: [ubuntu-latest, ubuntu-20.04, macos-latest, windows-latest]
-        # we cannot simply define a matrix with python versions 3.6
-        # and 3.10 directly because python 3.6 is not available on
-        # ubuntu-latest anymore. The workaround is to make the job
-        # conditional on the "skip-job" variable specified via matrix
-        # includes and only set it to true for illegal combinations
-        python-version: ["3.6", "3.10"]
         include:
-          - skip-job: false
 
-          - os: ubuntu-latest
-            python-version: 3.6
-            skip-job: true
+          # we cannot simply define a matrix for [ubuntu-latest,
+          # macos-latest, windows-latest] times [python-3.6 ,
+          # python-3.10] directly because python 3.6 is not available on
+          # ubuntu-latest anymore. The workaround is to make a long 1D
+          # matrix and set the matrix variables manually via
+          # includes. This is pretty clunky, but job-conditionals are
+          # not allowed to depend on matrix variables and other
+          # solutions like preparing the matrix using a separate job
+          # seem to be even more cumbersome...
+          - os: "ubuntu-20.04"
+            python-version: "3.6"
+          - os: "ubuntu-latest"
+            python-version: "3.8"
+          - os: "ubuntu-latest"
+            python-version: "3.10"
+          - os: "windows-latest"
+            python-version: "3.6"
+          - os: windows-latest
+            python-version: "3.10"
+          - os: "macos-latest"
+            python-version: "3.6"
+          - os: "macos-latest"
+            python-version: "3.10"
+
       fail-fast: false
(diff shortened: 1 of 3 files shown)


juleq:
Yes, old stuff must go, we need to keep maintenance to a minimum, IMHO.

Thanks for fixing, should I merge already?

andlaus:
> Thanks for fixing, should I merge already?

let's wait a bit for what @zariiii9003 has to say. Maybe he knows a better way got skip a configuration of a test matrix and/or an "official" way of  specify system dependent requirements?

andlaus:
Also, I added a few screwups to the workflow file. (sorry. I had a hard time testing this before opening the PR.)

Stay tuned...

andlaus:
alright, now it seems to work (at the expense of serious clunkiness in the workflow file).

Side note: During this excercise I learned to love to hate YAML...

.github/workflows/pythonpackage.yml
```diff
@@ -43,8 +43,32 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.6", "3.10"]
+        include:
```
zariiii9003, line 46: I would suggest dropping 3.6 support instead. It reached end of life status one year ago.
Keep using the matrix, drop 3.6, add 3.11.

setup.py
```diff
@@ -11,6 +12,24 @@ def find_version():
...
+        'bitstruct>=8.15.1',
+        'python-can>=3.3.4',
+        'textparser>=0.21.1',
+        'diskcache',
+        'argparse_addons',
+        'typing_extensions>=3.10.0.0',
+        'crccheck',
+    ]
+
+    if sys.platform == 'win32':
```
zariiii9003, line 26: There is syntax for making dependencies platform dependent. Python-can used `'windows-curses;platform_system=="Windows" and platform_python_implementation=="CPython"'`

However, i would avoid making the same mistake and instead add windows-curses as an optional dependency.
You would need to catch the ImportError on Windows and add a meaningful error message though.

zariiii9003 (changes requested):

.github/workflows/pythonpackage.yml
andlaus: ok, done. unfortunately, python 3.11 does not seem to work with our test setup, and the error message is rather cryptic:

https://github.com/cantools/cantools/actions/runs/3674907063/jobs/6213690296

maybe you have an idea? (I stayed on 3.10 in the meantime.)

andlaus:
alright, by popular demand I hereby degree to raise the minimum python version to 3.8. (IMO we should not raise it beyond that as long as ubuntu 20.04 support still matters to us.)

juleq:
Great. Just say the word :).

andlaus:
I guess this should be merged now and if we find a solution for the python-3.11 issue, do this in another PR...

zariiii9003:
@juleq @andlaus  You just released 37.3, but the minimum supported python version was raised to 3.8. So it should have been a major version bump and mentioned in the changelog.

andlaus:
good point. is it possible to retract a version on pypi? if no, the exercise would be a bit pointless IMO...

juleq:
Whoops. Trigger happy me. Thank you for catching. Yanked, deleted, bumped and released.

juleq:
@andlaus Yanked it from pypi few minutes after @zariiii9003 commented.

juleq:
Pypi taking its time to release, maybe some rate limiting.

Issue #487, opened 2022-09-28
[suggestion] update API doc in readthedocs


ziyangfu:
The last successful build of doc was 2 years ago, and add many new features at the period of 2 years. maybe rebuild the doc manually is a good idea. I do not know why automatic build failed recently.

zariiii9003:
A [new version of the docs](https://cantools.readthedocs.io/en/latest/) has been built with version 38, but the images added in #252 and #277 fail to load because of the relative paths.

Maybe @erzoe could fix that?

erzoe:
@zariiii9003 I have opened a pull request: https://github.com/cantools/cantools/pull/512

PR #511, opened 2022-12-31
Test on python 3.11, build docs again

zariiii9003:
Python 3.11 just needed a version update of coverage.py and pytest-cov. There is a failing test though.
I also added a .readthedocs.yml file to address #487

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -25,7 +25,7 @@ import cantools
 # -- General configuration ------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
-# needs_sphinx = '1.0'
+# needs_sphinx = "1.0"
 
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
@@ -45,11 +45,11 @@ templates_path = ["_templates"]
 
 # The suffix(es) of source filenames.
 # You can specify multiple suffix as a list of string:
-# source_suffix = ['.rst', '.md']
+# source_suffix = [".rst", ".md"]
 source_suffix = ".rst"
 
 # The encoding of source files.
-# source_encoding = 'utf-8-sig'
+# source_encoding = "utf-8-sig"
 
 # The master toctree document.
 master_doc = "index"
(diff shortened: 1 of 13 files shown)


docs/conf.py
```diff
@@ -34,8 +33,13 @@
 extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.viewcode',
+    'sphinx_rtd_theme',
+    "sphinx.ext.intersphinx",
```
andlaus, line 37: for consistency, please use single quotes if the string in question does not contain single quotes.

(further instances of this heinous atrocity below..)

andlaus (approved):
looks good.

docs/conf.py
zariiii9003: i used black for consistency 😄

docs/conf.py
```diff
@@ -113,156 +117,149 @@
...
+# html_search_scorer = 'scorer.js'
 
 # Output file base name for HTML help builder.
-htmlhelp_basename = 'cantoolsdoc'
+htmlhelp_basename = "cantoolsdoc"
 
 # -- Options for LaTeX output ---------------------------------------------
 
 latex_elements = {
-# The paper size ('letterpaper' or 'a4paper').
```
andlaus, line 213: hm, black seems to have slighly messed up here...

tox.ini
```diff
@@ -4,16 +4,16 @@ isolated_build = true
 
 [testenv]
 deps =
-    pytest==7.0.*
-    pytest-cov==3.0.0
-    coverage==6.2
+    pytest==7.1.*,>=7.1.2
```
andlaus, line 7: isn't that simply `>= 7.1.2`?

andlaus: ok, no: it prevents 7.2 and above to be used. I wonder if it isn't better to allow new major releases of the dependencies: the test environment will break occasionally, but so will it now once the specified versions become unsupported...

zariiii9003: I copied `pytest==7.1.*,>=7.1.2` from python-can. Back then this was the only combination of pytest and coverage, that worked for python 3.11.
 I updated these in the last commit, but i would keep them pinned to a specific version. It's annoying when the CI fails in PRs, just because there is a new pytest version.

andlaus: ok. I don't mind either.

andlaus:
@juleq: IMO this can be merged...

PR #512, opened 2023-01-05
Changed relative links to absolute links

erzoe:
trying to fix [this issue](https://github.com/cantools/cantools/issues/487#issuecomment-137139501)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,7 +154,7 @@ Optionally a format can be specified after a signal, separated by a colon.
 
     $ cat candump-2021-01-04_180521.log | python3 -m cantools plot tests/files/dbc/abs.dbc '*33.*fl:-<' '*33.*fr:->' - '*33.*rl:-<' '*33.*rr:->'
 
-.. image:: docs/plot-2-subplots.png
+.. image:: https://github.com/cantools/cantools/raw/master/docs/plot-2-subplots.png
 
 Signals with a different range of values can be displayed in the same subplot on different vertical axes by separating them with a comma.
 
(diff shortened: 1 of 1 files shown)


juleq:
Thanks. Do you have an easy way to try or should I just do a patch release?

zariiii9003:
There's another one on line 146

erzoe:
> There's another one on line 146

How embarrassing. Yes you are right, thank you. I have added another commit.

erzoe:
> Thanks. Do you have an easy way to try or should I just do a patch release?

Well, I had verified that the images are visible in the README on my fork. But that is not a good test, because they were visible there before the changes, too.

Now I have checked with grep that I have not missed any links again and it looks good:

```bash
$ grep docs/ README.rst 
.. image:: https://github.com/cantools/cantools/raw/master/docs/plot-1.png
.. image:: https://github.com/cantools/cantools/raw/master/docs/plot-2-subplots.png
.. image:: https://github.com/cantools/cantools/raw/master/docs/plot-2-axes.png
.. image:: https://github.com/cantools/cantools/raw/master/docs/plot-seaborn.png
.. image:: https://github.com/cantools/cantools/raw/master/docs/monitor.png
```

erzoe:
By the way, what would be the right way to deal with links in merge requests? Because if I used absolute links like this I would not be able to test them before the merge. Using absolute links to my fork would not seem right because if I would update an image on my fork during development they would show unmerged development in the official documentation. But using relative links has proven to cause problems, too.

andlaus:
> By the way, what would be the right way to deal with links in merge requests?

IMO, just use the URIs where you expect them to go. If it takes a few tries to get it right, so be it... (I'm quite surprised/shocked that relative links don't work here, though.)

andlaus:
@juleq: I suppose it is worth to give this a shot and merge...

PR #513, opened 2023-01-11
Add CLI to disable snake case conversion and use original casing found in the dbc

GraysonNocera:


juleq:
Hi. 

Maybe having a getter that considers the option would be a more flexible approach and it would be in kist one place. This is just a suggestion to consider, I am not a user of the generator -> do not assume that I have fully thought that through :).

PR #515, opened 2023-01-19
Allow Decimal value to be encoded in a signal field

deronek:
### Issue description
PR #317 introduced storing initial value of a signal in a `Tester` object as a Decimal. 
When trying to use `Tester` class with DBC including initial values for a signal (`GenSigStartValue`), an exception is thrown, because function `_encode_fields` doesn't consider the Decimal object as neither float or int and tries to use it as a choice.

Unrelated to `Tester` class, this should also manifest every time user wants to encode a Decimal object as a value of a signal.

### Example
Attached very simple example DBC generated using Vector CANdb++. Signal `Random_Value` contains a non-zero `GenSigStartValue` attribute value.
[VEHICLE_CAN.zip](https://github.com/cantools/cantools/files/10461102/VEHICLE_CAN.zip)

When trying to create a Tester object using that DBC (full traceback: [traceback.txt](https://github.com/cantools/cantools/files/10461120/traceback.txt)):
```
  File "...\venv\Lib\site-packages\cantools\database\utils.py", line 98, in encode_data
    unpacked = _encode_fields(fields, data, scaling)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\venv\Lib\site-packages\cantools\database\utils.py", line 85, in _encode_fields
    unpacked[field.name] = field.choice_string_to_number(str(value))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\venv\Lib\site-packages\cantools\database\can\signal.py", line 334, in choice_string_to_number
    raise ValueError(f"Signal {self.name} has no choices.")
ValueError: Signal Random_Value has no choices.`
```

### Implemented change
My change implements detecting the Decimal object and using `field.decimal` to calculate the encoded value.

Let me know if you suggest something else regarding this issue.


andlaus:
looks okay assuming that computations involving decimal objects do not introduce rounding errors. I believe that was the reason for introducing them, but I'm not sure. @juleq weren't you involved in this?

deronek:
Added fix after mypy static analysis.

cantools/database/utils.py
```diff
@@ -79,8 +79,8 @@ def _encode_fields(fields: Sequence[Union["Signal", "Data"]],
             unpacked[field.name] = _transform(value)  # type: ignore[operator]
             continue
 
-        if isinstance(value, Decimal):
-            unpacked[field.name] = (value - field.decimal.offset) / field.decimal.scale
+        if isinstance(value, Decimal) and isinstance(field, Signal):
```
deronek, line 82: I am not very happy with the second check, but I guess we need it since class `Data` does not have the `decimal` attribute. I don't see easier way to implement this

juleq:
@andlaus Yes, I think we took part in that discussion. I think the background was to have load/dump cycles without diffs due to non-representability of a given value in 754 float IIRC.

cantools/database/utils.py
```diff
@@ -78,6 +79,10 @@ def _encode_fields(fields: Sequence[Union["Signal", "Data"]],
             unpacked[field.name] = _transform(value)  # type: ignore[operator]
             continue
 
+        if isinstance(value, Decimal) and isinstance(field, Signal):
+            unpacked[field.name] = (value - field.decimal.offset) / field.decimal.scale  # type: ignore[operator]
```
zariiii9003, line 83: You shouldn't ignore mypy here. `offset` and `scale` can be `None`

deronek: Hi, thanks for the comment. 

If the user wants to encode a Decimal into a signal, the code **should** break when the decimal offset and scale are not available. If they are available, my change implements the correct calculation.

Alternatively, we can fall back to using float offset and scale, but this defeats the purpose of the Decimal - making sure that the physical representation of the signal is not affected by rounding errors.

deronek:
Hi @andlaus, I recommend merging the PR as it fixes a functionality which is a blocker in certain use cases.

cantools/database/utils.py
andlaus: > If the user wants to encode a Decimal into a signal, the code should break when the decimal offset and scale are not available. If they are available, my change implements the correct calculation.

I basically agree with both of you: I think that this case should get checked and if it applies an exception should be raised explicitly...

andlaus:
@deronek: okay. a conflict with master cropped up since this PR was opened. Once that's fixed I'll merge if @zariiii9003 does not object either...

cantools/database/utils.py
deronek: Sure, I will implement explicit check for this case (and make sure there are no conflicts with master) next week.

zariiii9003: Can bitstruct encode `decimal.Decimal` instances? Do we need to cast to int or float here?
We should add a test for this.

albsod:
Unfortunately, this PR does not seem to work when using the Tester class. I tried running the following commands with deronek's DBC file in the current directory:

```
import can
import cantools
bus = can.Bus(interface="socketcan", channel="vcan0",
              bitrate=250000)
database = cantools.database.load_file("VEHICLE_CAN.dbc")
tester = cantools.tester.Tester(dut_name="VehicleManagementModule",
                                     database=database,
                                     can_bus=bus,
                                     bus_name="VEHICLE_CAN")
```

It gives the following error:

```
little_packed = formats.little_endian.pack(unpacked)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'decimal.Decimal' object cannot be interpreted as an integer
```

This does not occur in the PR #538 recently submitted by my colleague.

albsod:
That's curious, @deronek. I'm running on GNU/Linux and this is what I get:

```
python3
Python 3.11.2 (main, Feb 12 2023, 00:48:52) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import can
>>> import cantools
>>> bus = can.Bus(interface="socketcan", channel="vcan0",
...               bitrate=250000)
>>> database = cantools.database.load_file("VEHICLE_CAN.dbc")
>>> tester = cantools.tester.Tester(dut_name="VehicleManagementModule",
...                                      database=database,
...                                      can_bus=bus,
...                                      bus_name="VEHICLE_CAN")
  File "/home/albin/.python-environments/.venv/lib/python3.11/site-packages/cantools/tester.py", line 351, in __init__
    self._messages[message.name] = Message(message,
                                   ^^^^^^^^^^^^^^^^
  File "/home/albin/.python-environments/.venv/lib/python3.11/site-packages/cantools/tester.py", line 137, in __init__
    self.update(self._prepare_initial_signal_values())
  File "/home/albin/.python-environments/.venv/lib/python3.11/site-packages/cantools/tester.py", line 159, in update
    self._update_can_message()
  File "/home/albin/.python-environments/.venv/lib/python3.11/site-packages/cantools/tester.py", line 250, in _update_can_message
    data = self.database.encode(pruned_data,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/albin/.python-environments/.venv/lib/python3.11/site-packages/cantools/database/can/message.py", line 943, in encode
    encoded, padding_mask, all_signals = self._encode(self._codecs,
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/albin/.python-environments/.venv/lib/python3.11/site-packages/cantools/database/can/message.py", line 780, in _encode
    encoded = encode_data(data,
              ^^^^^^^^^^^^^^^^^
  File "/home/albin/.python-environments/.venv/lib/python3.11/site-packages/cantools/database/utils.py", line 105, in encode_data
    little_packed = formats.little_endian.pack(unpacked)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'decimal.Decimal' object cannot be interpreted as an integer
```

albsod:
> Hi @albsod, I'm unable to reproduce the issue using your code, it works well. #538 looks good with additional changes, but it's missing one check which I implement here (I'll put it as a comment there).

Now that your additional check has been added to PR #538, should we go with that one instead considering that it seems to be working on both Linux and Windows?

deronek:
Issue could not be reproduced after #536, because Tester does not use Decimal to calculate the initial values anymore.

PR #516, opened 2023-01-23
Relaxed an PDU assert for ARXML format.

PKrza:
Instead of breaking the parser if PDU is none, just return an empty message. The same result occurs if a PDU does not contain any signals.

### DIFF
--- a/cantools/database/can/formats/arxml/system_loader.py
+++ b/cantools/database/can/formats/arxml/system_loader.py
@@ -758,7 +758,24 @@ class SystemLoader(object):
         # things like multiplexed and container messages, this is not
         # the case...
         pdu = self._get_pdu(can_frame)
-        assert pdu is not None
+        if pdu is None:
+            return Message(bus_name=bus_name,
+                           frame_id=frame_id,
+                           is_extended_frame=is_extended_frame,
+                           is_fd=is_fd,
+                           name=name,
+                           length=length,
+                           senders=[],
+                           send_type=None,
+                           cycle_time=None,
+                           signals=[],
+                           contained_messages=None,
+                           unused_bit_pattern=0xff,
+                           comment=None,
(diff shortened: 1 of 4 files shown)


andlaus (commented):
the change itself looks good, but I would like to have a test case. (just curious: what is a frame without a PDU useful for?) You can modify e.g. [`system-4.2.arxml`](https://github.com/cantools/cantools/blob/master/tests/files/arxml/system-4.2.arxml) to feature a frame without a PDU and modify [`test_database.py`](https://github.com/cantools/cantools/blob/master/tests/test_database.py) accordingly.

PKrza:
I have some arxml files that define messages without PDUs and signals, and all these messages are Diagnostic messages with user defined PDUs. So in a way, these messages are just a placeholders and to reserve IDs on a CAN bus.

andlaus:
> I have some arxml files that define messages without PDUs and signals, and all these messages are Diagnostic messages with user defined PDUs. So in a way, these messages are just a placeholders and to reserve IDs on a CAN bus.

hm, okay. These kinds of messages are usually given PDUs of type `GENERAL-PURPOSE-PDU` , `USER-DEFINED-I-PDU` or `N-PDU`. Given that this is AUTOSAR, I suppose that there is more than one way to do it...

@juleq: Once the tests pass, I think this PR is ready to be merged...

andlaus:
> [...] Once the tests pass [...]

@PKrza: the test that fails is `test_list.py`. fixing it should come down to copy-and-pasting the results of `cantools list [OPTIONS] system-4.2.arxml` to the test...

andlaus:
all tests pass now; nice! @juleq: I think this is ready to be merged

andlaus:
thanks for merging. I think that this change in conjunction with #517 justify doing a point release?

juleq:
Sure thing. Done.

PR #521, opened 2023-02-15
Decode Formula updated

AnandSingh-0619:
If both attributes are present, the data are scaled before the offset is added. When scaled data are written, the application should first subtract the offset and then divide by the scale factor.
For example, if x is to be sent and the field has an offset of 400 and scaling factor of 10, the final data becomes:
y = [(Scaling factor) * (x)] + offset

Now when this message is decoded, the scaling factor being 0.1 and offset being -400,
x = (y + offset) / scaling factor

Below is the link for the standard convention:
https://www.eyascopublic.com/mehelp/index.html#!Documents/scalefactorandoffset.htm

https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch08.html#:~:text=8.1.,-Packed%20Data&text=If%20both%20attributes%20are%20present,divide%20by%20the%20scale%20factor.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "ros.distro": "melodic"
+}
--- a/cantools/database/utils.py
+++ b/cantools/database/utils.py
@@ -145,7 +145,7 @@ def decode_data(data: bytes,
                     pass
 
             if scaling:
-                decoded[field.name] = field.scale * value + field.offset
+                decoded[field.name] = (value + field.offset) * field.scale 
                 continue
             else:
                 decoded[field.name] = value


andlaus:
I'm pretty sure that your patch is wrong: given an internal value `x` (i.e., the value which is send over the wire), a scaling factor `s` and an offset `o`, the physical value `y` (i.e., the value which corresponds to `x` in the physical world) is defined as
```
y = x*s + o
```

(if this was not the case, none of my cantools-based scripts would work.)

AnandSingh-0619:
The cases for which this error would occur is when you have both offset and scaling factor in a field, hence error might not be common.
You can check the link I shared.

AnandSingh-0619:
You can add test cases where you encode and decode data using the condition I mentioned.
Scaling factor: 10
Offset: 400
After sending the data on can try to parse the data with your earlier formula and the corrected formula.
Scaling factor : 0.1
Offset : -400

andlaus:
> This would be the case when you send the value. On the receiving end, when you parse the value you are calculating x using the y received, hence the formula changes. I can across this error when I was trying to parse my trace file using a standard .dbc file.

this has nothing to do with the sender or the receiver -- it is for the conversion from internal to physical values. the reverse is obviously achieved by resolving above equation to `x`:
```
x = (y - o)/s
```

I'm still 99.95% sure that the current code is correct...

AnandSingh-0619:
Suppose one node is sending data over CAN Bus with the above field offset and scaling factor.
My x=195
Now this is encoded as:
y = [(Scaling factor) * (x)] + offset
y = (10*195)+400= 2350
This decimal value is converted to hex and transmitted on CAN Bus.
Now for the receiving node to decode the recived message, it'll convert from hex to dec
Then
x = (y + offset) * scaling factor
x = (2350 + (-400)) *0.1
x= 195

If you go by the old formula you used, then
x = (y  * scaling factor) + offset
x = (2350 * 0.1) - 400
x = -165 

In the case of decoding messages using DBC files, the formula should be updated to get the correct readings.

andlaus:
> My x=195

you probably want to your 'y' to be '195'. let's say that this represents the vehicle's velocity in kilometers per hour, to ease the argument. (N.B.: you're probably driving on a German Autobahn ;))

> This decimal value is converted to hex and transmitted on CAN Bus.

it is send as a binary number according to the specification of the communication description file (e.g. DBC).

let me try to clear the confusion: let's assume that above vehicle velocity signal is transmitted using 10 bits with a resolution of 0.25 km/h per increment (i.e., `s=0.25`) and a maximum reverse speed of `30 km/h` (i.e., `o=-30`). The sending ECU now measures the road speed of the vehicle (`y`) and needs to convert it to the value `x` before sending it over the wire. this is done using `x = (y - o)/s = (195 - -30)/0.25 = 900`. the sender receives this value embedded in some frame, and inferes the vehicle's velocity using the forward equation `y = x*s + o = 900*0.25 + -30 = 195`. You can argue that there might be a better way to do this, but this is definitively the way it is done in any system which is supported by `cantools`...

andlaus:
I suppose this issue is settled, let's close the PR. If you have any further questions, feel free to add more comments...

PR #525, opened 2023-02-17
do not intercept raised exceptions in debug mode

andlaus:
also, lower the minium log level for output to "logging.DEBUG" in this case.

The rationale for these changes is to make development of functionality easier by keeping the tracebacks of raised exceptions meaningful and seeing the debugging messages of the file format loaders...

### DIFF
--- a/cantools/database/__init__.py
+++ b/cantools/database/__init__.py
@@ -1,19 +1,17 @@
 import os
-from typing import Union, Optional, TextIO, MutableMapping, cast
+from typing import MutableMapping, Optional, TextIO, Union, cast
 from xml.etree import ElementTree
 
-from .errors import ParseError
-from .errors import Error
-from ..compat import fopen
-from . import can
-from . import diagnostics
-from . import utils
-import textparser
 import diskcache
+import textparser
+
+from ..compat import fopen
+from ..typechecking import StringPathLike
+from . import can, diagnostics, utils
 
 # Remove once less users are using the old package structure.
 from .can import *
-from ..typechecking import StringPathLike
+from .errors import Error, ParseError
 
 
 class UnsupportedDatabaseFormatError(Error):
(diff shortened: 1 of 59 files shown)


andlaus:
@juleq: I don't see any reason why this should not be merged, do you?

juleq:
So it optionally increases verbosity? All for it. Just say it :).

andlaus:
> So it optionally increases verbosity?

yes. Also the tracebacks of exceptions thrown by file parsers do not get trashed anymore, i.e. it should be much easier to find the location where such an exception was raised in debug mode.

on the flipside, auto-detection the format of strings will not work properly in debug mode anymore (fortunately, loading from files is unaffected). IMO that is a very reasonable price to pay...

cantools/database/__init__.py
```diff
@@ -366,24 +372,32 @@ def load_can_database(fmt: str) -> can.Database:
             return load_can_database('arxml')
         except (ElementTree.ParseError, ValueError) as e:
             e_arxml = e
+            if debug:
```
zariiii9003, line 375: Maybe we could just raise UnsupportedDatabaseFormatError here with the debug flag?

andlaus: good suggestion (didn't know this trick yet). Thanks for the hint!

I've implemented it this way and IMO the result is quite a bit cleaner and shorter now...

andlaus:
> So it optionally increases verbosity? All for it. Just say it :).

thanks to @zariiii9003's proposal it now evens works  with format auto-detection :). (Well, it works exactly the same way as it did before...)

zariiii9003 (approved):
Looks good, but there are unused imports now: ElementTree and textparser

andlaus:
>  there are unused imports now: ElementTree and textparser

Is this related due to this PR? (how to find this out?) anyway, `pyflakes` produces quite a few warnings:
```
pyflakes cantools
cantools/__init__.py:8:1: '.tester' imported but unused
cantools/__init__.py:9:1: '.j1939' imported but unused
cantools/__init__.py:10:1: '.logreader' imported but unused
cantools/__init__.py:11:1: '.errors.Error' imported but unused
cantools/__init__.py:14:1: '.database as db' imported but unused
[...]
```

I guess that most of these are incorrect because many of these sub-modules are imported for the benefit of downstream code. I'll have a look (in another PR)...

zariiii9003:
Yes, they were used in the except clauses. Maybe we should add a linter to the CI pipeline. [ruff](https://github.com/charliermarsh/ruff) seems like a good choice lately.

andlaus:
@zariiii9003: even with #528 merged, this does not produce any warnings in the linter stages. it should, right?

zariiii9003:
> @zariiii9003: even with #528 merged, this does not produce any warnings in the linter stages. it should, right?

I only activated the pyupgrade check in the first PR, so this is not checked yet. I will create more PRs for that.

zariiii9003:
I configured ruff to ignore unused imports in __init__ files, but now it doesn't complain in this case.
The suggested solution by ruff is `consider adding to `\_\_all__` or using a redundant alias`. Both not great 🤔

All that code shouldn't be inside the init module tbh

andlaus:
> All that code shouldn't be inside the init module tbh

agreed, but that refactoring will need to be the subject of another PR. (actually, I would like to flatten the internal directory structure of cantools significantly and (maybe) remove some cruft while at it. let's see when i come to it...)

For this PR, I've manually removed the unused the unused imports that where not used by any test, so let's get this merged (with the implication of reverting it if this turns out to be a stupid move). thanks a lot for your input, @zariiii9003!

PR #528, opened 2023-02-26
Run ruff linter in CI

zariiii9003:
In this PR i only activated the pyupgrade checks. The other checks can be activated in follow-up PRs, otherwise it will be impossible to review.

So this PR mostly removes `object` as parent class and uses f-strings where possible.


andlaus (approved):
looks good as far as I can tell

zariiii9003:
@andlaus should i wait for a second review or just merge?

andlaus:
just go ahead and merge. I think this is quite low risk...

PR #529, opened 2023-03-01
Check for unused imports

zariiii9003:
I activated the isort (I) and unused-imports (F401) checks.

### DIFF
--- a/cantools/__init__.py
+++ b/cantools/__init__.py
@@ -1,18 +1,15 @@
-import sys
 import argparse
 import importlib
-import pathlib
 import os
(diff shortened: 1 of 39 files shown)


andlaus (approved):
looks good. the only issue which might crop up with this is that it becomes harder to reproduce CI failures locally, in particular for "drive-by" contributors. (Let's see if this becomes a problem...)

zariiii9003:
That shouldn't be a problem, if we update the contribution instructions in the readme https://github.com/cantools/cantools/blob/master/README.rst#contributing

PR #530, opened 2023-03-02
Activate pyflakes check

zariiii9003:


cantools/database/can/formats/arxml/system_loader.py
```diff
@@ -1796,16 +1794,16 @@ def _load_linear_scale(self, compu_scale, decimal):
         if lower_limit is not None and \
              upper_limit is not None and \
              lower_limit > upper_limit:
-            LOGGER.warning(f'An valid interval should be provided for '
```
andlaus, line 1799: can we disable `F541`? I'd like to use F-strings even for situations where nothing is formatted...

zariiii9003: That's weird but it should be possible. Are you sure? 😄

andlaus: >  Are you sure?

quite: IMO, having to use two different kinds of strings for things like

```python
print(f"there is a problem ({id=}) with whatever stupid thing I wanted to do "
      f"and this needs to be fixed before we can proceed because if not, "
      f"world #{world_id} is gonna end")
```

looks pretty weird...

(also, I like to use f-strings for stuff which could potentially include formatting even if they do not at the moment.)

I'm not hell-bent on that matter, though, so if you *really* want F541, let's keep it active...

zariiii9003: I don't have a strong opinion about that one. I'll make the change and force push.

andlaus:
going in. thanks!

PR #531, opened 2023-03-06
Make scale and offset attribute of Decimal mandatory

zariiii9003:
Related to the discussion in #515

### DIFF
--- a/cantools/database/can/formats/sym.py
+++ b/cantools/database/can/formats/sym.py
@@ -305,7 +305,10 @@ def _load_signal_type_and_length(type_, tokens, enums):
     enum = None
     minimum = None
     maximum = None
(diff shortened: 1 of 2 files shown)


andlaus (approved):
seems good. This probably means that numerical physical values are now always `float`?

zariiii9003:
> This probably means that numerical physical values are now always `float`?

How so? I hope i did not change any behavior.

andlaus:
> How so? I hope i did not change any behavior.

Not sure, but possibly the internal value could now always be transformed via the scaling equation
```
y = x*s + o
```

with `s` and `o` being `float`.

to be clear: I am all for this, let's get it merged.

andlaus:
#555 should also fix this? (by removing the Decimal objects)

PR #536, opened 2023-03-13
Add raw_initial, scaled_initial and raw_invalid to Signal

### LINKED ISSUE #534: Signal.initial type discrepancy
Currently for dbc-files the `initial`-attribute of a `Signal` instance contains the raw integer value.
But for arxml it contains the physical or even a `NamedSignalValue`.

@andlaus What do you think about setting the `initial` attribute to the value of the `initial_int` in this line?
https://github.com/cantools/cantools/blob/df21fe3db7144e7e72e9882a661c5d4ea8c1dce7/cantools/database/can/formats/arxml/system_loader.py#L1485

zariiii9003:
Closes #534 

@andlaus Could you take a look at the failing test? I cannot find the initial value in the arxml file 🤔


cantools/database/can/formats/arxml/system_loader.py
```diff
@@ -1236,14 +1235,14 @@ def _load_multiplexed_pdu(self, pdu, frame_name, next_selector_idx):
             # specified indepently of that of the message. how should
             # this be handled?
 
-        if selector_signal.initial in selector_signal.choices:
-            selector_signal.initial = \
-                selector_signal.choices[selector_signal.initial]
+        if selector_signal.raw_initial in selector_signal.choices:
+            selector_signal.raw_initial = \
```
andlaus, line 1239: once the choices are resolved, this becomes a physical (/scaled) value!? (i.e., it should be `scaled_initial`.) On that matter, I think the 'scaled_' prefix looks a bit weird for selector signals (but not a big deal IMO)...

cantools/database/can/formats/arxml/system_loader.py
```diff
@@ -1236,14 +1235,14 @@ def _load_multiplexed_pdu(self, pdu, frame_name, next_selector_idx):
...
+                selector_signal.choices[selector_signal.raw_initial]
 
-        if not isinstance(selector_signal.invalid, NamedSignalValue) and \
-           selector_signal.invalid in selector_signal.choices:
-            selector_signal.invalid = \
-                selector_signal.choices[selector_signal.invalid]
+        if not isinstance(selector_signal.raw_invalid, NamedSignalValue) and \
```
andlaus, line 1242: In my understanding, `raw_` values should never be `NamedSignalValue`s? (rather something like `Union[int, float, bytes]`?)

cantools/database/can/formats/arxml/system_loader.py
```diff
@@ -1440,8 +1439,7 @@ def _load_signal(self, i_signal_to_i_pdu_mapping):
             return None
 
         # Default values.
-        initial = None
-        invalid = None
+        raw_initial = None
         minimum = None
         maximum = None
```
andlaus, line 1444: shouldn't `.minimum` and `.maximum` be handled the same way as `.invalid` and `.initial`?

cantools/database/can/formats/arxml/system_loader.py
```diff
@@ -1478,49 +1476,40 @@ def _load_signal(self, i_signal_to_i_pdu_mapping):
         # loading initial values is way too complicated, so it is the
         # job of a separate method
         initial = self._load_arxml_init_value_string(i_signal, system_signal)
```
andlaus, line 1478: that should be `raw_initial`?!

cantools/database/can/signal.py
```diff
@@ -247,12 +249,26 @@ def __init__(self,
         #: ``True``.
         self.is_signed: bool = is_signed
 
-        #: The initial value of the signal, or ``None`` if unavailable.
-        self.initial: Optional[int] = initial
+        # deprecated `initial` attribute
+        self._initial: Optional[Union[int, SignalValueType]] = raw_initial
```
andlaus, line 253: AFAIK, internal values can also be `float` or `bytes`. (not sure if the latter is currently implemented by anything, though...)

also -- because some file formats currently pass internal and others physical values here -- we need to keep this as a separate constructor argument, e.g., `legacy_initial`...

cantools/database/can/signal.py
```diff
@@ -247,12 +249,26 @@ def __init__(self,
...
 
-        #: The value representing that the signal is invalid,
+        #: The raw initial value of the signal, or ``None`` if unavailable.
```
andlaus, line 255:
```suggestion
        #: The internal representation of the initial value of the signal, or ``None`` if unavailable.
```

cantools/database/can/signal.py
```diff
@@ -247,12 +249,26 @@ def __init__(self,
...
+        self.raw_initial: Optional[int] = raw_initial
+
+        #: The scaled initial value of the signal, or ``None`` if unavailable.
```
andlaus, line 258:
```suggestion
        #: The initial value of the signal in units of the physical world, or ``None`` if unavailable.
```

cantools/database/can/signal.py
```diff
@@ -305,6 +321,41 @@ def __init__(self,
             # multi-lingual dictionary
             self.comments = comment
 
+    @property
+    def initial(self) -> Optional[Union[int, SignalValueType]]:
+        warnings.warn(
```
andlaus, line 326: will this produce a warning every time the property is accessed or only for the first time? (IOW, how annoying do we want to make these warnings?)

also: `.invalid`, `.minimum`, and `.maximum` need the same treatment.

cantools/database/can/signal.py
```diff
@@ -305,6 +321,41 @@ def __init__(self,
...
+    @initial.setter
+    def initial(self, value: Union[int, SignalValueType]) -> None:
+        warnings.warn(
+            "The `Signal.initial` attribute is deprecated and scheduled for removal in April 2024. "
+            "Use `Signal.raw_initial` or `Signal.scaled_initial` instead.",
+            DeprecationWarning,
+        )
+        self._initial = value
+
+    def raw_to_scaled(self, raw: int) -> SignalValueType:
```
andlaus, line 342: a docstring for functions meant for public consumption would be nice.

cantools/database/can/signal.py
```diff
@@ -191,8 +193,8 @@ def __init__(self,
                  length: int,
                  byte_order: ByteOrder = 'little_endian',
                  is_signed: bool = False,
-                 initial: Optional[int] = None,
-                 invalid: Optional[int] = None,
+                 raw_initial: Optional[int] = None,
+                 raw_invalid: Optional[int] = None,
```
andlaus, line 197:
```suggestion
                 raw_initial: Optional[Union[int, float]] = None,
                 raw_invalid: Optional[Union[int, float]] = None,
```

(and possibly also `bytes`, but I think this is not supported by anything at the moment...)

cantools/database/can/signal.py
```diff
@@ -191,8 +193,8 @@ def __init__(self,
                  length: int,
```
andlaus, line 193: TODO for another day: we should make this constructor (and others like it, erhm `Message`) keyword-only: instantiating an object with a dozen or so positional parameters is asking for trouble IMO...

cantools/database/can/signal.py
```diff
@@ -247,12 +249,26 @@ def __init__(self,
...
+        )
+
+        #: The raw value representing that the signal is invalid,
+        #: or ``None`` if unavailable.
+        self.raw_invalid: Optional[int] = raw_invalid
+
+        #: The scaled value representing that the signal is invalid,
         #: or ``None`` if unavailable.
-        self.invalid: Optional[int] = invalid
+        self.invalid: Optional[SignalValueType] = (
```
andlaus, line 269: this should be `self.scaled_invalid`?

cantools/database/can/signal.py
```diff
@@ -305,6 +321,41 @@ def __init__(self,
...
+        )
+        self._initial = value
+
+    def raw_to_scaled(self, raw: int) -> SignalValueType:
+        with contextlib.suppress(KeyError, TypeError):
+            return self.choices[raw]  # type: ignore[index]
+
+        return raw * self.scale + self.offset
+
+    def scaled_to_raw(self, scaled: SignalValueType) -> int:
```
andlaus, line 348:
```suggestion
    def scaled_to_raw(self, scaled: SignalValueType) -> Union[int, float]:
```

cantools/database/can/signal.py
```diff
@@ -305,6 +321,41 @@ def __init__(self,
...
+        if isinstance(scaled, (float, int)):
+            _transform = float if self.is_float else round
+            if self.offset == 0 and self.scale == 1:
+                # treat special case to avoid introduction of unnecessary rounding error
+                return _transform(scaled)  # type: ignore[operator,no-any-return]
```
andlaus, line 353: once the return type is adapted above, the `type: ignore` should become unnecessary...

cantools/database/can/signal.py
```diff
@@ -305,6 +321,41 @@ def __init__(self,
...
+
+            return _transform((scaled - self.offset) / self.scale)  # type: ignore[operator,no-any-return]
+
+        return self.choice_string_to_number(str(scaled))
```
andlaus, line 357:
```suggestion
        if isinstance(scaled, (str, NamedSignalValue)):
            return self.choice_string_to_number(str(scaled))
        raise NotImplemented(f'Conversion of physical values of type "{type(scaled)}" to raw data')
```

andlaus (changes requested):
There are still a few issues with this, but I like the direction.

The remaining file formats (`SYM` and `KCD`?) need to be adapted as well...

zariiii9003:
You're too fast, it's not ready for review. But thanks for all the hints 😄 
Could you help me with this test? Where is this `0` in the arxml file? 

https://github.com/cantools/cantools/blob/df21fe3db7144e7e72e9882a661c5d4ea8c1dce7/tests/test_database.py#L4540

andlaus:
> You're too fast, it's not ready for review. But thanks for all the hints

don't let yourself be stressed by me ;)

> Could you help me with this test? Where is this 0 in the arxml file?

this should be specified as the initial dynamic part alternative for the multiplexer (hopefully):

https://github.com/cantools/cantools/blob/master/tests/files/arxml/system-3.2.3.arxml#L361-L365

zariiii9003:
@andlaus I'm not sure anymore. Maybe it is indeed better to add `raw_invalid` and `raw_initial`, and make `initial`, `invalid`, `minimum` and `maximum` all scaled without a `scaled_` prefix. It will break user code of dbc users, but if we bump the major version number and highlight it in the changelog, it should be acceptable.

andlaus:
> Maybe it is indeed better to add raw_invalid and raw_initial, and make initial, invalid, minimum and maximum all scaled without a scaled_ prefix. 

+1

> It will break user code of dbc users, but if we bump the major version number and highlight it in the changelog, it should be acceptable.

I would support that. If we get too many complaints ([ehem](#388)), we can add switch over to the prefixes...

cantools/database/can/formats/arxml/system_loader.py
zariiii9003: No, this is the string value, i'll rename the variable

cantools/database/can/signal.py
zariiii9003: mypy doesn't know the return value of `_transform`, so it will complain here

cantools/database/can/signal.py
zariiii9003: Done, but with TypeError

cantools/database/can/signal.py
zariiii9003: This is obsolete since invalid, minimum and maximum will not be renamed.

cantools/database/can/signal.py
andlaus: weird. how come?

zariiii9003: Why weird? 😄 

The python docs say: 

> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

andlaus (commented):
I'm happy with this. Do I see this correctly: `signal.{minimum,maximum}` are not yet handled and the non-ARXML and non-DBC file formats are not adapted yet?

cantools/database/can/signal.py
andlaus: I misinterpreted you: I thought that mypy complained about the new construct, while you probably said that you raise a `TypeError` instead of `NotImplemented`.

cantools/database/can/signal.py
```diff
@@ -321,28 +317,13 @@ def __init__(self,
...
-        )
-        self._initial = value
-
     def raw_to_scaled(self, raw: int) -> SignalValueType:
         with contextlib.suppress(KeyError, TypeError):
             return self.choices[raw]  # type: ignore[index]
 
+        if self.offset == 0 and self.scale == 1:
+            # treat special case to avoid introduction of unnecessary rounding error
+            return raw
```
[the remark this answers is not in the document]
zariiii9003: I'd disagree here. [PEP 484](https://peps.python.org/pep-0484/#the-numeric-tower) says:

> when an argument is annotated as having type float, an argument of type int is acceptable.

For large integers the float conversion might introduce a loss of precision for no good reason.

zariiii9003:
> I'm happy with this. Do I see this correctly: `signal.{minimum,maximum}` are not yet handled and the non-ARXML and non-DBC file formats are not adapted yet?

The tests pass 😄 
No, i didn't look at other formats to be honest. I will do that tomorrow.

Also, i did not really plan on changing `minimum` and `maximum`. The corresponding raw values would only depend on the bit count and data types. Is that really useful?

andlaus:
> No, i didn't look at other formats to be honest. I will do that tomorrow.

+1. if they use positional arguments to construct `Signal` objects it might even work as-is. (we should check before merging.)

> Also, i did not really plan on changing minimum and maximum. The corresponding raw values would only depend on the bit count and data types. Is that really useful?

useful: IDK, but for consistency's sake it would IMO be a good idea not to treat these quantities any different from `initial` and `invalid`. That said, I'm okay with merging the PR as-is and I'll follow up with a separate PR for `minimum` and `maximum`. (I can do this for you if you don't feel like it.)

zariiii9003:
I don't see any other usage of `initial` or `invalid`.

cantools/database/can/signal.py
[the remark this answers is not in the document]
zariiii9003: We could add `int` to `SignalValueType` to make it more explicit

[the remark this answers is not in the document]
andlaus: I'd be okay with that, but I'd prefer  `float` to be used for all signal values that represent a physical quantity. Anyway, not a biggie IMO...

andlaus:
> The corresponding raw values would only depend on the bit count and data types.

ah, btw: that's not necessarily true. it is possible to specify a narrower range than the scaled minimum and maximum of the raw value. e.g., if the necessary raw value does not align to a power of two...

zariiii9003:
> > The corresponding raw values would only depend on the bit count and data types.
> 
> ah, btw: that's not necessarily true. it is possible to specify a narrower range than the scaled minimum and maximum of the raw value. e.g., if the necessary raw value does not align to a power of two...

There's also the question: how do you define the raw min and max? The invalid and initial values are often not included in the physical min and max values. So, do the raw min/max values include intial and invalid or not?

andlaus:
yep, it makes sense to assign these quantities to a named value representing e.g. "Not Available". But as I see it, these are all reasons why `signal.{minimum,maximum}` should be handled the same as `signal.{invalid,initial}`?!

andlaus:
> So, do the raw min/max values include intial and invalid or not?

I just checked: For the files I have, the invalid value is never included, while for the initial one it depends: sometimes it is the same as invalid, sometimes it is a numeric value...

zariiii9003:
> > So, do the raw min/max values include intial and invalid or not?
> 
> I just checked: For the files I have, the invalid value is never included, while for the initial one it depends: sometimes it is the same as invalid, sometimes it is a numeric value...

And sometimes it is a separate named init value. 

For me, a raw min/max is too ambiguous. As a user, it might not be clear, whether it is a simple conversion of the physical min/max values or if it includes values outside the physical range.

andlaus:
> And sometimes it is a separate named init value.

in this case the non-postfixed member of `Signal` should be set to `None` while the corresponding `_raw' member should be set to the specified value. (I'd argue, though, that this is a bug of the respective database file, and should produce a warning...)

> For me, a raw min/max is too ambiguous. As a user, it might not be clear, whether it is a simple conversion of the physical min/max values or if it includes values outside the physical range.

As far as I understood this, the specified min/max values must always be in the physical range with the corresponding raw values. In some low-level cases -- e.g., writing database files -- one needs to use the raw values, so it would be more convenient if they were directly exposed. The potential confusion should IMO be fixed using the documentation.

On a different note: What I noticed while having a closer look at my files is that many signals are integers in the real world (sequence counters, checksums, etc.). maybe we should allow `signal.{scale,offset}` to be `None` and return an integer as the scaled type in this case. Stuff for another PR...

zariiii9003:
> On a different note: What I noticed while having a closer look at my files is that many signals are integers in the real world (sequence counters, checksums, etc.). maybe we should allow `signal.{scale,offset}` to be `None` and return an integer as the scaled type in this case. Stuff for another PR...

That should already be covered by this:
```python3
        if self.offset == 0 and self.scale == 1:
            # treat special case to avoid introduction of unnecessary rounding error
            return raw
```
That is the "no scaling" special case 😄

andlaus:
>  That should already be covered by this:

yeah, but IMO this strikes me as a bit inelegant ("duck typing"). Anyway, let's go this route for now.

I'd like to get this PR merged soonish: did you check the SYM and KCD parsers? if no, shall I help on that front?

zariiii9003:
> I'd like to get this PR merged soonish: did you check the SYM and KCD parsers? if no, shall I help on that front?

They do not use initial or invalid as far as i can see.

andlaus:
>  They do not use initial or invalid as far as i can see.

okay. feel free to merge, then!

andlaus (approved):

PR #538, opened 2023-03-22
Fix handling of initial values and allow the use of none-strict mode in Tester

hugfro:
When parsing a DBC file with initial values, scaling is always applied. The following line from _prepare_initial_signal_values uses signal.decimal
```
initial_sig_values[signal.name] = (signal.initial * signal.decimal.scale) + signal.decimal.offset
```
This means that all initial values are given the type decimal.Decimal. Consequently the function _encode_fields always call choice_string_to_number since the type check only looks for int and float:
```
if isinstance(value, (float, int)):
```

This PR adds
 - a check for decimal.Decimal type in  _encode_fields,
 - a check for non-scaling mode for initial values,
 - the option to run Tester in non-strict mode. 

The above changes hopefully solve the case where initial values are present but out of range. In addition, decimal type will be handled as such and not as string.

### DIFF
--- a/cantools/database/utils.py
+++ b/cantools/database/utils.py
@@ -2,6 +2,8 @@
 
 import os.path
 import re
+import decimal
+from ..database.can.signal import Signal
 from collections import OrderedDict
 from typing import (
     TYPE_CHECKING,
@@ -29,7 +31,6 @@ if TYPE_CHECKING:
     from ..database.can.attribute import Attribute
     from ..database.can.message import Message
     from ..database.can.node import Node
-    from ..database.can.signal import Signal
     from ..database.diagnostics import Data
 
 try:
@@ -85,6 +86,19 @@ def _encode_fields(fields: Sequence[Union["Signal", "Data"]],
                 continue
 
             unpacked[field.name] = _transform(value)  # type: ignore[operator]
+            continue
+        elif isinstance(value, decimal.Decimal) and isinstance(field, Signal):
+            _transform = decimal.Decimal if field.is_float else round
+            if scaling:
+                if field.offset == 0 and field.scale == 1:
+                    # treat special case to avoid introduction of unnecessary rounding error
+                    unpacked[field.name] = _transform(value)  # type: ignore[operator]
+                    continue
+
+            if field.decimal is not None:
+                unpacked[field.name] = _transform((value - field.decimal.offset) / field.decimal.scale | None)  # type: ignore[operator]
+            else:
+                raise ValueError(f"Offset and Scale are not available as the type Decimal for signal: {Signal}")
             continue
 
         unpacked[field.name] = field.choice_string_to_number(str(value))
--- a/cantools/tester.py
+++ b/cantools/tester.py
@@ -274,7 +274,13 @@ class Message(UserDict):
             maximum = 0 if not signal.maximum else signal.maximum
             if signal.initial:
                 # use initial signal value (if set)
-                initial_sig_values[signal.name] = (signal.initial * signal.decimal.scale) + signal.decimal.offset
+                if self.scaling:
+                    initial_sig_values[signal.name] = \
+                                                    (signal.initial *
+                                                     signal.decimal.scale) + \
+                                                    signal.decimal.offset
+                else:
+                    initial_sig_values[signal.name] = signal.initial
             elif signal.is_multiplexer:
                 initial_sig_values[signal.name] = mplex_settings.get(signal.name, 0)
             elif minimum <= 0 <= maximum:
--- a/cantools/tester.py
+++ b/cantools/tester.py
@@ -120,7 +120,8 @@ class Message(UserDict):
                  input_queue,
                  decode_choices,
                  scaling,
-                 padding):
+                 padding,
+                 strict):
         super().__init__()
         self.database = database
         self._mplex_map = invert_signal_tree(database.signal_tree)
@@ -129,6 +130,7 @@ class Message(UserDict):
         self.decode_choices = decode_choices
         self.scaling = scaling
         self.padding = padding
+        self.strict = strict
         self._input_list = input_list
         self.enabled = True
         self._can_message = None
@@ -249,7 +251,8 @@ class Message(UserDict):
         pruned_data = self.database.gather_signals(self.data)
         data = self.database.encode(pruned_data,
                                     self.scaling,
-                                    self.padding)
+                                    self.padding,
+                                    self.strict)
         self._can_message = can.Message(arbitration_id=arbitration_id,
                                         is_extended_id=extended_id,
                                         data=data)
@@ -315,7 +318,8 @@ class Tester:
                  on_message=None,
                  decode_choices=True,
                  scaling=True,
-                 padding=False):
+                 padding=False,
+                 strict=True):
         self._dut_name = dut_name
         self._bus_name = bus_name
         self._database = database
@@ -354,7 +358,8 @@ class Tester:
                                                        self._input_queue,
                                                        decode_choices,
                                                        scaling,
-                                                       padding)
+                                                       padding,
+                                                       strict)
 
         listener = Listener(self._database,
                             self._messages,


hugfro:
I just noticed that #515 addresses the issue of the decimal types. The proposed solution might be more refined. However the changes made in Tester are still relevant.

cantools/database/utils.py
```diff
@@ -86,6 +87,15 @@ def _encode_fields(fields: Sequence[Union["Signal", "Data"]],
...
             unpacked[field.name] = _transform(value)  # type: ignore[operator]
             continue
+        elif type(value) == decimal.Decimal:
+            _transform = decimal.Decimal if field.is_float else round
+            if scaling:
+                if field.offset == 0 and field.scale == 1:
+                    # treat special case to avoid introduction of unnecessary rounding error
+                    unpacked[field.name] = _transform(value)  # type: ignore[operator]
+                    continue
+            unpacked[field.name] = _transform((value - field.decimal.offset) / field.decimal.scale)  # type: ignore[operator]
```
deronek, line 97: If `field` is a `Data` object, this line raises an exception. I think we need an explicit check here

albsod: Could you please give an example that would raise an exception?

deronek:
```
import decimal
from cantools.database.diagnostics import Did, Data

data = Data(
    name='ExampleData',
    start=0,
    length=8,
)
did = Did(
    identifier=0xFF,
    name='ExampleDid',
    length=1,
    datas=[data]
)

# Will work fine
did.encode(
    {'ExampleData': 123},
    scaling=False
)

# Will raise an exception
did.encode(
    {'ExampleData': decimal.Decimal(123)},
    scaling=False
)
```

albsod: I get this too, but adding the additional check `isinstance(field, Signal)` gives the following error instead:

```
# Will raise an exception
>>> did.encode(
...     {'ExampleData': decimal.Decimal(123)},
...     scaling=False
... )
  File "/home/albin/.python-environments/.venv/lib/python3.11/site-packages/cantools/database/diagnostics/did.py", line 93, in encode
    encoded = encode_data(data,
              ^^^^^^^^^^^^^^^^^
  File "/home/albin/.python-environments/.venv/lib/python3.11/site-packages/cantools/database/utils.py", line 122, in encode_data
    unpacked = _encode_fields(fields, data, scaling)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/albin/.python-environments/.venv/lib/python3.11/site-packages/cantools/database/utils.py", line 109, in _encode_fields
    unpacked[field.name] = field.choice_string_to_number(str(value))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/albin/.python-environments/.venv/lib/python3.11/site-packages/cantools/database/diagnostics/data.py", line 62, in choice_string_to_number
    raise ValueError(f"Data {self.name} has no choices.")
ValueError: Data ExampleData has no choices.

```
By the way, I get the same error with the code in #515.

deronek: Yes, it highlights that although it's expected that `encode_data` can receive `decimal.Decimal` values to be encoded, `Data` class has no properties that would allow to encode them. I think it would be sensible for `Data` class to have similar implementation of `Decimal` as `Signal`.

hugfro: Thanks for the comment, I added a check for the data type `Signal`. Which leaves the Data class to be handled as before. 
Probably it would be a nice idea to implement `Decimal` in the `Data` class. However I believe that it is out of the scope for this PR.

cantools/database/utils.py
```diff
@@ -86,6 +87,15 @@ def _encode_fields(fields: Sequence[Union["Signal", "Data"]],
 
             unpacked[field.name] = _transform(value)  # type: ignore[operator]
             continue
+        elif isinstance(field, decimal.Decimal) and isinstance(field, Signal):
```
deronek, line 90: Should probably be `isinstance(value, decimal.Decimal)` instead. This condition will never trigger otherwise

deronek:
https://github.com/cantools/cantools/blob/5bc6ff3abdbba30169117a8fe2bb483f8b1e1830/cantools/database/utils.py#L33

I think this line needs to be moved outside the conditional. Otherwise it seems to break as `Signal` is not imported

deronek (changes requested):

zariiii9003:
According to the type annotations, Decimal encoding is not supported:
```python3
SignalValueType = Union[int, float, str, "NamedSignalValue"]
SignalMappingType = Mapping[str, SignalValueType]

def _encode_fields(fields: Sequence[Union["Signal", "Data"]],
                   data: SignalMappingType,
                   scaling: bool,
                   ) -> Dict[str, Union[int, float]]:
```

bitstruct cannot encode Decimal instances. And even if it could, what would be the point of doing that? The Decimal would be converted into binary data with a limited resolution anyway, so why not use float or int instead?

The Decimals should be used for parsing and writing database files only.

hugfro:
Thank you for the comment, the reason for this PR was because of the line ` initial_sig_values[signal.name] = (signal.initial * signal.decimal.scale) + signal.decimal.offset` in `tester.py`.
This line introduces the decimal typ if a signal has an initial value. The decimal type propagates down to `utils.py`  and the `_encode_fields` function. For my dbc file many signals have a initial value which will give them the type decimal.

Running the tester class with my dbc file raises exceptions due too that decimal type. 

Regarding:
> The Decimals should be used for parsing and writing database files only.

I am then unsure of why the decimal type is used in the `tester.py` file. We are parsing the dbc file using the decimal type, but in order to use signals that are of the type decimal we would need to convert them to float anyway which would introduce rounding errors. 

I am not opposed to the idea of using floats, I am just unsure of how to use the current implementation. And if we were to cast the decimals to floats where should that be done? Should we in `utils.py` convert all decimal types to float before we encode them using the current implementation? Or should the `tester.py` not use decimal types since we will cast them to float anyway?

andlaus:
maybe we should think about removing the decimal stuff altogether: If I understood correctly, the reason why it was introduced was to avoid rounding errors. Given the fact that it leads to other bugs (like this one) and that it comes with a -- IMO -- pretty large maintenance burden, it might be reasonable to instead just accept these rounding errors?

@juleq: I believe that you have some skin in this game, right?

juleq:
I think the reason was to make sure load and dump produce equivalent output when nothing has changed in between. When float conversions happen, load and dump might not match any longer. And testing the load/dump roundtrip becomes harder then.

If we only want to use decimal in "private", we can use accessors to give the user a more type-friendly API.

andlaus:
during my struggle with `signal.{minimum,maximum}` today I also fell over this once or twice. I think we could reduce our assurances that roundtripped files are only "reasonably close" to the original, i.e., the round-tripped files should be compared semantically, not byte-for-byte. (as far as I can see this is relatively easy to do.)

zariiii9003:
> Thank you for the comment, the reason for this PR was because of the line ` initial_sig_values[signal.name] = (signal.initial * signal.decimal.scale) + signal.decimal.offset` in `tester.py`. This line introduces the decimal typ if a signal has an initial value. 

I see, that makes sense. I actually removed the decimal in #536 already.

juleq:
> during my struggle with `signal.{minimum,maximum}` today I also fell over this once or twice. I think we could reduce our assurances to that roundtripped files are only "reasonably close" to the original, i.e., the round-tripped files should be compared semantically, not byte-for-byte. (as far as I can see this is relatively easy to do.)

True, but one gets a (unexpectedly) colorful diff when e.g. just using cantools to modify ids in a dbc. But since I am not so much available for maintenance, it is not for me to object 😉.

andlaus:
is this issue fixed by #555?

hugfro:
Running cantools from master the issues referenced in this PR no longer exists. I will close this PR.

PR #541, opened 2023-03-31
Misc fixes

andlaus:
This PR contains a bunch of smaller fixes and improvements for stuff which I recently stumbled over and which are IMO not enough to open a pull request on their own. Probably the change with most potential for controversy is 298d2c8718ed74e7fb378d156c2f26b7738d5b29, which causes the `list` tool to treat `signal.{initial,invalid,minimum,maximum}` as physical values. (The ARXML loader has been doing this for quite some time by now, the other formats will hopefully be switched to using physical quantities for all user-facing values [soon](https://github.com/cantools/cantools/pull/536).)


andlaus:
@zariiii9003 or @juleq: can you have a look at this? For this PR I recommend looking at the patches individually as they are all pretty small and should be self-contained. (in principle, each of them could be done as a separate PR...)

zariiii9003:
Yes, i wanted to get #536 merged first, i'll get that done today, i think.

cantools/subparsers/list.py
```diff
@@ -1,11 +1,42 @@
...
+    - ``_format_val(1.234, 'm', '%.2f')`` results in '1.23 m'
+    - ``_format_val('IAmAnEnum', 'm', '%.2f')`` results in 'IAmAnEnum'
+    - ``_format_val(1.234)`` results in '1.234'
+    """
+    if val is None or \
+       not unit or \
+       isinstance(val, str) or \
+       isinstance(val, NamedSignalValue):
+
+        return f'{{:{value_format_specifier}}}'.format(val)
```
zariiii9003, line 26: You could just use an f-string without the `.format` call

cantools/subparsers/list.py
```diff
@@ -1,11 +1,42 @@
...
+
+    e.g.
+    - ``_format_val(1.234, 'm', '%.2f')`` results in '1.23 m'
+    - ``_format_val('IAmAnEnum', 'm', '%.2f')`` results in 'IAmAnEnum'
+    - ``_format_val(1.234)`` results in '1.234'
+    """
+    if val is None or \
+       not unit or \
+       isinstance(val, str) or \
+       isinstance(val, NamedSignalValue):
```
zariiii9003, line 24: This could be refactored into sth like `needs_unit = isinstance(val, (int, float)) and unit is not None`

cantools/subparsers/list.py
```diff
@@ -1,11 +1,42 @@
...
+
+        return f'{{:{value_format_specifier}}}'.format(val)
+
+    return f'{{:{value_format_specifier}}} {unit}'.format(val)
+
+
+def _print_message(message,
+                   indent='',
+                   no_format_specifics=False,
```
zariiii9003, line 33: `no_format_specifics=False` is a bit awkward. It's not really clear what that means, without looking further into the code. You could add type annotations here, too.

cantools/subparsers/list.py
```diff
@@ -215,11 +240,12 @@ def _do_list_nodes(can_db, args):
 
         _print_node(node)
 
-def _do_list_messages(can_db, args):
+def _do_list_messages(can_db, args, values_format_specifier):
```
zariiii9003, line 243: You could add type annotations.

cantools/subparsers/list.py
```diff
@@ -287,6 +315,13 @@ def add_subparser(subparsers):
         const=True,
         required=False,
         help='Do not print extended CAN messages.')
+    list_parser.add_argument(
+        '--no-format-specifics',
```
zariiii9003, line 319: Maybe `skip-format-specifics` would be more readable throughout the code.

cantools/subparsers/list.py
andlaus: I wanted to do this, but this is actually not so easy:

the format specifier is supposed to be passed as a function argument, and as far as I now, nested F-strings are not possible? the problem is this:

```python
value = 1.2345
format_specifier = "%.2f"

format_string = f"the value is {{value:{format_specifier}}}"
print(format_string) # how to get the "the value is 1.23"?
```

zariiii9003: This works:
```python3
value = 1.2345
format_specifier = ".2f"

format_string = f"the value is {value:{format_specifier}}"
print(format_string) # how to get the "the value is 1.23"?
```

cantools/subparsers/list.py
andlaus: I added annotations to the whole subparser. fully annotating the rest of cantools is a project for future me ;)

cantools/subparsers/list.py
andlaus: cool! I assumed that such recursive formatting constructs were not supported, but I obviously didn't check. Fixed, thanks!

zariiii9003 (approved):

PR #543, opened 2023-04-09
Activate pycodestyle checks

zariiii9003:


cantools/subparsers/monitor.py
```diff
@@ -409,8 +409,7 @@ def try_update_message(self):
     def _try_update_container(self, dbmsg, timestamp, data):
         try:
             decoded = dbmsg.decode(data, decode_containers=True)
-        except:
-            formatted = None
+        except:  # noqa: E722  # TODO: introduce cantools.DecodeError & cantools.EncodeError
```
zariiii9003, line 412: @andlaus We should introduce something like `cantools.DecodeError` & `cantools.EncodeError`. Currently the `decode` and `encode` functions can raise all kinds of exceptions, which make those hard to catch.

andlaus: I agree.

andlaus (approved):
Looks good

Issue #547, opened 2023-04-20
generate_c_source add #defines for signal names.


guy-radford-sunswap:
I am using the `generate_c_source` for a project I am writing and would find it useful to have `#define`'s for the signal names.

Would this be a feature worth adding to the project? I am happy to look at raising a PR.

Thank you

guy-radford-sunswap:
PR: #548

guy-radford-sunswap:
Is there a plan to do a release soon?
Thank you

andlaus:
I'd like to keep the number of API breaking releases to a minimum. I thus would like to get #555 merged for before releasing version 39.0.0. @zariiii9003: do you have any objections to #555?

guy-radford-sunswap:
That makes sense, I am pulling `master` for the moment until this gets released.

PR #548, opened 2023-04-20
Add option to generate #defines for the Signal names when generating C source.

guy-radford-sunswap:
This PR adds a command line option to include #defines for the CAN Signal names.

Format
```
#define DATABASE_NAME_MESSAGE_NAME_SIGNAL_NAME "signal_name"
```

This PR for issue: #547


README.rst
```diff
@@ -344,6 +344,16 @@ See `motohawk_no_floating_point_numbers.h`_ and
 `motohawk_no_floating_point_numbers.c`_ for the contents of the
 generated files.
 
+In this example we use ``--include_signal_names`` to generate
+``#define`` definitions for the CAN Frame Signal names.
+
+.. code-block:: text
+
+   $ python3 -m cantools generate_c_source --include_signal_names tests/files/dbc/motohawk.dbc
```
andlaus, line 352: that option should use minuses instead of underscores (`--include-signal-names`) and it should probably be called `--generate-signal-name-macros`. That said, I don't really see any harm in including these macros unconditionally, if it is decided that we want them...

tests/files/c_source/motohawk_include_signal_names.c
```diff
@@ -0,0 +1,175 @@
+/**
```
andlaus, line 1: this, and the corresponding in this PR, seems to be spurious: It should be generated as part of running the tests?

andlaus (commented):
In principle I'm not opposed to this, but I have some trouble in seeing the use case: wouldn't it be easier to just write `"signal_name"` instead of `DATABASE_NAME_MESSAGE_NAME_SIGNAL_NAME` in your code?

guy-radford-sunswap:
Updated based on feedback. Thank you

tests/files/c_source/motohawk_include_signal_names.h
```diff
@@ -0,0 +1,211 @@
+/**
```
andlaus, line 1: you also need to remove this file from the PR...

andlaus:
thanks. Can you please explain what the advantage of these macros relative to just using the strings with the names directly is?

guy-radford-sunswap:
Using "signal_name" is open typos and errors, whereas a #define will allow the IDE to predict and highlight errors, also a static analyzer will also detect this before you get to running the tests.

andlaus:
fair enough. can you put the message names through the same treatment?

guy-radford-sunswap:
Yes sure :) Do you want the message(frame) names to be optional too or lumped in with the signal names?

andlaus:
As I see it there's no reason not to generate all of these macros unconditionally. (I might have missed something...)

guy-radford-sunswap:
Makes sense to me... other then then files will be a little bigger, but the compiler will removed anything that is unsed.

andlaus:
disk storage is cheap these days ;) (compile time might be another issue, but I doubt that this will be more than a few milliseconds even for pretty large databases.)

guy-radford-sunswap:
Done

* Added Frame Name Macros
* Added Signal Name Macros
* Updated Test `expect` files to support auto-generation of the above Macros.

andlaus:
all green, let's merge. Thanks for your contribution.

PR #552, opened 2023-04-30
Add default signal initialization function to C sources generator

ArseniSoitu:
Hello.

Could it be useful to generate C initialization function for every message,
which can set all fields(signals) to initial values from .dbc file?

For example, in case of generating sources from example "vehicle.dbc" database we have:

```
int vehicle_rt_imu06_accel_init(struct vehicle_rt_imu06_accel_t *msg_p)
{
    if (msg_p == NULL) return -1;
    memset(msg_p, 0, sizeof(msg_p));

    uint8_t tmp_validity_accel_lateral = 1;
    memcpy((void*)(&msg_p->validity_accel_lateral), (void*)&tmp_validity_accel_lateral, 1);

    uint8_t tmp_validity_accel_vertical = 0;
    memcpy((void*)(&msg_p->validity_accel_vertical), (void*)&tmp_validity_accel_vertical, 1);

    uint8_t tmp_accuracy_accel = 127;
    memcpy((void*)(&msg_p->accuracy_accel), (void*)&tmp_accuracy_accel, 1);

    int16_t tmp_accel_longitudinal = 32767;
    memcpy((void*)(&msg_p->accel_longitudinal), (void*)&tmp_accel_longitudinal, 2);

    int16_t tmp_accel_lateral = -30000;
    memcpy((void*)(&msg_p->accel_lateral), (void*)&tmp_accel_lateral, 2);

    int16_t tmp_accel_vertical = 16120;
    memcpy((void*)(&msg_p->accel_vertical), (void*)&tmp_accel_vertical, 2);

    return 0;
}
```

And in case of absence of initial values we have:
```
int vehicle_rt_sb_speed_init(struct vehicle_rt_sb_speed_t *msg_p)
{
    if (msg_p == NULL) return -1;
    memset(msg_p, 0, sizeof(msg_p));

    return 0;
}
```

In both cases there is a check for nullptr and filling all fields with zeroes by default.
So we can use it like this:
```
struct vehicle_rt_imu06_accel_t vehicle_rt_imu06_accel;
int result = vehicle_rt_imu06_accel_init(&vehicle_rt_imu06_accel);
...
```
In this case you can be sure that your object is initialized correctly by default


cantools/database/can/c_source.py
```diff
@@ -403,6 +402,27 @@
...
+ * @return zero(0) on success or (-1) in case of nullptr argument.
+ */
+int {database_name}_{message_name}_init(struct {database_name}_{message_name}_t *msg_p);
+'''
+
+MESSAGE_DEFINITION_INIT_FMT = '''\
+int {database_name}_{message_name}_init(struct {database_name}_{message_name}_t *msg_p)
+{{
+    if (msg_p == NULL) return -1;
+    memset(msg_p, 0, sizeof(msg_p));
```
andlaus, line 420: this is incorrect: `sizeof(msg_p)` is the size of the pointer itself, not the object it points to:

```suggestion
    memset(msg_p, 0, sizeof(*msg_p));
```

ArseniSoitu: Hello.

Thank for pointing it to me.
It's not clear for me, how to deal with tests related to comparing generated files to files from tests/files_c_source directory.
Should I manually update files in tests/files_c_source ?

andlaus: > Should I regenerate source files using dbc's from tests/files/dbc and then place the into tests/c_source/ ?

basically yes. Maybe the tests should test the API of the generated source files instead of comparing them  byte-for-byte, but that's a topic for another day...

ArseniSoitu: Thank you.
So, I'll regenerate source files from .dbc's and see how it will work

ArseniSoitu: Hello.
I'm still afraid of replacing existing files from tests/c_source/.
How do you think, maybe it will be better to add additional flag like `--gen-initial-values` to cantools?
In this case initializing functions will be added to C sources only in case of this flag is passed to the tool.
All the old tests will still work properly.
I'll have to generate new files with initializing functions inside and write new tests for those files.
But these new files and tests will not replace existing ones.

andlaus: > I'm still afraid of replacing existing files from tests/c_source/

this is pretty easy to do: remove the comment from https://github.com/cantools/cantools/blob/master/tests/test_command_line.py#L73 and run test_command_line:
```bash
pytest tests/test_command_line.py
```

this regenerates all C source files. if everything looks okay after inspecting what's changed via git , add the comment back to the test and commit the modified files.

> How do you think, maybe it will be better to add additional flag like --gen-initial-values to cantools?

I'd prefer not to add too many knobs like this, in particular because I don't think that there are any notable disadvantages in having these functions around even if they are unused ...

cantools/database/can/c_source.py
```diff
@@ -566,6 +567,14 @@
     {type_name} {name}{length};\
 '''
 
+INIT_SIGNAL_BODY_TEMPLATE_FMT = '''\
+    {type_name} tmp_{signal_name} = {signal_initial};
+    memcpy((void*)(&msg_p->{signal_name}),
+           (void*)&tmp_{signal_name},
+           sizeof({type_name}));
```
andlaus, line 574: why does a simple
```C
msg_p->{signal_name} = {signal_initial};
```
not work here?

cantools/database/can/c_source.py
```diff
@@ -1523,11 +1530,14 @@ def _generate_definitions(database_name, messages, floating_point_numbers, use_f
 
                 signal_definitions.append(signal_definition)
 
-            signals_init_body += init_signal_body_template.format(type_name=signal.type_name,
-                                                                  signal_initial=signal.initial,
-                                                                  signal_name=signal.snake_name,
-                                                                  signal_data_length=int(signal.type_length / 8)
-                                                                  ) if signal.initial is not None else ""
+            if add_initializers and signal.initial:
+                init_value = float(signal.initial) * signal.scale + signal.offset
```
andlaus, line 1534:
```suggestion
                init_value = signal.initial
```

since #536 `signal.initial`  uses physical values for all supported file formats to make it consistent with `signal.{minimum,maximum}`. (before #536 it depended on the file format loader.)

andlaus: also, it seems like `INIT_SIGNAL_BODY_TEMPLATE_FMT` expects the internal representation of the value, i.e.,
```suggestion
                init_value = signal.initial_raw
```

cantools/database/can/c_source.py
```diff
@@ -1653,7 +1664,8 @@ def generate(database,
              floating_point_numbers=True,
              bit_fields=False,
              use_float=False,
-             node_name=None):
+             node_name=None,
+             add_initializers=None):
```
andlaus, line 1668:
```suggestion
             add_initializers=False):
```

cantools/subparsers/generate_c_source.py
```diff
@@ -118,4 +119,8 @@ def add_subparser(subparsers):
     generate_c_source_parser.add_argument(
         '--node',
         help='Generate pack/unpack functions only for messages sent/received by the node.')
+    generate_c_source_parser.add_argument(
```
andlaus, line 122: as previously mentioned: I would like to generate these functions unconditionally..

cantools/database/can/c_source.py
ArseniSoitu: Okay, thank you.
I'll try this approach

cantools/database/can/c_source.py
ArseniSoitu: Yes, it works.
From my point of view, *memcpy()* is able to produce more efficient code

andlaus: this is certainly not true here: you first copy the data conventionally,  so you have two copy operations. also, -- in contrast to `memcpy` -- the compiler is aware of the exact properties of the type to be copied, so it  almost always produces code that is at least as fast...

cantools/database/can/c_source.py
ArseniSoitu: Thank you for pointing it out, my mistake.
I think exactly about assigning physical initial value in C code, not raw

cantools/subparsers/generate_c_source.py
ArseniSoitu: Okay, I'll remove additional argument

cantools/database/can/c_source.py
ArseniSoitu: But you are correct, I cannot work around the first assignment.
I'll fix it

cantools/database/can/c_source.py
```diff
@@ -568,10 +568,7 @@
 '''
 
 INIT_SIGNAL_BODY_TEMPLATE_FMT = '''\
-    {type_name} tmp_{signal_name} = {signal_initial};
-    memcpy((void*)(&msg_p->{signal_name}),
-           (void*)&tmp_{signal_name},
-           sizeof({type_name}));
+    msg_p->{signal_name} = {signal_initial};
```
andlaus, line 572: if I did get that correctly, the values in these structures are stored using the [internal representation of the signal values](https://github.com/cantools/cantools/blob/master/tests/files/c_source/motohawk.h#L90) ...

cantools/database/can/c_source.py
```diff
@@ -1530,14 +1526,9 @@ def _generate_definitions(database_name, messages, floating_point_numbers, use_f
...
-            if add_initializers and signal.initial:
-                init_value = float(signal.initial) * signal.scale + signal.offset
-                init_value = int(init_value) if not signal.is_float else float(init_value)
-                signals_init_body += INIT_SIGNAL_BODY_TEMPLATE_FMT.format(type_name=signal.type_name,
-                                                                          signal_initial=init_value,
-                                                                          signal_name=signal.snake_name,
-                                                                          signal_data_length=int(signal.type_length / 8)
-                                                                          )
+            if signal.initial:
+                signals_init_body += INIT_SIGNAL_BODY_TEMPLATE_FMT.format(signal_initial=signal.initial,
```
andlaus, line 1530: ... which makes this wrong, because `signal.initial` is the value of the signal in the physical world.
```suggestion
                signals_init_body += INIT_SIGNAL_BODY_TEMPLATE_FMT.format(signal_initial=signal.initial_raw,
```

ArseniSoitu: Yes, you're absolutely right, I made mistake yesterday.

Bit it led me to the following question:
Maybe it is also useful to have api for set/get signals in messages?

So that use case of initializing and further data accessing will look like this? 
(using **RT_IMU06_Accel** message from **vehicle.dbc** as example)
```
struct vehicle_rt_imu06_accel_t msg;
vehicle_rt_imu06_accel_init(struct vehicle_rt_imu06_accel_t *msg_p);
...
float physical_value = 100.0f;
vehicle_rt_imu06_accel_validity_accel_longitudinal_set_raw(&msg,
                                    vehicle_rt_imu06_accel_validity_accel_longitudinal_encode(physical_value));
...
float physical_value;
uint8_t raw_value;
vehicle_rt_imu06_accel_validity_accel_longitudinal_get_raw(&msg, &raw_value);
physical value =  vehicle_rt_imu06_accel_validity_accel_longitudinal_decode(raw_value);
```
What do you think about it?

andlaus: given that this is C, and C is supposed to be somewhat painful, I think it is not necessary: you can already assign physical values to a signal using
```C
msg.{signal_name} = {db_name}_{message_name}_{signal_name}_encode(1.234);
```

I think though, that in the long term a C++ generator should be added with an API like this:
```C++
msg.{signal_name}.assign_raw(1234);
msg.{signal_name}.assign(1.234);
```

I would want this codegen to use a proper template engine like [Jinja](https://jinja.palletsprojects.com), though...

ArseniSoitu: You're right, I wanted to decrease number of cases of explicit assignment to structure fields in user code

ArseniSoitu: Maybe modify api for decode like this?
```
float physical_value;
vehicle_rt_imu06_accel_validity_accel_longitudinal_get_real(&msg, &physical_value);
```
In this case function **vehicle_rt_imu06_accel_validity_accel_longitudinal_decode** is called internally.
I really think that explicitly accessing data members of object in user code is a good approach

andlaus: I don't really understand what the advantage of this would be compared to
```C
msg.validity_accel_longitudinal = vehicle_rt_imu06_accel_validity_accel_longitudinal_decode(physical_value);
```
(except for saving a few characters of boiler-plate. IMO that is a problem which is more or less intrinsic to plain-C.)

ArseniSoitu: Okay, I've understood your point of view.
Let's forget about setters/getters

cantools/database/can/c_source.py
```diff
@@ -417,11 +417,11 @@
  *
  * @return zero(0) on success or (-1) in case of nullptr argument.
  */
-int {database_name}_{message_name}_init(struct {database_name}_{message_name}_t *msg_p);
```
andlaus: since everything in the C code is in terms of raw values anyways, the concept of  initializing these objects to the init values is independent from  whether you do this using physical or raw values, these functions should not exhibit `_raw` suffixes. (they are basically the constructors of these structures...)

IOW, please revert this patch...

ArseniSoitu: Okay, I'll revert last changes

cantools/subparsers/generate_c_source.py
```diff
@@ -117,4 +117,8 @@ def add_subparser(subparsers):
...
+        '--add-initializers',
```
andlaus, line 121: this argument seems to be spurious now.

ArseniSoitu: Apologize for inattentiveness.
Fixed

andlaus:
ok, looks all good now, IMO. let's merge. Thanks for your persistence!

ArseniSoitu:
@andlaus, thank you for your patience and readiness to conversation

PR #554, opened 2023-05-08
Add group_comment_signals flag to DBC output

mon:
Some more DBC output knobs to twiddle :)

Currently, comments are serialised as
- Node comment 1
- Node 1 message comment 1
- Node 1 message comment n
- Node comment 2
- Node 2 message comment 1

This changes things to be
- Node comment 1
- Node comment 2
- Node 1 message comment 1
- Node 1 message comment n
- Node 2 message comment 1

### DIFF
--- a/cantools/database/can/database.py
+++ b/cantools/database/can/database.py
@@ -323,11 +323,15 @@ class Database:
                       sort_attribute_signals:type_sort_signals=SORT_SIGNALS_DEFAULT,
                       sort_attributes:type_sort_attributes=None,
                       sort_choices:type_sort_choices=None,
-                      shorten_long_names:bool=True) -> str:
+                      shorten_long_names:bool=True,
+                      group_comment_signals:bool=False) -> str:
         """Return the database as a string formatted as a DBC file.
            sort_signals defines how to sort signals in message definitions
            sort_attribute_signals defines how to sort signals in metadata -
               comments, value table definitions and attributes
+           If group_comment_signals==True, `CM_ SG_` lines appear after all
+              `CM_ BO_` lines, instead of immediately after their containing
+              message
 
         """
         if not self._sort_signals and sort_signals == SORT_SIGNALS_DEFAULT:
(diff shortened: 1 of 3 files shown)


andlaus:
what's the point of this? (the result seems to be quite equivalent to me...)

mon:
I'm making a dbc formatter - cantools already has a wealth of ways to adjust how DBC files are serialised (signal order, attribute order, enum order, long name cutoffs), this just adds one more thing.

andlaus:
IMO we should strive to remove knobs if they are of limited use, not add new ones. In this particular case, the loaded database should even be identical, so I really don't see the point...

mon:
Alright, well, just to be clear: Are the previous knobs going to (eventually) be removed, or is it just a case of not adding new ones?

andlaus:
In the medium term, nothing existing will be removed, and even in the long term I'll only advocate to go after stuff which is has only esthetic effects on the produced output files (which, as far as I can see, is not much).

The reason why I'd like to limit options is that I'd like to keep cantools usable for the 98.5% of people who don't care about such things and forward the remaining people to tools like yours. (We can't possibly maintain this for each and every file format which we support.)

mon:
Ok! It's very understandable as a maintenance burden. I'll consider ways to decouple cantools output from my formatting steps.

andlaus:
if you publish your DBC-linter, we can mention it in the cantools README if you want...

mon:
Oh, that might be nice! It's very opinionated so far (and potentially even incompatible with some tools due to converting to Unix line endings), so I may have to consider whether to do a full release, or a "this is what is possible" gist that others can extend.

mon:
Closing for now, and hopefully I can revisit this later by publishing my formatter. Thanks for your patience explaining your reasoning!

andlaus:
thanks for your patience and understanding. don't let yourself be deterred to contribute in the future by this :)

mon:
Good news! I've been able to publish my formatter. Hopefully it explains why I've been asking for all these features :)
https://github.com/mon/dbc_format

PR #555, opened 2023-05-09
Remove usage of Decimal objects

andlaus:
The issue with the use of Decimal objects to avoid rounding errors is that most people (me included) were not very confident in its usage which made cantools (IMO) disproportionately harder to modify and it lead to lurking bugs as well as maintenance and performance issues if the decimal number objects were used for computations instead of the "normal"  ones.

The most difficult drawback of this patch is that DBC files cannot be compared byte-for-byte anymore because rounding errors can be introduced during a load-store round-trip. What we do instead is ensure that "round tripped" file is "close enough" to the original one. Besides getting around the issue that the two database files might not be 100% identical due to rounding errors, this approach also allows to compare databases stemming from different formats (with minor modifications to avoid comparing the file format specific information).

Unfortunately, with this patch the output of the C code generator changes slightly, causing some code churn...


andlaus:
@juleq: I would like you to have a look at this because I believe that you have some skin in that game...

juleq:
@andlaus Fine for me. If it eases maintenance, we should go for it.

ArseniSoitu:
Maybe convert `raw_inital` to `int` [here](https://github.com/cantools/cantools/pull/555/files#diff-1acab215ceba08bc3a7a4127077b79e3c900a87e14ab1c5779149d59c8f59943L1561)?
```
Like this:
...
if signal.initial:
    signals_init_body += INIT_SIGNAL_BODY_TEMPLATE_FMT.format(signal_initial=int(signal.raw_initial),
                                                              signal_name=signal.snake_name)
...
```
In this case there will be no floating point values in initialization functions

andlaus:
> Maybe convert `raw_inital` to `int` [here](https://github.com/cantools/cantools/pull/555/files#diff-1acab215ceba08bc3a7a4127077b79e3c900a87e14ab1c5779149d59c8f59943L1561)?
> [...]
> In this case there will be no floating point values in initialization functions

the problem with that is that `.raw_initial` is not always an integer. (only if `signal.is_float` is `False`.)

ArseniSoitu:
> > Maybe convert `raw_inital` to `int` [here](https://github.com/cantools/cantools/pull/555/files#diff-1acab215ceba08bc3a7a4127077b79e3c900a87e14ab1c5779149d59c8f59943L1561)?
> > [...]
> > In this case there will be no floating point values in initialization functions
> 
> the problem with that is that `.raw_initial` is not always an integer. (only if `signal.is_float` is `False`.)

So maybe like this?
```
if signal.initial:
    signals_init_body += INIT_SIGNAL_BODY_TEMPLATE_FMT.format(signal_initial=signal.raw_initial if signal.is_float else int(signal.raw_initial),
                                                              signal_name=signal.snake_name)
```

andlaus:
> So maybe like this?
> [...]

If I understood this correctly, that code should be a no-op...

> Also, old _format_decimal() method could be converted to smth like _format_value() and placed into the scope
of Signal class in c_source to produce accurate output

as long as nothing too fancy is needed to format the output (AFAICS this currently is `round(value, 5)` and `round(value)`), and as long as there are only a handful of places where this is needed, I prefer to do this in-place because it is easier to see what exactly happens...

guy-radford-sunswap:
> The most difficult drawback of this patch is that DBC files cannot be compared byte-for-byte anymore because rounding errors can be introduced during a load-store round-trip. What we do instead is ensure that "round tripped" file is "close enough" to the original one. Besides getting around the issue that the two database files might not be 100% identical due to rounding errors, this approach also allows to compare databases stemming from different formats (with minor modifications to avoid comparing the file format specific information).

Hi, 
Please can this be explained and maybe an example? 

Surely if you load a DBC file and then write it back to disk it should be 100% identical value wise? 
I don't see how the use of decimals for scaling etc can be allowed to vary between reading and writing? Those values are important and this could affect the encoded and decoded signal value.

Apologies if I have the wrong end of the stick.
Thank you
Guy

andlaus:
> Surely if you load a DBC file and then write it back to disk it should be 100% identical value wise?

yes, modulo rounding errors (mainly stemming from the fact that you have to convert `float` to/from strings using decimal representation). i.e. you will "only" get about 16 decimal digits of precision...

andlaus:
alright, I've rebased this on top of the latest master version. @zariiii9003: I'd like to get this merged and do a 39.0 release directly thereafter...

zariiii9003:
> alright, I've rebased this on top of the latest master version. @zariiii9003: I'd like to get this merged and do a 39.0 release directly thereafter...

i'm still not 100% convinced, that we should remove it.... 😬
The lossless roundtrip is nice and some users might rely on it, who knows? Maybe it's enough to make the `decimal` attribute of the `Signal` class private?

mon:
Coming from a less familiar contributor - I like the fact that cantools generally doesn't mess with the file. Most of my issues with `Decimal` came from there being no type hints or obvious naming to identify something like a signal's `value` as being anything other than a `float`.

Could the same goal be reached by type annotating Decimal where it appears?

andlaus:
> I like the fact that cantools generally doesn't mess with the file.

Be aware that if the file in question was not dumped by cantools, it almost certainly does mess with it. Also, as far as I know, it currently only works well with DBC files, while for other file formats significantly fewer efforts where made in this regard; For ARXML (I currently have a private PoC branch for write support) this would even be next to impossible AFAICS...

> Most of my issues with Decimal came from there being no type hints or obvious naming to identify something like a signal's value as being anything other than a float.

As far as I can see, this is a typical experience which leads to [quite](#515) [a](https://github.com/cantools/cantools/commit/a96b22ac4fc777becdc3ea08e848bf30ad46b9f0#diff-1acab215ceba08bc3a7a4127077b79e3c900a87e14ab1c5779149d59c8f59943L860-L868) [few](https://github.com/cantools/cantools/commit/be9abad6d3835b2b2fa55410dac29908e9347fbc#diff-d8ec78b16f54213d3dcffd8815e15934fd86dc7c136acc36a38a5bb60758fdfeL274) [issues](https://github.com/cantools/cantools/commit/db7712ab94022e89312a8b5e13224b1fe4666967), even for seasoned contributors.

@zariiii9003 
> The lossless roundtrip is nice and some users might rely on it, who knows? Maybe it's enough to make the decimal attribute of the Signal class private?

Yeah. Maybe a lot of the confusion stems from the fact that this is not marked as private. Another issue that personally confuses me a lot is that the class for storing these objects in `Signal` is called [`Decimal`](https://github.com/cantools/cantools/blob/master/cantools/database/can/signal.py#L12) (defined in `signal.py`). IMO, the least which we should do is to get rid of that class and store what it contains separately as standard `decimal.Decimal` objects  (a la `signal._decimal_scale`, `signal._decimal_offset`, etc.)

That said, do you think that roundtripping is a very useful feature outside of unit tests? I think it would be much more useful to have lossless rountrip across file formats like:
```python
dbc_str = open('my_file.dbc').read()
orig_db = cantools.database.load_string(dbc_str, 'dbc')
sym_str = orig_db.as_sym_string()
sym_db = cantools.database.load_string(sym_str, 'sym')
dbc2_str = orig_db.as_dbc_string()
assert dbc_str == dbc2_str
```

(it currently might work in some cases, but I cannot see a way to guarantee this.)

for people who need to compare databases, we could promote the [`objects_similar()`](https://github.com/cantools/cantools/pull/555/files#diff-82fbc96aa3f082d4667c09a2a35aec050c73120de9174e1f37bef26ef9cd3115R30) function of this patch to the `Database` class, so that databases can be compared using something like
```python
db.is_similar(db2, tolerance=1e-8)
```

cantools/database/can/c_source.py
```diff
@@ -843,39 +842,26 @@ def _format_comment(comment):
...
+
     if minimum is not None and maximum is not None:
-        return '{}..{} ({}..{} {})'.format(
-            _format_decimal(signal._signal.scaled_to_raw(minimum)),
-            _format_decimal(signal._signal.scaled_to_raw(maximum)),
-            minimum,
-            maximum,
-            signal.unit)
+        return \
+            f'{phys_to_raw(signal._signal.scaled_to_raw(minimum))}..' \
```
zariiii9003, line 853: What's the purpose of this? `phys_to_raw(signal._signal.scaled_to_raw(minimum))`

andlaus: this ensures that the minimum/maximum printed here are of the correct type: for some signals, it can happen that the minimum or maximum is specified as an integer but the signal is defined to internally be `float` (i.e., `signal.is_float` is true, yet `isinstance(signal.maximum, int)` also applies. this happens for example with the `S9` signal in [`tests/files/dbc/multiplex_2.dbc`](https://github.com/cantools/cantools/blob/master/tests/files/dbc/multiplex_2.dbc).) While I consider this a minor bug in the file format loaders where this happens, I was not in the mood of hunting it down and the this work-around is not too bad IMO, mainly because it makes the code a bit more fool-proof...

In case you are wondering about `signal._signal`: the `signal` object here is of type [`Signal` from `c_source.py`](https://github.com/cantools/cantools/blob/master/cantools/database/can/c_source.py#L582), and the conversion stuff did not work for it for some reason which I forgot. (I think this code should be refactored...)

andlaus: ok, I renamed `phys_to_raw` to `cast_to_raw_type`. I hope the new name is much clearer...

andlaus: I also fixed the minor bug in the conversion code which caused this: https://github.com/cantools/cantools/pull/555/commits/084d48e83a47a7935659b36049935fb18b1b55cb

cantools/database/conversion.py
```diff
@@ -175,7 +175,7 @@ def numeric_scaled_to_raw(
             _raw = quotient
         else:
             _raw /= self.scale
-        return round(_raw)
+        return float(_raw) if self.is_float else round(_raw)
```
zariiii9003, line 178: `self.is_float` is always `False` here, so this is not necessary.

tests/files/c_source/floating_point_use_float.c
```diff
@@ -326,12 +326,12 @@ int floating_point_use_float_message3_init(struct floating_point_use_float_messa
 
 uint8_t floating_point_use_float_message3_signal3_encode(float value)
 {
-    return (uint8_t)((value - -0.125f) / 0.001f);
+    return (uint8_t)((value - -0.125) / 0.001);
```
zariiii9003, line 329: Why did you change the float-literal to a double?

andlaus:
okay: I've rebased this on the current master and also moved the "fuzzy compare" function to the Database (`db1.is_similar(db2)`) class and added a parameter that allows it to not include format specific information, i.e., databases stemming from different file formats can now be compared.

@zariiii9003: Do you think this can be merged soon? (If not, I'd like to tag a release without it this week...)

zariiii9003 (approved):

zariiii9003:
@andlaus let's start with a pre-release (e.g. beta or release candidate) to reduce the amount of new issues and complaints

andlaus:
> @andlaus let's start with a pre-release (e.g. beta or release candidate) to reduce the amount of new issues and complaints

I agree, but how to do this? as far as I know the release process currently works by adding a tag in git and the rest is done via github actions (including the upload to pip). IOW, how can we tell pip that the new release is not supposed to be the default one?

thanks for the approval, BTW...

zariiii9003:
I think it only depends on the version number. "39.0.0rc0" should work.

andlaus:
okay, let's try this...

andlaus:
done: https://github.com/cantools/cantools/releases/tag/39.0.0-rc1 . Let's see how it works out...

mon:
The only major change I'm seeing in serialisation is all my `GenSigStartValue` (which is a FLOAT) defs are now being written out as `0.0` instead of `0`. I quite prefer this, as it makes it clear that the values are floating point. It's all looking good, thanks for your work!

PR #561, opened 2023-05-11
rename `field` to `signal`

andlaus:
this renames the `field` variables of the encoding and decoding utility functions back to `signal` and prepares them for piecewise linear functions.

### DIFF
--- a/cantools/autosar/end_to_end.py
+++ b/cantools/autosar/end_to_end.py
@@ -1,14 +1,14 @@
 # Utilities for calculating the CRC of the AUTOSAR end-to-end
 # protection specification
 
-from typing import Optional, Union
+from typing import ByteString, Optional, Union
 
 import crccheck  # type: ignore
 
 from ..database.can.message import Message
 
 
-def compute_profile2_crc(payload : Union[bytes, bytearray],
+def compute_profile2_crc(payload: ByteString,
                          msg_or_data_id: Union[int, Message]) -> Optional[int]:
     """Compute the CRC checksum for profile 2 of the AUTOSAR end-to-end
     protection specification.
@@ -53,7 +53,7 @@ def compute_profile2_crc(payload : Union[bytes, bytearray],
     # do the actual work
     return int(crccheck.crc.Crc8Autosar().calc(crc_data))
 
-def apply_profile2_crc(payload : Union[bytes, bytearray],
+def apply_profile2_crc(payload: ByteString,
                        msg_or_data_id: Union[int, Message]) \
   -> Optional[bytearray]:
     """Compute the CRC checksum for profile 2 of the AUTOSAR end-to-end
@@ -74,7 +74,7 @@ def apply_profile2_crc(payload : Union[bytes, bytearray],
     return result
 
 
-def check_profile2_crc(payload : Union[bytes, bytearray],
+def check_profile2_crc(payload: ByteString,
                        msg_or_data_id: Union[int, Message]) -> Optional[bool]:
     """Check if the AUTOSAR E2E checksum for profile 2 of the AUTOSAR
     end-to-end protection specification is correct.
@@ -92,7 +92,7 @@ def check_profile2_crc(payload : Union[bytes, bytearray],
 
     return crc == crc2
 
-def compute_profile5_crc(payload : Union[bytes, bytearray],
+def compute_profile5_crc(payload: ByteString,
                          msg_or_data_id: Union[int, Message]) -> Optional[int]:
     """Compute the CRC checksum for profile 5 of the AUTOSAR end-to-end
     protection specification.
@@ -140,7 +140,7 @@ def compute_profile5_crc(payload : Union[bytes, bytearray],
 
     return int(result)
 
-def apply_profile5_crc(payload : Union[bytes, bytearray],
+def apply_profile5_crc(payload: ByteString,
                        msg_or_data_id: Union[int, Message]) \
   -> Optional[bytearray]:
     """Compute the AUTOSAR E2E checksum for profile 5 of the AUTOSAR
@@ -164,7 +164,7 @@ def apply_profile5_crc(payload : Union[bytes, bytearray],
 
     return result
 
-def check_profile5_crc(payload : Union[bytes, bytearray],
+def check_profile5_crc(payload: ByteString,
                        msg_or_data_id: Union[int, Message]) -> Optional[bool]:
     """Check if the AUTOSAR E2E checksum for profile 5 of the AUTOSAR
     end-to-end protection specification is correct.
--- a/cantools/autosar/end_to_end.py
+++ b/cantools/autosar/end_to_end.py
@@ -1,14 +1,14 @@
 # Utilities for calculating the CRC of the AUTOSAR end-to-end
 # protection specification
 
-from typing import ByteString, Optional, Union
+from typing import Optional, Union
 
 import crccheck  # type: ignore
 
 from ..database.can.message import Message
 
 
-def compute_profile2_crc(payload: ByteString,
+def compute_profile2_crc(payload: bytes,
                          msg_or_data_id: Union[int, Message]) -> Optional[int]:
     """Compute the CRC checksum for profile 2 of the AUTOSAR end-to-end
     protection specification.
@@ -53,7 +53,7 @@ def compute_profile2_crc(payload: ByteString,
     # do the actual work
     return int(crccheck.crc.Crc8Autosar().calc(crc_data))
 
-def apply_profile2_crc(payload: ByteString,
+def apply_profile2_crc(payload: bytes,
                        msg_or_data_id: Union[int, Message]) \
   -> Optional[bytearray]:
     """Compute the CRC checksum for profile 2 of the AUTOSAR end-to-end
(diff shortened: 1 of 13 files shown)


andlaus:
as an additional bonus, this PR also adds "inverse choices dicts", which should significantly increase encoding performance for messages which contain many signals that exhibit a lot of choices. @zariiii9003: would be great if you gave this a spin with your DBC file...

zariiii9003:
> as an additional bonus, this PR also adds "inverse choices dicts", which should significantly increase encoding performance for messages which contain many signals that exhibit a lot of choices. @zariiii9003: would be great if you gave this a spin with your DBC file...


Sure. I'm not sure how many choices that message actually has, but decoding with `decode_choices=True` seems significantly faster. The other measurements are a little slower, probably because of additional function calls. I tested with python 3.11, the function call penalty might be higher with earlier python versions.
Overall, the performance looks fine to me.


### With `decode_choices`
Script:
```python3
unpacked = {signal.name: min(0x40, signal.maximum) for signal in message.signals}
packed = bytes([random.randint(0, 255) for _ in range(message.length)])

print(f"{len(message.signals)=}")
n = 10_000
dt = timeit.repeat('data = message.encode(unpacked, strict=False, scaling=True)', globals=globals(), repeat=5, number=n)
print(f"encode time: {min(dt)/n * 1e6} us")
dt = timeit.repeat('message.decode(packed, decode_choices=True, scaling=True)', globals=globals(), repeat=5, number=n)
print(f"decode time: {min(dt)/n * 1e6} us")
```

master:
```
len(message.signals)=39
encode time: 12.071810000270489 us
decode time: 18.893299999763258 us
```

field_to_signal:
```
len(message.signals)=39
encode time: 13.660729999901378 us
decode time: 7.343199999741046 us
```

### Without `decode_choices`
Script:
```python3
unpacked = {signal.name: min(0x40, signal.maximum) for signal in message.signals}
packed = bytes([random.randint(0, 255) for _ in range(message.length)])

print(f"{len(message.signals)=}")
n = 10_000
dt = timeit.repeat('data = message.encode(unpacked, strict=False, scaling=True)', globals=globals(), repeat=5, number=n)
print(f"encode time: {min(dt)/n * 1e6} us")
dt = timeit.repeat('message.decode(packed, decode_choices=False, scaling=True)', globals=globals(), repeat=5, number=n)
print(f"decode time: {min(dt)/n * 1e6} us")
```

master:
```
len(message.signals)=39
encode time: 12.39417000033427 us
decode time: 6.245800000033341 us
```

field_to_signal:
```
len(message.signals)=39
encode time: 14.142059999721823 us
decode time: 6.78241000023263 us
```

### Without `scaling`
Script:
```python3
unpacked = {signal.name: min(0x40, signal.maximum) for signal in message.signals}
packed = bytes([random.randint(0, 255) for _ in range(message.length)])

print(f"{len(message.signals)=}")
n = 10_000
dt = timeit.repeat('data = message.encode(unpacked, strict=False, scaling=False)', globals=globals(), repeat=5, number=n)
print(f"encode time: {min(dt)/n * 1e6} us")
dt = timeit.repeat('message.decode(packed, decode_choices=False, scaling=False)', globals=globals(), repeat=5, number=n)
print(f"decode time: {min(dt)/n * 1e6} us")
```

master:
```
len(message.signals)=39
encode time: 8.633569999801693 us
decode time: 4.039819999889005 us
```

field_to_signal:
```
len(message.signals)=39
encode time: 10.973010000088834 us
decode time: 4.345120000289171 us
```

cantools/database/can/signal.py
```diff
@@ -332,9 +331,9 @@ def raw_to_scaled(
         :return:
             The calculated scaled value
         """
-        if decode_choices:
-            with contextlib.suppress(KeyError, TypeError):
-                return self.choices[raw]  # type: ignore[index]
+        if decode_choices and self.choices and raw in self.choices:
+            assert isinstance(raw, int)
```
zariiii9003, line 335: `isinstance` is slow and we shouldn't use `assert` outside of test code.
I'd suggest `return self.choices[cast(int, raw)]`, because the `raw in self.choices` check guarantees, that `raw` is an integer.

cantools/database/can/signal.py
```diff
@@ -386,15 +386,17 @@ def comment(self, value: Optional[str]) -> None:
...
-
-        raise KeyError(f"Choice {string} not found in Signal {self.name}.")
+    def set_choices(self, choices: Optional[Choices]) -> None:
+        self.choices: Optional[Choices] = choices
+        self._inverse_choices: Optional[Dict[str, int]] = None
+        if choices is not None:
+            # we simply assume that the choices are invertible
+            self._inverse_choices = { str(x[1]): x[0] for x in choices.items() }
+
+    def choice_to_number(self, choice: Union[str, NamedSignalValue]) -> int:
```
zariiii9003, line 396: I hope you'll remember all these API changes when you create the next release 😄

cantools/database/utils.py
```diff
@@ -66,49 +67,43 @@ def start_bit(data: Union["Data", "Signal"]) -> int:
...
+        value = data[signal.name]
 
-        if isinstance(value, (str, NamedSignalValue)):
-            unpacked[field.name] = field.choice_string_to_number(str(value))
-            continue
+        raw_value: Union[int, float]
+        if scaling:
+            raw_value = signal.scaled_to_raw(value)
+        else:
+            if isinstance(value, (str, NamedSignalValue)):
```
zariiii9003, line 82: I think there is something wrong here. If `scaling` is False, then `value` must be the raw value. The `choice_to_number()` was already performed in `scaled_to_raw()`

cantools/database/utils.py
```diff
@@ -135,41 +130,39 @@ def decode_data(data: bytes,
...
-            else:
-                decoded[field.name] = value
-
+            value = unpacked[signal.name]
         except KeyError:
             if not allow_truncated:
                 raise
+            continue
+
+        if decode_choices and signal.choices is not None and value in signal.choices:
```
zariiii9003, line 156: Maybe we could simplify this too like in the encoding function, if the performance allows it:
```python3
if scaling:
    decoded[signal.name] = signal.raw_to_scaled(value, decode_choices)
else:
    decoded[signal.name] = value
```

zariiii9003 (changes requested):

cantools/database/can/signal.py
```diff
@@ -386,15 +386,17 @@ def comment(self, value: Optional[str]) -> None:
...
-            raise ValueError(f"Signal {self.name} has no choices.")
-
-        for choice_number, choice_value in self.choices.items():
-            if str(choice_value) == str(string):
-                return choice_number
-
-        raise KeyError(f"Choice {string} not found in Signal {self.name}.")
+    def set_choices(self, choices: Optional[Choices]) -> None:
+        self.choices: Optional[Choices] = choices
+        self._inverse_choices: Optional[Dict[str, int]] = None
```
zariiii9003, line 391: This should be defined in `__init__`

cantools/database/can/signal.py
andlaus: I've pushed a new version of the patch. I'm a bit out of my depths of where the remaining performance regressions stem from. (IMO they are acceptable, though.) Can you help out?

cantools/database/can/signal.py
andlaus: yeah. If not that should not be a too big deal as I'm pretty sure that not too many people use this function as it is pretty low level... (the old function name was a bit of a misnomer IMO, because it also accepted `NamedSignalValue` objects.)

cantools/database/utils.py
andlaus: this is pretty twisted:

> If scaling is False, then value must be the raw value.

it can also be a choice (i.e., 'str' or 'NamedSignalValue'). it can be argued that considering choices should only be done if `scaling` is `True`, thouhgh. (the existing code does it unconditionally, so I thought I keep it this way...)

cantools/database/utils.py
andlaus: yes, if we decide that choices are only considered if `scaling` is `True` and get rid of the `decode_choices` parameter. (I'm a bit on the line about whether this is a good idea.)

cantools/autosar/end_to_end.py
```diff
@@ -1,14 +1,14 @@
...
-from typing import Optional, Union
+from typing import ByteString, Optional, Union
 
 import crccheck  # type: ignore
 
 from ..database.can.message import Message
 
 
-def compute_profile2_crc(payload : Union[bytes, bytearray],
+def compute_profile2_crc(payload: ByteString,
```
zariiii9003, line 11: I think that's a bad idea https://github.com/python/mypy/issues/4871

andlaus:
do you consider this to be ready @zariiii9003?

andlaus:
> Data/Signal is also used in utils.create_encode_decode_formats. You could adapt those names, too.

good point. I hope to have captured the non-diagnostic code paths [here](https://github.com/cantools/cantools/pull/561/commits/ef775516497bf43609b2ee971db62caa17af6492)

zariiii9003 (approved):
I think this is fine. Let's merge this and then i'll have some rebasing or merging to do 😩

PR #563, opened 2023-05-12
Implement Conversion Classes

zariiii9003:
Followup to the discussion in #537
The tests pass and i don't see any performance regression yet.

@andlaus @1atabey1 This is still in draft stage, but i'd like to hear your opinion.

Open points:

- scale, offset, choices and is_float exist in both `Signal` and `BaseConversion` classes. At the very least, the `Signal` attributes should be converted to properties, which point to the attributes of `Signal.conversion`
- add comments where necessary
- documentation update


cantools/database/can/conversion.py
```diff
@@ -0,0 +1,190 @@
...
+    def raw_to_scaled(
+        self,
+        raw_value: Union[int, float],
+        decode_choices: bool = True,
+    ) -> SignalValueType:
+        if decode_choices:
+            try:
+                return self.choices[raw_value]  # type: ignore[index]
+            except (AttributeError, KeyError):
+                pass
```
andlaus, line 142: to make debugging easier, I prefer not to have exceptions in the happy path.
```suggestion
        if decode_choices and raw_value in self.choices:
            return self.choices[cast(int, raw_value)]
```
(possibly, this even might run slightly faster...)

cantools/database/can/conversion.py
```diff
@@ -0,0 +1,190 @@
...
+    if choices is None:
+        if scale == 1 and offset == 0:
+            return IdentityConversion(is_float=is_float)
+
+        if _is_integer(scale) and _is_integer(offset) and not is_float:
+            return LinearIntegerConversion(scale=int(scale), offset=int(offset))
+
+        return LinearConversion(scale=scale, offset=offset, is_float=is_float)
+
+    return NamedSignalConversion(scale=scale, offset=offset, choices=choices)
```
andlaus, line 186: hm. we possibly might need an `NamedSignalIntegerConversion` because it is possible to define scaling factor of 1 and an integer offset with some named values (alternatively, all conversion classes could either return integers or floats)

cantools/database/can/conversion.py
```diff
@@ -0,0 +1,190 @@
...
+    scale: float = 1,
+    offset: float = 0,
+    choices: Optional[Choices] = None,
+    is_float: bool = False,
+) -> BaseConversion:
+    if choices is None:
+        if scale == 1 and offset == 0:
+            return IdentityConversion(is_float=is_float)
+
+        if _is_integer(scale) and _is_integer(offset) and not is_float:
```
andlaus, line 181: this is not correct: even if the scaling factor is an integer, the result can still be a `float`:
```suggestion
        if scale == 1 and _is_integer(offset) and not is_float:
```

cantools/database/can/conversion.py
```diff
@@ -0,0 +1,190 @@
...
+        else:
+            self._inverse_choices = None
+
+    def choice_to_number(self, choice: Union[str, "NamedSignalValue"]) -> int:
+        if self._inverse_choices is None:
+            raise KeyError(f"Signal does not exhibit any choices")
+        return self._inverse_choices[str(choice)]
+
+
+def conversion_factory(
```
andlaus, line 171: depending on your taste, you might want to consider making this method a `@staticmethod` of `BaseConversion` so that it is called via something like this:
```python
BaseConversion.factory(...)
```

andlaus (commented):
I support this and would like to get it merged as soon as it is ready...

cantools/database/can/conversion.py
```diff
@@ -0,0 +1,190 @@
...
+            choices=None,
+            is_float=False,
+        )
+
+    def raw_to_scaled(
+        self,
+        raw_value: Union[int, float],
+        decode_choices: bool = True,
+    ) -> SignalValueType:
+        if decode_choices:
```
andlaus, line 138: `self.choices` can be `None` here (?):

```suggestion
        if decode_choices and self.choices:
```

cantools/database/can/conversion.py
andlaus, line 142: to make debugging easier, I prefer not to have exceptions in the happy path.
```suggestion
        if (choice := self.choices.get(raw_value)) is not None:
            return choice
```
(possibly, this even might run slightly faster...)

cantools/database/can/conversion.py
```diff
@@ -0,0 +1,190 @@
...
+        return round(_raw)
+
+
+class NamedSignalConversion(BaseConversion):
+    is_float = False
+
+    def __init__(self, scale: float, offset: float, choices: Choices) -> None:
+        self.scale = scale
+        self.offset = offset
+        self._inverse_choices: Optional[Dict[str, int]] = None
```
andlaus, line 122: if the inverse choices are optional here, the choices should be too. alternatively, just declare the type of `self._inverse_choices` here:
```suggestion
        self._inverse_choices: Optional[Dict[str, int]]
```

zariiii9003: Fixed: Choices should be mandatory for NamedSignalConversion

cantools/database/can/conversion.py
zariiii9003: Choices should be mandatory for NamedSignalConversion, so that check is unnecessary

cantools/database/can/conversion.py
andlaus: ok, I suppose I got confused by the inverse transformation: the proposed code is correct in the sense that it guarantees that any given integer raw value is to be scaled into an integer. I still don't like the "duck typed" feel of this, but I cannot come up with something better either..

cantools/database/can/conversion.py
zariiii9003: I was curious about this, so i measured it 😄 

```python3
import timeit


class MyClass:
    choices = {
        1: "1",
    }

    def raw_to_scaled1(self, raw_value):
        try:
            return self.choices[raw_value]  # type: ignore[index]
        except (AttributeError, KeyError):
            pass

        return None

    def raw_to_scaled2(self, raw_value):
        if (choice := self.choices.get(raw_value)) is not None:
            return choice

        return None


instance = MyClass()

n = 100_000
dt = timeit.repeat('instance.raw_to_scaled1(1)', globals=globals(), repeat=20, number=n)
print(f"{min(dt)/n * 1e6} us - try-except happy path")
dt = timeit.repeat('instance.raw_to_scaled1(2)', globals=globals(), repeat=20, number=n)
print(f"{min(dt)/n * 1e6} us - try-except happy miss")

dt = timeit.repeat('instance.raw_to_scaled2(1)', globals=globals(), repeat=20, number=n)
print(f"{min(dt)/n * 1e6} us - walrus happy path")
dt = timeit.repeat('instance.raw_to_scaled2(2)', globals=globals(), repeat=20, number=n)
print(f"{min(dt)/n * 1e6} us - walrus happy miss")
```

Python 3.11:
```
0.05078900000171416 us - try-except happy path
0.2410190000000512 us - try-except happy miss
0.0671630000033474 us - walrus happy path
0.06490199999916513 us - walrus happy miss
```

so yeah, yours looks better 👍

cantools/database/can/conversion.py
zariiii9003: I like this, done.

cantools/database/can/conversion.py
zariiii9003: NamedSignalConversion calls the factory internally, so it always uses the correct conversion anyway

zariiii9003:
Here's the performance with of this [PR](https://github.com/cantools/cantools/pull/563/commits/96c35a188b323b3cea59411d1decb9204d2bacb2) :

```
signal_count=39, signals_with_choices=36
encode scaled numeric: 12.786000006599352 us
encode scaled choices: 12.166000014985912 us
encode raw numeric: 7.9149999874061905 us
encode raw choices: 11.801000000559725 us
decode, no scaling, no choices: 4.096000011486467 us
decode, scaling, no choices: 8.483999990858138 us
decode, no scaling, choices: 16.614000014669728 us
decode, scaling, choices: 9.739999986777548 us
decode, no scaling, no choices, allow_truncated: 1.910999999381602 us
```

[Master branch](https://github.com/cantools/cantools/commit/e206109f0506e01eeddc492ce99983ffc53e8180):

```
signal_count=39, signals_with_choices=36
encode scaled numeric: 10.641999979270622 us
encode scaled choices: 22.46300002298085 us
encode raw numeric: 7.559000005130656 us
encode raw choices: 21.396000011009164 us
decode, no scaling, no choices: 3.9209999886224978 us
decode, scaling, no choices: 5.720000008295756 us
decode, no scaling, choices: 15.386000013677405 us
decode, scaling, choices: 17.01699999102857 us
decode, no scaling, no choices, allow_truncated: 3.9780000224709515 us
```

[andlaus/piecewise_linear](https://github.com/1atabey1/cantools/pull/2/commits/38deeed0f14f51ef32ec178abc9e1be5b9a380a3);

```
signal_count=39, signals_with_choices=36
encode scaled numeric: 22.53899998322595 us
encode scaled choices: 25.100000020756852 us
encode raw numeric: 10.096999976667576 us
encode raw choices: 22.768999988329597 us
decode, no scaling, no choices: 3.9979999928618786 us
decode, scaling, no choices: 11.583999985305127 us
decode, no scaling, choices: 5.404000003181864 us
decode, scaling, choices: 10.812000000441913 us
decode, no scaling, no choices, allow_truncated: 4.25300000642892 us
```

Measured with python 3.11 on a i7-11850H

andlaus:
> Here's the performance with of this PR [...]

interesting. I suppose the much better encoding performance is due to "inverse choice dictionaries"? the few performance regressions in your results are quite acceptable IMO...

zariiii9003:
@andlaus rebase is done

1atabey1 (approved):
I'm a big fan of this change and would also like to see it merged soon! It certainly makes the implementation of my piecewise linear conversion significantly simpler, thanks!

andlaus (approved):
LGTM

PR #565, opened 2023-05-16
DBC: parse bus comments even when no DBName specified

mon:
This PR fixes two discrepancies I found vs CANdb++'s bus comment behaviour (I assume CANdb++ is the standard to match here?).

Firstly, a DBName is not actually required for a bus comment to be valid, and the filename (without extension) is used if one isn't specified. This auto-name isn't serialised on save.

Secondly, multiple `CM_` entries get concatenated together. This behaviour is frankly a bit odd and feels wrong, especially since CANdb++ will then serialise the concatenated version to a single `CM_`. But since it's "correct" behaviour, I've added it.

I'm happy to amend the PR to remove the comment concatenation if you'd prefer to just drop extra CM_ entries (which is what cantools does at the moment).

Screenshot of CANdb++ behaviour on the newly added test dbc:
image


cantools/database/can/formats/dbc.py
```diff
@@ -1675,7 +1677,7 @@ def _load_bus(attributes, comments):
     try:
         bus_name = attributes['database']['DBName'].value
     except KeyError:
-        return None
+        bus_name = ''
```
andlaus, line 1680: not sure if empty bus names are valid...

andlaus (commented):
looks fair enough to me,  but @juleq is the expert here...

cantools/database/can/formats/dbc.py
mon: Yeah, this is my main concern as well. I'm sure `None` is more-invalid, so hopefully juleq can chime in with better detail.

Other options could be a dummy bus name "bus" or somehow bring forward the DBC filename to use as CANdb++ does.

andlaus:
can you see any obstacles to merging this, @juleq?

juleq:
Yes, candb++ is authoritative, thanks for checking and adapting. Merging now.

PR #569, opened 2023-05-24
dbc: dump_string: set GenSigStartValue from Signal.raw_initial

malsyned:
I modeled this after the way GenMsgCycleTime is synchronized with Message.cycle_time.

### DIFF
--- a/cantools/database/can/formats/dbc.py
+++ b/cantools/database/can/formats/dbc.py
@@ -573,6 +573,18 @@ def _create_GenMsgCycleTime_definition():
                                type_name='INT',
                                minimum=0,
                                maximum=2**16-1)
+def _create_GenSigStartValue_definition():
+    return AttributeDefinition('GenSigStartValue',
+                               default_value=0,
+                               kind='SG_',
+                               type_name='FLOAT',
+                               minimum=0,
+                               maximum=100000000000)
+
+def _need_startval_def(database):
+    return any(s.raw_initial is not None
+               for m in database.messages
+               for s in m.signals)
 
 def _dump_attribute_definitions(database):
     ba_def = []
@@ -587,6 +599,9 @@ def _dump_attribute_definitions(database):
     if 'GenMsgCycleTime' not in definitions:
         definitions['GenMsgCycleTime'] = \
             _create_GenMsgCycleTime_definition()
+    if 'GenSigStartValue' not in definitions and _need_startval_def(database):
+        definitions['GenSigStartValue'] = \
+            _create_GenSigStartValue_definition()
 
     def get_value(definition, value):
         if definition.minimum is None:
@@ -696,6 +711,9 @@ def _dump_attribute_definition_defaults(database):
     if 'GenMsgCycleTime' not in definitions:
         definitions['GenMsgCycleTime'] = \
             _create_GenMsgCycleTime_definition()
+    if 'GenSigStartValue' not in definitions and _need_startval_def(database):
+        definitions['GenSigStartValue'] = \
+            _create_GenSigStartValue_definition()
 
     for definition in definitions.values():
         if definition.default_value is not None:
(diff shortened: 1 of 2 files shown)


andlaus (approved):
LGTM. @juleq , @zariiii9003 ?

cantools/database/can/formats/dbc.py
```diff
@@ -778,9 +796,23 @@ def get_value(attribute):
...
+            sig_attributes = OrderedDict()
             if signal.dbc is not None and signal.dbc.attributes is not None:
-                for attribute in signal.dbc.attributes.values():
-                    attributes.append(('signal', attribute, None, message, signal))
+                sig_attributes = signal.dbc.attributes
+
+            # synchronize the attribute for the signal start value with
+            # the start value specified by the message object
+            if signal.raw_initial is None and 'GenSigStartValue' in sig_attributes:
+                del sig_attributes['GenSigStartValue']
```
zariiii9003, line 807: You are modifying `signal.dbc.attributes` here, because you reassigned `sig_attributes` on line 802. You should call `sig_attributes.update(sig_attributes)` or create a deep copy of `sig_attributes`.

malsyned: I wondered about that, but if that's a bug, it also exists the way `message.dbc.attributes['GenMsgCycleTime']` is handled in the current code right above my additions:
```python3
        # retrieve the ordered dictionary of message attributes
        msg_attributes = OrderedDict()
        if message.dbc is not None and message.dbc.attributes is not None:
            msg_attributes = message.dbc.attributes

        # synchronize the attribute for the message cycle time with
        # the cycle time specified by the message object
        if message.cycle_time is None and 'GenMsgCycleTime' in msg_attributes:
            del msg_attributes['GenMsgCycleTime']
```

It looks like that was by design according to the commit message for fcb2bcb1b5c0277034e61711aff06b541eee4dda. Looks like the original author was taking the opportunity at DBC write time to synchronize the Message DbcSpecifics with any changes made to the Message.cycle_time value. Since what I'm doing is an analogous operation on Signals, I used the same pattern.

If you think that's wrong, though, I can fix it. Do you want it fixed for `message.dbc.attributes['GenMsgCycleTime']` as well?

andlaus: > That was originally added in https://github.com/cantools/cantools/commit/fcb2bcb1b5c0277034e61711aff06b541eee4dda. @andlaus can you comment on what your thinking was there? It looks to me like you were taking the opportunity at DBC write time to synchronize the Message DbcSpecifics with any changes made to the Message.cycle_time value.

I don't remember exactly, but I think that was the reason why. I'm pretty sure this can be implemented more elegantly, though.

> Since what I'm doing is an analogous operation on Signals, I used the same pattern.

fine with me ;). if I understood @zariiii9003 correctly, he just wants to have this done in a more pythonic fashion?

andlaus: > You are modifying signal.dbc.attributes here, because you reassigned sig_attributes on line 802. You should call sig_attributes.update(signal.dbc.attributes) or create a deep copy of signal.dbc.attributes.

NB that the whole database is already deepcopied before it is written: https://github.com/cantools/cantools/blob/master/cantools/database/can/formats/dbc.py#L1862

zariiii9003: In my opinion, dumping the database should have no side effects. The database should remain unchanged.

malsyned: Given that the deepcopy already happens at the top (I'd missed that -- thanks @andlaus) I think your request is already met by this PR in its current form, do you agree @zariiii9003? Does it need anything else before you'd want to merge it?

zariiii9003: > > In my opinion, dumping the database should have no side effects. The database should remain unchanged.
> 
> agreed. that's why the database object is deepcopied before doing the modifications. (It's a bit hacky but it has been like this since at least 2020 (cf [89b974cb3b0921a057bff352afe84fa844e23a8d](https://github.com/cantools/cantools/commit/89b974cb3b0921a057bff352afe84fa844e23a8d#diff-5f3a4048a6ecb5c96501d101dd167bb439681e5c3eb0b1a1e7d024dfe0f2b8f1R1443))

I see. That module needs some refactoring 😄

zariiii9003 (approved):

cantools/database/can/formats/dbc.py
andlaus: agreed. it is far from the only one, though ;)

PR #577, opened 2023-06-07
Provide docker container and make all tests run

wolframkriesing:
This PR provides the following:
- [x] allows to start a docker container where all the things get installed into and all tests etc. can be run in
- [x] adjust the installation instructions, it said to install using `requirements.txt` but there is none, adjusted the docs here
- [x] make all the tests `make test` pass (which meant adding some deps and tiny fixes in the tests)

These were all the things I needed to have `make test` pass, I am not sure if this is the way it is wanted, so I am leaving this here as an PR looking for comments/input.


README.rst
```diff
@@ -399,7 +399,7 @@ Contributing
 
    .. code-block:: text
 
-      python3 -m pip install -r requirements.txt
+      python3 setup.py install
```
zariiii9003, line 402: `setup.py install` is deprecated.

setup.py
```diff
@@ -35,6 +35,10 @@ def find_version():
           'argparse_addons',
           'typing_extensions>=3.10.0.0',
           'crccheck',
+          'matplotlib',
+          'parameterized',
+          'codespell',
+          'nala',
```
zariiii9003, line 41: These are test dependencies, not runtime dependencies. They must not be included in `install_requires`

README.rst
wolframkriesing: where can i find the `requirements.txt`? On main I don't see one in the root dir, what am I missing?

setup.py
wolframkriesing: where do i put them, I guess the question is still where is the requirements.txt

README.rst
zariiii9003: The readme file is a bit outdated. We use `tox` to run the tests, so the test dependencies are declared in `tox.ini`

wolframkriesing: will adjust, thanks

PR #579, opened 2023-06-12
set name option for multiplexer signal

hyihua:
The function `_dump_message` uses multiplexer signal id as the signal name. However, it might be incompatible in some platforms, such as the screenshot shown.

image

In addition, using the ID as the name of the multiplexer signal is not clear enough for people to understand the meaning of this multiplexer. For this issue, I'd suggest that if the multiplexer signal has an original name, use the signal name rather than use the signal id.

### DIFF
--- a/cantools/database/can/formats/sym.py
+++ b/cantools/database/can/formats/sym.py
@@ -923,7 +923,7 @@ def _dump_message(message: Message, signals: List[Signal], min_frame_id: TypingO
         hex_multiplexer_id = format(multiplexer_id, 'x').upper()
         multiplexer_signal_name = multiplexer_signal.name
         if not multiplexer_signal_name:
-            multiplexer_signal_name = hex_multiplexer_id
+            raise ValueError(f"The name of the multiplexer signal with ID {str(hex_multiplexer_id)} is empty. The database is corrupt.")
         message_str += f'Mux="{multiplexer_signal_name}" {_convert_start(multiplexer_signal.start, multiplexer_signal.byte_order)},{multiplexer_signal.length} {hex_multiplexer_id}h {m_flag}\n'
     for signal in signals:
         message_str += f'Sig="{_get_signal_name(signal)}" {_convert_start(signal.start, signal.byte_order)}\n'
--- a/cantools/database/can/formats/sym.py
+++ b/cantools/database/can/formats/sym.py
@@ -921,7 +921,10 @@ def _dump_message(message: Message, signals: List[Signal], min_frame_id: TypingO
         if multiplexer_signal.byte_order == 'big_endian':
             m_flag = '-m'
         hex_multiplexer_id = format(multiplexer_id, 'x').upper()
-        message_str += f'Mux="{hex_multiplexer_id}" {_convert_start(multiplexer_signal.start, multiplexer_signal.byte_order)},{multiplexer_signal.length} {hex_multiplexer_id}h {m_flag}\n'
+        multiplexer_signal_name = hex_multiplexer_id
+        if multiplexer_signal.name != '':
+            multiplexer_signal_name = multiplexer_signal.name
+        message_str += f'Mux="{multiplexer_signal_name}" {_convert_start(multiplexer_signal.start, multiplexer_signal.byte_order)},{multiplexer_signal.length} {hex_multiplexer_id}h {m_flag}\n'
     for signal in signals:
         message_str += f'Sig="{_get_signal_name(signal)}" {_convert_start(signal.start, signal.byte_order)}\n'
     return message_str
--- a/cantools/database/can/formats/sym.py
+++ b/cantools/database/can/formats/sym.py
@@ -921,9 +921,9 @@ def _dump_message(message: Message, signals: List[Signal], min_frame_id: TypingO
         if multiplexer_signal.byte_order == 'big_endian':
             m_flag = '-m'
         hex_multiplexer_id = format(multiplexer_id, 'x').upper()
-        multiplexer_signal_name = hex_multiplexer_id
-        if multiplexer_signal.name != '':
-            multiplexer_signal_name = multiplexer_signal.name
+        multiplexer_signal_name = multiplexer_signal.name
+        if not multiplexer_signal_name:
+            multiplexer_signal_name = hex_multiplexer_id
         message_str += f'Mux="{multiplexer_signal_name}" {_convert_start(multiplexer_signal.start, multiplexer_signal.byte_order)},{multiplexer_signal.length} {hex_multiplexer_id}h {m_flag}\n'
     for signal in signals:
         message_str += f'Sig="{_get_signal_name(signal)}" {_convert_start(signal.start, signal.byte_order)}\n'
--- a/tests/files/sym/test_multiplex_dump.sym
+++ b/tests/files/sym/test_multiplex_dump.sym
@@ -3,6 +3,7 @@ Title="Untitled"
 
 {SIGNALS}
 Sig=MultiplexedSig unsigned 8 /max:1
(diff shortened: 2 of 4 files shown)


cantools/database/can/formats/sym.py
```diff
@@ -921,7 +921,10 @@ def _dump_message(message: Message, signals: List[Signal], min_frame_id: TypingO
         if multiplexer_signal.byte_order == 'big_endian':
             m_flag = '-m'
         hex_multiplexer_id = format(multiplexer_id, 'x').upper()
-        message_str += f'Mux="{hex_multiplexer_id}" {_convert_start(multiplexer_signal.start, multiplexer_signal.byte_order)},{multiplexer_signal.length} {hex_multiplexer_id}h {m_flag}\n'
+        multiplexer_signal_name = hex_multiplexer_id
+        if multiplexer_signal.name != '':
+            multiplexer_signal_name = multiplexer_signal.name
```
andlaus, line 926: flipping around the logic here is probably easier to read:
```suggestion
        multiplexer_signal_name = multiplexer_signal.name
        if not multiplexer_signal_name:
            multiplexer_signal_name = hex_multiplexer_id
```

cantools/database/can/formats/sym.py
```diff
@@ -921,7 +921,10 @@ def _dump_message(message: Message, signals: List[Signal], min_frame_id: TypingO
         if multiplexer_signal.byte_order == 'big_endian':
             m_flag = '-m'
         hex_multiplexer_id = format(multiplexer_id, 'x').upper()
-        message_str += f'Mux="{hex_multiplexer_id}" {_convert_start(multiplexer_signal.start, multiplexer_signal.byte_order)},{multiplexer_signal.length} {hex_multiplexer_id}h {m_flag}\n'
+        multiplexer_signal_name = multiplexer_signal.name
+        if not multiplexer_signal_name:
```
andlaus, line 925: hm, come think about it, I think that all signals always have to have names, so this `if` statement can probably be removed entirely. I just wonder why line 926 seems to be covered by the unit tests...

hyihua: That's right. When I create a symbol and want to set the name of the multiplexer empty, I cannot do that because it's invalid, so all signals always have to have names. However, if we rename the name through python code, like [here](https://github.com/cantools/cantools/pull/579/files#diff-82fbc96aa3f082d4667c09a2a35aec050c73120de9174e1f37bef26ef9cd3115R3415), it works and we can rename the multiplexer to be empty. In that case, the name of the multiplexer would be a empty string.

I think [this suggestion](https://github.com/cantools/cantools/pull/579#issuecomment-1587986038) is great, so I update the test case to cover this case. Refer to [the change](https://github.com/cantools/cantools/pull/579/files#diff-82fbc96aa3f082d4667c09a2a35aec050c73120de9174e1f37bef26ef9cd3115R3412) for function `test_multiplex_sym_dump`.

tests/test_database.py
```diff
@@ -3414,18 +3411,29 @@ def test_multiplex_dump(self):
 
     def test_multiplex_sym_dump(self):
         db = cantools.db.load_file('tests/files/sym/test_multiplex_dump.sym')
+        # change the name of the multiplexer signal to empty to trigger the condition in function _dump_message
+        db.messages[1].signals[0].name = ''
         dumped_db = cantools.db.load_string(db.as_sym_string())
-        dumped_msg = dumped_db.get_message_by_frame_id(0x100)
 
-        # Note: cantools database cannot support multiple multiplexer signal names, so SYM file names the multiplexer
-        # signal after the multiplexer id (Hence, 2A, not MultiplexerSig)
```
andlaus, line 3421: @samc1213: I don't get that comment. Do you remember what you meant to say here? (In my understanding, assigning more than one multiplexer selector signal to any multiplexed signal does not make sense anyway?)

hyihua: [The relevant pull request.](https://github.com/cantools/cantools/pull/431).

hyihua: @andlaus Are there any updates for this? Is there anything we can do to help approve this pull request?

cantools/database/can/formats/sym.py
andlaus: ok. I think that we should not support corrupt databases. IOW,  `multiplexer_signal.name` should be assumed to be a non-empty string. once this is cleaned up, this PR can be merged IMO...

tests/test_database.py
andlaus: I am just confused why this comment was made in the first place, but I think that it was spurious. (so everything is alright here as far as I can see...) I'll merge the PR once the special treatments for empty signal names (and the corresponding test) are removed...

cantools/database/can/formats/sym.py
hyihua: You're right. We should not support that. I would prefer the program raises an error to inform that.

tests/files/sym/test_multiplex_dump.sym
```diff
@@ -3,6 +3,7 @@ Title="Untitled"
 
 {SIGNALS}
 Sig=MultiplexedSig unsigned 8 /max:1
+Sig=NormalSig unsigned 8 /max:1
```
andlaus, line 6: I think that the changes to this file (and the ones in `test_database.py` where you check the `TestMultiplexer` message) are now redundant?

hyihua: Yes. I reverted the change on that sym file and updated the function `test_multiplex_sym_with_empty_signal_name_dump`.

tests/test_database.py
hyihua: The files are already updated. Please let me know if there are any other issues. We really hope that this feature can be merged and it's an important feature for our product implementation. Thanks @andlaus

andlaus:
once the CI is happy (seems like there is a typing mistake somewhere), I'll merge this.

hyihua:
> once the CI is happy (seems like there is a typing mistake somewhere), I'll merge this.
But it seems like the error was not introduced by my changes. 
image

andlaus:
hm, this is what the github action logs say:
Screenshot_20230623_214042
@zariiii9003: Any idea what's going on here and how to fix it?

andlaus:
okay: the linter is not run on the tests, so
```
ruff check --fix ./cantools
```
fixes the issue. (it is unrelated to this PR...)

hyihua:
> okay: the linter is not run on the tests, so
> 
> ```
> ruff check --fix ./cantools
> ```
> 
> fixes the issue. (it is unrelated to this PR...)

I run this command locally before, but just feel like it's not related to this pull request and then I should push the changes to this pull request. But anyway, we should fix the issue. Made a new commit for that.

andlaus:
I think it would have been slightly better if you put the fixes `ruff` made to the tests into a separate PR, but since the CI is happy and IMO they are all janitorial changes which we would want to have at some point anyway, I'll merge. thanks for the contribution!

hyihua:
> I think it would have been slightly better if you put the fixes `ruff` made to the tests into a separate PR, but since the CI is happy and IMO they are all janitorial changes which we would want to have at some point anyway, I'll merge. thanks for the contribution!

That's a great suggestion. Thanks for that.

hyihua:
@andlaus, may I ask about when you will release a new version of `cantools`?

tests/test_database.py
samc1213: I'm sorry for the late reply. I can't recall why I thought SYM multiplexer and cantools database representations didn't match, and decided to use the multiplexer value as the signal name. I thought during parsing we would lose the name of each multiplexer, so we had to use the value or something. I tried to reproduce but couldn't.

Issue #584, opened 2023-06-27
39.0.0rc1: cannot import `cantools` due to circular reference


mon:
Not quite sure what change led to this, but it works fine on 38. Installed rc1 via pip.

```
Python 3.11.0 (main, Oct 24 2022, 18:26:48) [MSC v.1933 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import cantools
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\mon\AppData\Local\Programs\Python\Python311\Lib\site-packages\cantools\__init__.py", line 13, in <module>
    from . import database as db  # isort: skip
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ImportError: cannot import name 'database' from partially initialized module 'cantools' (most likely due to a circular import) (C:\Users\mon\AppData\Local\Programs\Python\Python311\Lib\site-packages\cantools\__init__.py)
```

andlaus:
very weird. does it work with the git version? (i.e., maybe the issue is caused by the pypi package...)

BTW: we should mention your linter in the README before the final release. can you open a PR to that end?

mon:
Indeed, installing the git version (via `pip install -U .`) works fine.
```
>>> import cantools
>>> cantools.__version__
'39.0.0-rc1'
```

> we should mention your linter in the README before the final release

I'm not quite sure where's the best spot to link it! Maybe nearby the `Other C code generators:` section?

mon:
Actually, I spoke too soon, since I ran my test terminal in the cantools git repo.... It doesn't work, but a different error!

Site packages dir shortened to `....` for brevity

```
Python 3.11.0 (main, Oct 24 2022, 18:26:48) [MSC v.1933 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import cantools
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "....\cantools\__init__.py", line 13, in <module>
    from . import database as db  # isort: skip
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "....\cantools\database\__init__.py", line 8, in <module>
    from . import can, diagnostics, utils
  File "....\cantools\database\can\__init__.py", line 2, in <module>
    from .database import Database
  File "....\cantools\database\can\database.py", line 23, in <module>
    from .formats import arxml, dbc, kcd, sym
[...]

PR #587, opened 2023-06-27
Change to src repo structure

zariiii9003:
With this change, the tests will run on the packaged and installed code. Look at the paths in the coverage summary:

image

It looks like this on the master branch:
image

This should avoid bugs like in #584


tox.ini
```diff
@@ -1,8 +1,8 @@
 [tox]
 envlist = py
-isolated_build = true
```
andlaus, line 3: what prevents using isolated builds? (what does the `isolated_build` parameter do in the first place?)

andlaus (commented):
in principle, this is okay IMO. That said, be aware that I'd like to significantly flatten the source tree soonish, so it might be worthwhile to do such changes in one go. (maybe not...)

tox.ini
zariiii9003: isolated_build was removed in tox4, the package is always built in a separate venv

zariiii9003:
> in principle, this is okay IMO. That said, be aware that I'd like to significantly flatten the source tree soonish, so it might be worthwhile to do such changes in one go. (maybe not...)

I'd prefer not to change too much in a single PR.

andlaus:
> I'd prefer not to change too much in a single PR.

alright. let's do it piecemeal then...

andlaus:
@zariiii9003: anything which needs to be changed here or can it be merged? (I would like to not keep this open for too long because this has quite a bit of headache potential.)

zariiii9003:
> @zariiii9003: anything which needs to be changed here or can it be merged? (I would like to not keep this open for too long because this has quite a bit of headache potential.)

Done :slightly_smiling_face:

PR #589, opened 2023-06-28
Refactor code for  node mapped attributes.

malneni:
Currently, the node mapped attributes are stored within  the database-specific dbc_specifics. This refactoring proposes moving them to the more appropriate location within the message/signal dbc_specifics. It will be better when adding or removing message/signal, corresponding rel attributes will be easier to handle


cantools/database/can/formats/dbc.py
```diff
@@ -854,28 +854,26 @@ def get_value(attribute):
...
-                                          f'{signal_name} '
-                                          f'{get_value(attribute)};')
-            elif "node" in element:
-                for node_name, node_dict in element['node'].items():
-                    for attribute in node_dict.values():
+    for message in database.messages:
+        for signal in message.signals:
+            if signal.dbc is not None and signal.dbc.attributes_rel is not None:
+                for node_name, node_dict in signal.dbc.attributes_rel.items():
+                    for _, attribute in node_dict.items():
```
andlaus, line 861: if you don't use the keys anyways, you can use this:
```suggestion
                    for attribute in node_dict.values():
```

cantools/database/can/formats/dbc.py
```diff
@@ -854,28 +854,26 @@ def get_value(attribute):
...
+                                      f'BU_SG_REL_ '
                                       f'{node_name} '
-                                      f'{frame_id} '
+                                      f'SG_ '
+                                      f'{message.frame_id} '
+                                      f'{signal.name} '
                                       f'{get_value(attribute)};')
+        if message.dbc is not None and message.dbc.attributes_rel is not None:
+            for node_name, node_dict in message.dbc.attributes_rel.items():
+                for _, attribute in node_dict.items():
```
andlaus, line 871: same here...

andlaus (commented):
I like this in principle: the ARXML loader already takes the same approach via `signal.autosar` and `database.autosar`, etc. That said, I'm not really in a position to judge the merrits of DBC-specific stuff. Maybe @juleq can help out?

PR #592, opened 2023-07-03
Make cycle_time serialisation more consistent for DBC files

### LINKED ISSUE #540: Does GenMsgCycleTime have to be added when serialising to .dbc?
I have some .dbc files that lack this attribute and seem to work fine (in cantools, CANdb++, Vehicle Spy, SavvyCAN etc). I'd like to avoid having to post-process the generated .dbc to remove it.

Is there a technical reason it's auto-generated, or would it be safe to open a PR removing it?
It seems to have been added to fix a case where it should have been there but wasn't, but I wonder if more precise logic can be included.

### LINKED ISSUE #588: Messages using default GenMsgCycleTime get explicitly serialised
I'm not sure if this was caused by the `Decimal` removal or something else - my DBC specifies a default `GenMsgCycleTime` of 100, but any message using this value now gets an explicit `BA_` entry in the saved dbc.

mon:
- Moves default fallback to the getter instead of the initial load from DBC
- Only creates the `GenMsgCycleTime` attribute if it's missing _and_ needed
- Doesn't serialise default values if they haven't been changed

It _feels_ like the current test suite is sufficient to validate this change, but feel free to request some more.

Closes #588, closes #540.

### DIFF
--- a/cantools/database/can/formats/dbc.py
+++ b/cantools/database/can/formats/dbc.py
@@ -587,6 +587,12 @@ def _need_startval_def(database):
                for m in database.messages
                for s in m.signals)
 
+def _need_cycletime_def(database):
+    # If the user has added cycle times to a database which didn't start with them,
+    # we need to add the global attribute definition so the output DBC is valid
+    return any(m.cycle_time is not None
+               for m in database.messages)
+
 def _dump_attribute_definitions(database):
     ba_def = []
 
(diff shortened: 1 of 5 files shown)


mon:
^ Ah, bummer. I don't know which 2 lines have lost coverage, the coveralls website just says `SOURCE NOT AVAILABLE`.

cantools/database/can/formats/dbc.py
```diff
@@ -587,6 +587,10 @@ def _need_startval_def(database):
                for m in database.messages
                for s in m.signals)
 
+def _need_cycletime_def(database):
+    return any(m.cycle_time is not None
```
zariiii9003, line 591: Could you add an explanatory comment here? Otherwise this looks good to me.

mon: Oh, I figured it was similar enough to the function above it (`_need_startval_def`) to not need one either!

How about
```python
# If the user has added cycle times to a database which didn't start with them,
# we need to add the global attribute definition so the output DBC is valid
```

zariiii9003: Yes, that would be fine. Sorry for the late response

mon:
I note that the linter is now failing, but `master` has the same issues (they all seem to be "using str.format instead of f-strings"). Hopefully it's fine to merge as-is.

zariiii9003 (approved):

PR #599, opened 2023-08-03
User f-strings instead of str.format()

zariiii9003:
Make ruff happy


andlaus:
nice! this has also been on my TODO list for a while :)

PR #600, opened 2023-08-07
Add prog option to argparse for help messages

jack-champagne:
When running cantools, depending on how it is launched, the help message will give different program names after the word `usage:` which can be set to a distinct name. Instead of deferring this to the default (os.path.basename(sys.argv[0])) in argparse, specify program name explicitly in argparse options.

## Before:
```
$ python -m cantools --help
usage: __main__.py [-h] [-d] [--version] {decode,monitor,plot,convert,dump,generate_c_source,list} ...

Various CAN utilities.
...
```

(pip installed)
```
$ cantools --help
usage: cantools [-h] [-d] [--version]
                {decode,monitor,dump,convert,generate_c_source} ...

Various CAN utilities.
...
```

## After:
```
$ python -m cantools --help
usage: cantools [-h] [-d] [--version] {decode,monitor,plot,convert,dump,generate_c_source,list} ...

Various CAN utilities.
...
```

(pip installed)
```
$ cantools --help
usage: cantools [-h] [-d] [--version] {decode,monitor,plot,convert,dump,generate_c_source,list} ...

Various CAN utilities.
...
```

### DIFF
--- a/cantools/__init__.py
+++ b/cantools/__init__.py
@@ -49,6 +49,7 @@ def _load_subparser(subparser_name, subparsers):
 
 def _main():
     parser = argparse.ArgumentParser(
+        prog='cantools',
         description='Various CAN utilities.',
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )


jack-champagne:
This is a good point, maybe instead of my change, just a check to avoid it showing __main__.py when launched this way. I see the benefit of it showing the exec name

andlaus:
I guess it would be a good solution. That said, I think that the current patch is acceptable if @zariiii9003 and @juleq do not object...

zariiii9003:
> I guess it would be a good solution. That said, I think that the current patch is acceptable if @zariiii9003 and @juleq do not object...

I do not object, looks okay to me.

andlaus:
alright, let's go for it. Merging, thanks @jack-champagne

PR #601, opened 2023-08-09
Make ruff happy: Use format strings and isinstance where possible

jack-champagne:
Make ruff happier by making suggested changes and using f-strings and isinstance where possible. Get the pipeline passing again.


andlaus:
I think the patch is good, but it will lead to conflicts with @zariiii9003's #587 (which also fixes this AFAICS). I think it is better to merge #587 if there are no showstoppers to it. @zariiii9003 ?

Issue #606, opened 2023-09-13
Screen flicker with  python -m cantools monitor -c can1 BigFile.arxml


robertrau:
First, thank you for all the work put in to cantools, it have served me well.
I am a hardware engineer writing a tester in python.
Running on a Raspberry Pi 3 Model B Rev 1.2
with a MCP2518 based HAT board
Using: Python 3.9.2
with a large (7.2MiB) arxml file
CAN bus: 500kbps, < 10% bandwidth
There is a lot of screen flicker that makes the screen hard to read.

Is there a terminal setting for curses that can reduce the flicker, or anything I can to do make the screen more readable?

Thank you.

robertrau:
My message loop is 1.1 seconds, so all messages have been received several times, See video.
Also, one of my CAN bus devices is getting power cycles, so there are some unacknowledged CAN messages. Is this a problem?

https://github.com/cantools/cantools/assets/18400100/0aac6010-de9f-46af-a770-83fee00b6540

andlaus:
I don't think that this is a problem related to anything that happens on the bus, but it's probably by how the `monitor` tool updates the screen. Unfortunately I cannot debug this myself because the issue does not crop up on any of my machines...

robertrau:
I found a fix from here:
https://stackoverflow.com/questions/24964940/python-curses-tty-screen-blink
I changed line 94 of monitor.py from clear to erase
and the screen is now flicker free.
Tested scrolling, paging up and down, messing with filters and the screen
stays up to date with no garbage left anywhere.

andlaus:
great! can you make a pull request?

robertrau:
OK, happy to, but full disclosure. I am a old hardware engineer and I have only used github for my own stuff (just pull, commit, and push). So I read the github tutorial at 
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request#creating-the-pull-request
but got lost at step 2 where I should see my commits in a branch.
This is what I have done so far:
Created a work folder
Cloned the repo in the folder
edited the file with the change.

Please guide me through the next steps (sorry for needing this much hand holding).

andlaus:
okay, let's get this done. your next steps:

[...]

PR #609, opened 2023-09-24
Solved screen flicker problem

robertrau:
changed monitor.py to use curses erases instead of clear. Clear does a wholesale clear of everything on the screen where erase will only update the characters on the screen that need changing.
This addresses issue #606

### DIFF
--- a/src/cantools/subparsers/monitor.py
+++ b/src/cantools/subparsers/monitor.py
@@ -91,7 +91,7 @@ class Monitor(can.Listener):
 
     def redraw(self):
         # Clear the screen.
-        self._stdscr.clear()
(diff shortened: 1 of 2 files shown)


andlaus:
nice catch, thanks. unfortunately the unit tests currently don't pass with this change applied. (probably that's because the test for the `monitor` tool currently does not expect `erase()` to be called on `StdScr` objects.)

(to run the tests locally, use `pytest .`, then fix the issue, add make a new commit for the fix and push the new version of the branch to github: `git push personal master`)

zariiii9003:
The linter errors are unrelated, i'll fix those in a separate PR.

PR #619, opened 2023-10-28
Refactor dbc GenMsgCycleTime handling

### LINKED ISSUE #610: The value for "message.cycle_time" behaves differently depending on the cantools version.
In version 38.0.2, the cycle_time is output as defined in the DBC, but in version 39.2.0, messages without a cycle_time are output with a value other than None.

```
db = cantools.database.load_file('XXX.dbc')
for message in db.messages:
    # Tx message check
    if 'XXX' in message.senders:
        framesTx[message.name] = message
        txFrame.append(message.name)
        print(message.name, message.cycle_time)
```

In 38.0.2 and 39.0.0
```
XXX_200ms 200
XXX_200ms 200
XXX_FD None
XXX_GST None
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_200ms 200
XXX_200ms 200
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_00ms None
XXX_FD 100
```

In 39.2.0
```
XXX_200ms 200
XXX_200ms 200
XXX_FD 200
XXX_GST 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_200ms 200
XXX_200ms 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_00ms 200
XXX_FD 100
```

### LINKED ISSUE #618: Example in the repo not working as expected
import os
import cantools
import pandas
import openpyxl

#Get the current directory
current_dir = os.getcwd()

SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))

INPUT_DBC_PATH = os.path.join(SCRIPT_DIR, 'XXXXXXXXX.dbc')
OUTPUT_DBC_PATH = os.path.join(SCRIPT_DIR, 'XXXXXXXXXX_new.dbc')

print("Loading DBC database from '{}'.".format(INPUT_DBC_PATH))
db = cantools.db.load_file(INPUT_DBC_PATH)
message = db.get_message_by_name('XXXXXXX')
try:
    print("Input cycle time: ",message.dbc.attributes['GenMsgCycleTime'].value)
    message.dbc.attributes['GenMsgCycleTime'].value = 2000
    print("Output cycle time:",message.dbc.attributes['GenMsgCycleTime'].value)
except KeyError as e:
    raise e
print("Input frame id: ", hex(message.frame_id))
message.frame_id = 0x19FF70EC
print("Output frame id:", hex(message.frame_id))
print("Writing modified DBC database to '{}'.".format(OUTPUT_DBC_PATH))
with open(OUTPUT_DBC_PATH, 'w', newline="\r\n") as fout:
    fout.write(db.as_dbc_string())

image

When I tried to run the above file, the GenMsgCycleTime is not getting affected instead it remains the same. Do you know as why this can happen?

zariiii9003:
Closes #618 
Closes #610 (hopefully)

### DIFF
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -822,7 +822,7 @@ def _dump_attributes(database, sort_signals, sort_attributes):
         if gen_msg_cycle_time_def := database.dbc.attribute_definitions.get("GenMsgCycleTime"):
             if msg_cycle_time != gen_msg_cycle_time_def.default_value:
                 msg_attributes['GenMsgCycleTime'] = Attribute(
-                    value=message.cycle_time,
+                    value=msg_cycle_time,
                     definition=gen_msg_cycle_time_def,
                 )
             elif 'GenMsgCycleTime' in msg_attributes:
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -169,7 +169,7 @@ ATTRIBUTE_DEFINITION_GENMSGCYCLETIME = AttributeDefinition(
     minimum=0,
     maximum=2**16-1)
 
-ATTRIBUTE_DEFININTION_GENSIGSTARTVALUE = AttributeDefinition(
+ATTRIBUTE_DEFINITION_GENSIGSTARTVALUE = AttributeDefinition(
     name='GenSigStartValue',
     default_value=0,
     kind='SG_',
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -641,7 +641,7 @@ def _dump_attribute_definitions(database: InternalDatabase) -> List[str]:
     if 'GenMsgCycleTime' not in definitions and _need_cycletime_def(database):
         definitions['GenMsgCycleTime'] = ATTRIBUTE_DEFINITION_GENMSGCYCLETIME
     if 'GenSigStartValue' not in definitions and _need_startval_def(database):
-        definitions['GenSigStartValue'] = ATTRIBUTE_DEFININTION_GENSIGSTARTVALUE
+        definitions['GenSigStartValue'] = ATTRIBUTE_DEFINITION_GENSIGSTARTVALUE
 
     # create 'VFrameFormat' and 'CANFD_BRS' attribute definitions if bus is CAN FD
     if _bus_is_canfd(database):
@@ -876,7 +876,7 @@ def _dump_attributes(database, sort_signals, sort_attributes):
             elif signal.raw_initial is not None:
                 sig_attributes['GenSigStartValue'] = Attribute(
                     value=signal.raw_initial,
-                    definition=ATTRIBUTE_DEFININTION_GENSIGSTARTVALUE)
+                    definition=ATTRIBUTE_DEFINITION_GENSIGSTARTVALUE)
 
             # output all signal attributes
             for attribute in sig_attributes.values():
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -161,6 +161,23 @@ ATTRIBUTE_DEFINITION_BUS_TYPE = AttributeDefinition(
     default_value='CAN',
     type_name='STRING')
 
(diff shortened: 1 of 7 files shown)


src/cantools/database/can/formats/dbc.py
```diff
@@ -823,12 +817,18 @@ def get_value(attribute):
 
         # synchronize the attribute for the message cycle time with
         # the cycle time specified by the message object
-        if message._cycle_time is None and 'GenMsgCycleTime' in msg_attributes:
+        gen_msg_cycle_time_def: AttributeDefinition
+        msg_cycle_time = message.cycle_time or 0
+        if gen_msg_cycle_time_def := database.dbc.attribute_definitions.get("GenMsgCycleTime"):
+            if msg_cycle_time != gen_msg_cycle_time_def.default_value:
+                msg_attributes['GenMsgCycleTime'] = Attribute(
+                    value=message.cycle_time,
```
andlaus, line 825:
```suggestion
                    value=msg_cycle_time,
```

(probably)

andlaus (commented):
the patch has a few minor nits, but besides this, it looks good. I particularly like the fact that some of the special treatment for DBC files is removed from the `Message` class.

andlaus:
@zariiii9003: how to proceed with the CI errors? throw out pypy-3.9?

src/cantools/database/can/formats/dbc.py
zariiii9003: Yes, that was incorrect, thanks 👍

zariiii9003:
> @zariiii9003: how to proceed with the CI errors? throw out pypy-3.9?

That was ma fault, should be fixed soon: https://github.com/eerimoq/bitstruct/pull/37

src/cantools/database/can/formats/dbc.py
```diff
@@ -161,6 +161,23 @@
...
 
+ATTRIBUTE_DEFINITION_GENMSGCYCLETIME = AttributeDefinition(
+    name='GenMsgCycleTime',
+    default_value=0,
+    kind='BO_',
+    type_name='INT',
+    minimum=0,
+    maximum=2**16-1)
+
+ATTRIBUTE_DEFININTION_GENSIGSTARTVALUE = AttributeDefinition(
```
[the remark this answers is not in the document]
zariiii9003: Oh, i have to rename the others too...

andlaus:
okay, let's get over this hump. thanks, @zariiii9003!

mon:
Ah shoot, I didn't see this before it got merged. This actually creates a subtle behaviour change in how cycle times are serialised, and it's no longer possible to tell if a cycle time was intentionally set to the default value for a given message.

e.g. if a default cycle time of `100` is used, and I've set Msg_A's cycle time to `100`, this will be lost upon re-save.
I believe someone else mentioned needing this behaviour, I'd have to find the comment.

I really would have liked an example .dbc file in #610 to try and understand why `cycle_time` was returning `None`, as it shouldn't have done so. I don't think this fixes the right thing.

mon:
Also just to be clear, I don't personally need that behaviour, but it differs from CANdb++ so that's why I'm being hard about it.

zariiii9003:
> Ah shoot, I didn't see this before it got merged. This actually creates a subtle behaviour change in how cycle times are serialised, and it's no longer possible to tell if a cycle time was intentionally set to the default value for a given message.
> 
> e.g. if a default cycle time of `100` is used, and I've set Msg_A's cycle time to `100`, this will be lost upon re-save.
> I believe someone else mentioned needing this behaviour, I'd have to find the comment.
> 

That's true, but if that behavior is required, there should have been a test for it. To me it is acceptable to omit a value, if it is equal to the default value. Maybe I'll take a look at it later, should be easy to fix.

mon:
> if that behavior is required, there should have been a test for it
Indeed, I think this is ultimately on me for not adding a test when I made the original change 😅

PR #622, opened 2023-11-15
Container improvements

andlaus:
This pull request mainly contains a few improvements for working with container frames:

- The output of the `decode` utility is now quite a bit nicer (IMO)
- When encoding container frames and specifying the binary payload of child frames directly, the data do no longer needs to exhibit to the size specified by the message description (instead of raising an exception, a log message will be produced)
- by default, decoding frames now fails if the payload to be decoded is longer than expected. To decode such data, passing `allow_excess=True` to the respective decode function is now required
- besides ignoring some inconsistencies in the database, passing `--no-strict` to `cantools decode` now tries to decode incorrectly-sized frames on a best effort basis. (if the received data was too short, the decoded signal values might be incorrect, if it was too long, not all data can be decoded.)
- decoding container frames that contain incorrectly-sized child messages will no longer make decoding the whole container message fail: Instead of the decoded child message, the binary payload will be put into the result.

Besides this, this PR also corrects the URIs for the cantools homepage and the repository in `pyproject.toml`.

### DIFF
--- a/src/cantools/database/can/message.py
+++ b/src/cantools/database/can/message.py
@@ -1173,7 +1173,7 @@ class Message:
                                                    decode_containers=False,
                                                    allow_truncated=allow_truncated,
                                                    allow_excess=allow_excess)
-            except ValueError:
+            except (ValueError, DecodeError):
                 result.append((contained_message, bytes(contained_data)))
                 continue
 
--- a/src/cantools/database/can/message.py
+++ b/src/cantools/database/can/message.py
@@ -851,15 +851,16 @@ class Message:
             if isinstance(value, bytes):
                 # raw data
 
-                # ensure that the size of the blob corresponds to the
-                # one specified by the featured message.
+                # produce a message if size of the blob does not
+                # correspond to the size specified by the message
+                # which it represents.
                 if contained_message is not None and \
-                   len(value) != contained_message.length:
+                    len(value) != contained_message.length:
 
-                    raise EncodeError(f'Specified data for contained message '
-                                      f'{contained_message.name} is '
-                                      f'{len(value)} bytes instead of '
-                                      f'{contained_message.length} bytes')
+                    LOGGER.info(f'Specified data for contained message '
+                                f'{contained_message.name} is '
+                                f'{len(value)} bytes instead of '
+                                f'{contained_message.length} bytes')
 
                 contained_payload = value
 
--- a/src/cantools/database/can/message.py
+++ b/src/cantools/database/can/message.py
@@ -960,14 +960,16 @@ class Message:
                 data: bytes,
                 decode_choices: bool,
                 scaling: bool,
-                allow_truncated: bool) -> SignalDictType:
+                allow_truncated: bool,
+                allow_excess: bool) -> SignalDictType:
         decoded = decode_data(data,
                               self.length,
                               node['signals'],
                               node['formats'],
                               decode_choices,
                               scaling,
-                              allow_truncated)
+                              allow_truncated,
+                              allow_excess)
 
         multiplexers = node['multiplexers']
 
@@ -988,7 +990,8 @@ class Message:
                                         data,
                                         decode_choices,
                                         scaling,
-                                        allow_truncated))
+                                        allow_truncated,
+                                        allow_excess))
 
         return decoded
 
@@ -1060,7 +1063,8 @@ class Message:
                decode_choices: bool = True,
                scaling: bool = True,
                decode_containers: bool = False,
-               allow_truncated: bool = False
+               allow_truncated: bool = False,
+               allow_excess: bool = True,
                ) \
                -> DecodeResultType:
         """Decode given data as a message of this type.
@@ -1091,24 +1095,30 @@ class Message:
         ``False``, `DecodeError` will be raised when trying to decode
         incomplete messages.
 
+        If `allow_excess` is ``True``, data that is are longer than
+        the expected message length is decoded, else a `ValueError` is
+        raised if such data is encountered.
         """
 
         if decode_containers and self.is_container:
             return self.decode_container(data,
                                          decode_choices,
                                          scaling,
-                                         allow_truncated)
+                                         allow_truncated,
+                                         allow_excess)
 
         return self.decode_simple(data,
                                   decode_choices,
                                   scaling,
-                                  allow_truncated)
+                                  allow_truncated,
+                                  allow_excess)
 
     def decode_simple(self,
                       data: bytes,
                       decode_choices: bool = True,
                       scaling: bool = True,
-                      allow_truncated: bool = False) \
+                      allow_truncated: bool = False,
+                      allow_excess: bool = True) \
                       -> SignalDictType:
         """Decode given data as a container message.
 
@@ -1122,19 +1132,19 @@ class Message:
         elif self._codecs is None:
             raise ValueError('Codec is not initialized.')
 
-        data = bytes(data[:self._length])
-
         return self._decode(self._codecs,
                             data,
                             decode_choices,
                             scaling,
-                            allow_truncated)
+                            allow_truncated,
+                            allow_excess)
 
     def decode_container(self,
                          data: bytes,
                          decode_choices: bool = True,
                          scaling: bool = True,
-                         allow_truncated: bool = False) \
+                         allow_truncated: bool = False,
+                         allow_excess: bool = True) \
                          -> ContainerDecodeResultType:
         """Decode given data as a container message.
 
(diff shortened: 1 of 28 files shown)


tests/test_autosar.py
```diff
@@ -163,6 +162,7 @@ def test_autosar4_secoc(self):
 
         self.assertEqual(encoded, bytes.fromhex('000000003130'))
 
+        print(f"{encoded}")
```
zariiii9003, line 165: Is this intentional?

tests/test_plot.py
```diff
@@ -822,7 +821,7 @@ def test_no_decode_choices(self):
 """
 
         db = cantools.db.load_file(self.DBC_FILE_CHOICES)
-        choices = db.get_message_by_name("Foo").get_signal_by_name("Foo").choices
+        db.get_message_by_name("Foo").get_signal_by_name("Foo").choices
```
zariiii9003, line 824: This line can probably be removed.

tests/test_tester.py
```diff
@@ -1,14 +1,12 @@
...
 import can
 
 try:
-    from queue import Queue
-    from queue import Empty
+    from queue import Empty, Queue
 except ImportError:
-    from Queue import Queue
-    from Queue import Empty
+    from Queue import Empty, Queue
```
zariiii9003, line 9: This was necessary for python 2 and is obsolete now

tests/test_monitor.py
```diff
@@ -1,30 +1,27 @@
...
 try:
-    from unittest.mock import Mock
-    from unittest.mock import patch
-    from unittest.mock import call
+    from unittest.mock import Mock, call, patch
 except ImportError:
-    from mock import Mock
-    from mock import patch
-    from mock import call
+    from unittest.mock import Mock, call, patch
```
zariiii9003, line 16: This can be removed.

tests/test_list.py
```diff
@@ -1,24 +1,18 @@
...
 try:
-    from unittest.mock import Mock
-    from unittest.mock import patch
-    from unittest.mock import call
+    from unittest.mock import Mock, call, patch
 except ImportError:
-    from mock import Mock
-    from mock import patch
-    from mock import call
+    from unittest.mock import patch
```
zariiii9003, line 6: This can be removed.

tests/test_command_line.py
```diff
@@ -1,16 +1,16 @@
...
 import types
-import functools
+import unittest
 from pathlib import Path
 
 try:
     from unittest.mock import patch
 except ImportError:
-    from mock import patch
+    from unittest.mock import patch
```
zariiii9003, line 13: This can be removed.

src/cantools/database/utils.py
```diff
@@ -123,15 +123,31 @@ def decode_data(data: bytes,
...
                 scaling: bool,
                 allow_truncated: bool,
+                allow_excess: bool,
                 ) -> SignalDictType:
 
     actual_length = len(data)
-    if allow_truncated and actual_length < expected_length:
-        data = data.ljust(expected_length, b"\xFF")
+
+    if actual_length < expected_length:
```
zariiii9003, line 131: Would you mind nesting both length checks to optimize for the happy path?

```python3
if len(data) != expected_length:
    if allow_truncated:
        data = data.ljust(expected_length, b"\xFF")
    if allow_excess:
        data = data[:expected_length]
    if len(data) != expected_length:
        raise ValueError
```

src/cantools/database/can/message.py
```diff
@@ -851,15 +851,16 @@ def _encode_container(self,
...
+                # which it represents.
                 if contained_message is not None and \
-                   len(value) != contained_message.length:
+                    len(value) != contained_message.length:
 
-                    raise EncodeError(f'Specified data for contained message '
-                                      f'{contained_message.name} is '
-                                      f'{len(value)} bytes instead of '
-                                      f'{contained_message.length} bytes')
+                    LOGGER.info(f'Specified data for contained message '
```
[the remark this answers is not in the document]
andlaus: not sure if this is really applicable here: this branch should be taken relatively rarely and evaluating the f-string should be at least as fast as the call to `LOGGER.info()` itself?

src/cantools/database/utils.py
```diff
@@ -123,15 +123,31 @@ def decode_data(data: bytes,
...
+            raise ValueError(f"Wrong data size: {actual_length} instead of "
+                             f"{expected_length} bytes")
+        else:
+            # trim the payload data to match the expected size
+            data = data[:expected_length]
 
     unpacked = {
-        **formats.big_endian.unpack(data),
-        **formats.little_endian.unpack(data[::-1]),
+        **formats.big_endian.unpack(bytes(data)),
```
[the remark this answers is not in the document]
andlaus: because for whatever reason `data` now sometimes is a `bytearray` here and thus test_autosar.py fails
```
> python3 -m pytest -x --tb=native tests/test_autosar.py
[...]
  File "src/cantools/database/utils.py", line 149, in decode_data
    **formats.big_endian.unpack(data),
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected bytes, bytearray found
---------------------------------------------
```

I have not invested too much time in finding out why that is, or why `test_autosar` is so picky. (I suspect the latter to be `bitstruct`'s fault.) Stay tuned...

src/cantools/database/utils.py
```diff
@@ -123,15 +123,31 @@ def decode_data(data: bytes,
...
+        **formats.little_endian.unpack(bytes(data[::-1])),
     }
 
     if allow_truncated and not (scaling or decode_choices):
```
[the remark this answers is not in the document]
andlaus: good point. fixed: 951c1ee31f9349def058bf8e4f1ccc712e28121b

src/cantools/database/utils.py
andlaus: done: 19207387d6a97779d9bff4ac9cd6a6d84489db66 (I've added an outer `if` to get more comprehensive error codes, though...)

tests/test_tester.py
andlaus: the change itself was also ruff's fault :). fixed :)

tests/test_plot.py
andlaus: good point. Note that I did not touch that file manually. Instead this change was made by `ruff --fix .`...

tests/test_autosar.py
andlaus: oops, no. Thanks!

src/cantools/database/utils.py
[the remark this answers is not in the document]
andlaus: okay, the culprit is `bitstruct.c`: it does not accept `bytearray` objects for `CompiledFormatDict.unpack`. (`bitstruct` works fine...)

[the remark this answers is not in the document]
zariiii9003: Which version of bitstruct? It should work https://github.com/eerimoq/bitstruct/pull/31/files 🤔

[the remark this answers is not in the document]
andlaus: `8.17.0`. (Stupid me; I had a clone of the repo pointing at this old version lying around in my PYTHONPATH.) Updating fixed it! Nice, thanks for the fix!

src/cantools/database/can/message.py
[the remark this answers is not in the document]
zariiii9003: Yes it's okay.

src/cantools/database/utils.py
zariiii9003: But the error message is the same for both cases? What's the difference? 🤔

src/cantools/database/utils.py
```diff
@@ -123,21 +123,36 @@ def decode_data(data: bytes,
...
+            if not allow_truncated:
+                 raise ValueError(f"Wrong data size: {actual_length} instead of "
+                                  f"{expected_length} bytes")
+            else:
+                # pad the data with 0xff to prevent the codec from
+                # raising an exception. Note that all signals
+                # that contain garbage will be removed below.
+                data = data.ljust(expected_length, b"\xFF")
+
+        if actual_length > expected_length:
```
zariiii9003, line 142:
```suggestion
        else:
```

src/cantools/database/utils.py
andlaus: good point. changed: bb7055d7f4171d2aa57128b2761b2c4f67dd077d

zariiii9003 (approved):
Thanks! 😊

andlaus:
I have to thank for the review :)

let's merge this...

PR #623, opened 2023-11-15
Remove compat.py and version.py

zariiii9003:
- remove compat.py
- remove version.py
- update test dependencies
- test on CPython 3.12
- fix FutureWarnings and DeprecationWarnings


src/cantools/__init__.py
```diff
@@ -7,12 +7,12 @@
 
 from . import j1939, logreader, tester
 from .errors import Error
-from .version import __version__
 
 # Remove once less users are using the old package structure.
 from . import database as db  # isort: skip
 
 __author__ = 'Erik Moqvist'
+__version__ = '39.3.0'
```
andlaus, line 15: putting this here will not a smaller a maintainance burden than having `version.py`. maybe we can use something like [setuptools_scm](https://github.com/pypa/setuptools_scm)?

andlaus (approved):
looks good. I'd really like to do a release by just creating a tag on github, though...

src/cantools/__init__.py
zariiii9003: The maintenance is the same as before. But the `version.py` file is just a leftover from the time, when `__version__` was parsed inside `setup.py`. With `pyproject.toml` it serves no purpose.

Regarding setuptools_scm, i have no clue what that does 😄

andlaus: it basically determines at the latest tagged version in the git history and based on that creates a `version.py` file when making a pypi package or running `pip install -e .`. the only drawback is that it requires the build system to be based on setuptools (I think), but as far as I know, this is the case for cantools. (right?)

andlaus: (I can make a patch if this sounds interesting to you...)

zariiii9003: > (I can make a patch if this sounds interesting to you...)

Sure, i don't mind. 

> it basically determines at the latest tagged version in the git history and based on that creates a `version.py` file when making a pypi package or running `pip install -e .`. 

The file is optional though, right? I think the `__version__` value can be set dynamically with `importlib.metadata`

>the only drawback is that it requires the build system to be based on setuptools (I think), but as far as I know, this is the case for cantools. (right?)

Yes, it's setuptools.

andlaus: > The file is optional though, right? I think the __version__ value can be set dynamically with importlib.metadata

not sure, but the setuptools_scm documentation certainly makes the impression that the file is optional:

```
To enable version inference, you need to set the version dynamically in the project section of pyproject.toml:

[project]
# version = "0.0.1"  # Remove any existing version parameter.
dynamic = ["version"]

[tool.setuptools_scm]

Additionally, a version file can be written by specifying:

[tool.setuptools_scm]
version_file = "pkg/_version.py"
```

andlaus: okay, I made a PR for your PR: https://github.com/zariiii9003/cantools/pull/1

it's probably not optimal, but it seems to work ;)

zariiii9003: Then i'd prefer the *no file* implementation with importlib 😄

andlaus: I agree, but since this is quite a bit away from my home turf, I'm not sure how to do this. (any hints?)

andlaus: after a bit of digging. I found out that the `importlib.metadata` path already works with the patch I proposed:
```
>>> from importlib.metadata import version
>>> version('cantools')
'39.3.1.dev6+g20191b97'
```

that said, I'm a bit ambivalent of whether it is a good idea to set the `__version__` variable in `__init__py` this way because it makes things quite twisted...

zariiii9003: I merged your PR and updated it to use `importlib.metadata`. It works fine.

andlaus:
I just tagged the 39.4.0 release, mainly to see if everything works as expected. it does :)

Issue #625, opened 2023-11-17
TRC file support for plotting


joooeey:
I need to plot CAN data from TRC files in place of candump files.

There are (at least) two ways to go about this but none of them work out of the box:
- Add the ability to read TRC files directly into `cat data.trc | python -m cantools plot db.dbc`. TRC files are not supported in `cantools plot` right now. This requires a few new lines of code (basically two new regexes). I have this coded and tested locally. If there's interest from the maintainers to add this functionality, I'm happy to submit a PR including a unit test.
- Convert the TRC to candump with `python -m data.trc data.log` and then feed it into cantools: `cat data.log | python -m cantools plot db.dbc`. This doesn't work either because python-can adds an R at the end of every line in its candump (.log) and cantools doesn't know what to do with this R. I couldn't find any authoritative documentation on what that R is. In any case, python-can adds an R or a T in [this code line](https://github.com/hardbyte/python-can/blob/bc3f95544e7a7dbec0da855a0873c6ea502147fc/can/io/canutils.py#L183C29-L183C29) to indicate receipt or transmission. Perhaps we (I) can include this in the [`RE_CANDUMP_LOG` regex](https://github.com/cantools/cantools/blob/eb9e42d88a1e8dfde5d33cfff6a1fecfcdfbb531/src/cantools/subparsers/plot.py#L94) as well?

andlaus:
> Perhaps we (I) can include this in the [RE_CANDUMP_LOG regex](https://github.com/cantools/cantools/blob/eb9e42d88a1e8dfde5d33cfff6a1fecfcdfbb531/src/cantools/subparsers/plot.py#L94) as well?

we should probably go this route. I've never heard about the TRC file format, but if there alread is a way to convert it to the candump log format, let's leverage that. (PR welcome.)

joooeey:
> > Perhaps we (I) can include this in the [RE_CANDUMP_LOG regex](https://github.com/cantools/cantools/blob/eb9e42d88a1e8dfde5d33cfff6a1fecfcdfbb531/src/cantools/subparsers/plot.py#L94) as well?
> 
> we should probably go this route. I've never heard about the TRC file format, but if there alread is a way to convert it to the candump log format, let's leverage that. (PR welcome.)

[...]

PR #630, opened 2023-11-29
Plotting support for "R" or "T" at end of `candump -l` output lines.

joooeey:
At the end of every line of output of `candump -l`, an "R" for receipt or "T" for transmission could [show up](https://github.com/linux-can/can-utils/blob/41ea759237bbd156d753251c573c280d96873528/asc2log.c#L218). Currently, we can't read such files. This PR adds support for these two indicators only.

See issue #625.


andlaus:
> Is cantools called line by line? Or is it called once after the log file is complete? As of now, it is not possible to decode a single line with python-can, but it doesn't have to stay that way.

on linux, pipes are basically just normal files that cannot be seeked (but if I'm not mistaken, they should -- in principle -- work fine with binary data), so the lines are processed as soon as they hit the pipe. The current logparser reads the input line by line, but as far as I can see, that's mainly because all of the currently supported formats work this way. IIRC, the format auto-detection of the log reader in cantools works by testing a bunch of regular expressions on the first line read and on the consecutive lines it only uses the first regex which matched. I think a similar approach could be implemented with the logreaders of python-can: readers for ASCII formats would get a few "canary lines" while readers of binary formats could get a few "canary bytes" to determine if the format is applicable. IOW, having something like an `AnyFormatLogReader` which instantiates reader objects for all available formats and proceeds with the one that found the canary to be digestible would be nice. If you don't want to do this work yourself, I'll have a shot at this (it'll probably be a while until I have time, though.)

joooeey:
> Generally, most of the functionality in the cantools examples could be covered by a small API change from this
> 
> ```shell
> cat candump-2021-01-04_180521.log | python3 -m cantools plot tests/files/dbc/abs.dbc
> ```
> 
> to this
> 
> ```shell
> python3 -m cantools plot --database tests/files/dbc/abs.dbc candump-2021-01-04_180521.log
> ```
> 
> But i do not know enough about linux and the pipe operator for this one:
> 
> ```shell
> candump vcan0 | python3 -m cantools decode tests/files/dbc/motohawk.dbc
> ```
> 
> Is cantools called line by line? Or is it called once after the log file is complete? As of now, it is not possible to decode a single line with python-can, but it doesn't have to stay that way.

Yes, currently cantools is called line-by-line. To me it looks like the current API is to mirror the structure of other cantools functionality that work in real-time. But from my understanding of the code we don't currently support real-time plotting anyways.

I'd like to request to move on with this little patch if there are no concerns with it and keep discussing the other options on the issue (#625) that I opened.

zariiii9003:
@joooeey Can you also adapt the canutils regular expressions in `decode.py`?

joooeey:
> @joooeey Can you also adapt the canutils regular expressions in `decode.py`?

Good idea! I'll implement that on Monday.

joooeey:
With the commit that I just added, the `decode` command also works with streams with the "R" or "T" at the end. I've also included a test.

While digging through the code, I noticed that `logreader.py` and hence `decode.py` already supports TRC files too.

In the medium term, it would indeed make sense to rely on python-can functionally for parsing files. This would avoid many redundancies. However, it requires a lot of work to make the interfaces compatible (One example: The way timestamps are handled in `cantools/logreader.py` and `cantools/subparsers/plot.py` is completely different). I don't have time for contributing that work.

andlaus:
> While digging through the code, I noticed that logreader.py and hence decode.py already supports TRC files too.

I suppose this is a bit confusing: while pretty similar to `python-can`'s `LogReader`, the  `LogReader` class in `cantools` is something completely separate and while it seems to support TRC (at least partially), BLF and a bunch of other formats supported by `python-can`'s class are not. (I guess that's for historical reasons; I'd love to replace our own class by the one from `python-can`, but unfortunately the latter does not (yet?) support auto-detection of the format...)

as a quick fix, you could give converting the `plot` utility to the internal `LogReader` class a shot. (this shouldn't be too difficult...)

joooeey:
> as a quick fix, you could give converting the `plot` utility to the internal `LogReader` class a shot. (this shouldn't be too difficult...)

That was the first thing I tried yesterday. It's not quick either, mostly because timestamps are treated differently in `logreader.py` and `plot.py`. The way I see it, implementing a reader with auto-detection capability in `python-can` and using that would simplify `cantools` a lot and limit redundancies. Unfortunately, I don't have the time to contribute such a feature.

Does anything still need to be changed or added to this PR here?

zariiii9003:
>  I'd love to replace our own class by the one from `python-can`, but unfortunately the latter does not (yet?) support auto-detection of the format...)

It would be easier if we could assume, that the string piped to cantools is either a file path or a canutils-candump line. 
If the string is a filepath, we pass it to `can.LogReader()`, otherwise we instantiate a `can.CanutilsLogReader()` directly.

andlaus:
> Does anything still need to be changed or added to this PR here?

okay, let's merge this PR as a quick-fix with a mental note that `plot` should be converted to the internal `LogReader` and the internal logreader ought to be replaced by `python-can`'s.

joooeey:
Thanks for merging! It's helpful for our research partners and us to have this patch included in the library.

PR #637, opened 2023-12-28
Fix range check with negative scalings

### LINKED ISSUE #636: Message._assert_signal_values_valid() uses incorrect minimum raw values for signals with negative scale
If a message has a signal with a negative scale, then when `_assert_signal_values_valid()` is called with `scaling=False`, `min_effective` will be computed as `signal.conversion.numeric_scaled_to_raw(signal.minimum)`, which actually computes the maximum raw value, not the minimum raw value.

zariiii9003:
Closes #636

### DIFF
--- a/src/cantools/database/can/message.py
+++ b/src/cantools/database/can/message.py
@@ -750,36 +750,32 @@ class Message:
                 if signal_value_num is None:
                     raise EncodeError(f'Invalid value specified for signal '
                                       f'"{signal.name}": "{signal_value}"')
-
                 continue
 
+            if scaling:
+                scaled_value = signal_value
+            else:
+                if signal.conversion.choices and signal_value in signal.conversion.choices:
+                    # skip range check if raw value exists in value table
+                    continue
+
+                # scale signal to perform range check against minimum and maximum
+                scaled_value = signal.conversion.raw_to_scaled(
+                    raw_value=signal_value, decode_choices=False)  # type: ignore[assignment]
+
             if signal.minimum is not None:
-                min_effective = signal.minimum
-
-                # undo the scaling of the signal's minimum value if we
-                # are not supposed to scale the input value
-                if not scaling:
-                    min_effective = signal.conversion.numeric_scaled_to_raw(signal.minimum)
-
-                if signal_value < min_effective - signal.conversion.scale*1e-6:
+                if scaled_value < signal.minimum - abs(signal.conversion.scale)*1e-6:
                     raise EncodeError(
                         f'Expected signal "{signal.name}" value greater than '
-                        f'or equal to {min_effective} in message "{self.name}", '
-                        f'but got {signal_value}.')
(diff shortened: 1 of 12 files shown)


src/cantools/database/can/message.py
```diff
@@ -750,36 +750,32 @@ def _assert_signal_values_valid(self,
...
+                    # skip range check if raw value exists in value table
+                    continue
 
-                # undo the scaling of the signal's minimum value if we
-                # are not supposed to scale the input value
-                if not scaling:
-                    min_effective = signal.conversion.numeric_scaled_to_raw(signal.minimum)
+                # scale signal to perform range check against minimum and maximum
+                scaled_value = signal.conversion.raw_to_scaled(
+                    raw_value=signal_value, decode_choices=False)  # type: ignore[assignment]
```
andlaus, line 764: possibly it is better to cast this to `float` instead of using `# type: ignore`:
```suggestion
                scaled_value = cast(float, signal.conversion.raw_to_scaled(
                    raw_value=signal_value, decode_choices=False))
```

andlaus (commented):
looks good, but one of the `if` conditions looks to be inverted...

src/cantools/database/can/message.py
```diff
@@ -750,36 +750,32 @@ def _assert_signal_values_valid(self,
                 if signal_value_num is None:
                     raise EncodeError(f'Invalid value specified for signal '
                                       f'"{signal.name}": "{signal_value}"')
-
                 continue
 
-            if signal.minimum is not None:
-                min_effective = signal.minimum
+            if scaling:
```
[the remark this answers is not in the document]
andlaus: okay, got it: you're converting raw to scaled values here, and if scaling is enabled, the specified signal value already is scaled. a comment would probably helpful:
```suggestion
            # retrieve the signal's scaled value
            if scaling:
                # if scaling is enabled, the specified signal value is already scaled
```

src/cantools/database/can/message.py
```diff
@@ -750,36 +750,32 @@ def _assert_signal_values_valid(self,
...
+                scaled_value = signal_value
+            else:
+                if signal.conversion.choices and signal_value in signal.conversion.choices:
```
andlaus, line 758: that condition probably must go one level up (scaled values can also be enums)

zariiii9003: That's true, fixed.

src/cantools/database/can/message.py
zariiii9003: Done

src/cantools/database/can/message.py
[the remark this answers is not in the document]
zariiii9003: Done

andlaus (approved):

andlaus:
all cool now, let's get that merged!

PR #639, opened 2023-12-30
Add type annotations to c_source.py

### LINKED ISSUE #604: generate_c_source: --use-float should generate float constants
E.g. instead of "0.5" should be generated "0.5F".

zariiii9003:
- add type annotations to c_source.py
- use float literals if `--use-float` is set
- do not cast `double` to `float`

Closes #604

### DIFF
--- a/src/cantools/database/can/__init__.py
+++ b/src/cantools/database/can/__init__.py
@@ -1,7 +1,15 @@
-from .bus import Bus as Bus
-from .database import Database as Database
-from .message import DecodeError as DecodeError
-from .message import EncodeError as EncodeError
-from .message import Message as Message
-from .node import Node as Node
-from .signal import Signal as Signal
+__all__ = [
+    "Bus",
+    "Database",
+    "DecodeError",
+    "EncodeError",
+    "Message",
+    "Node",
+    "Signal",
+]
+
+from .bus import Bus
+from .database import Database
+from .message import DecodeError, EncodeError, Message
+from .node import Node
+from .signal import Signal
--- a/src/cantools/database/can/__init__.py
+++ b/src/cantools/database/can/__init__.py
@@ -1,13 +1,3 @@
-__all__ = [
-    "Bus",
-    "Database",
-    "DecodeError",
-    "EncodeError",
-    "Message",
-    "Node",
-    "Signal",
-]
-
 from .bus import Bus
 from .database import Database
 from .message import DecodeError, EncodeError, Message
--- a/src/cantools/database/can/__init__.py
+++ b/src/cantools/database/can/__init__.py
@@ -1,5 +1,7 @@
-from .bus import Bus
-from .database import Database
-from .message import DecodeError, EncodeError, Message
-from .node import Node
-from .signal import Signal
+from .bus import Bus as Bus
+from .database import Database as Database
+from .message import DecodeError as DecodeError
+from .message import EncodeError as EncodeError
+from .message import Message as Message
+from .node import Node as Node
+from .signal import Signal as Signal
--- a/src/cantools/database/can/c_source.py
+++ b/src/cantools/database/can/c_source.py
@@ -1,12 +1,23 @@
 import re
 import time
 import warnings
-from typing import TYPE_CHECKING, List, Optional, Set, Tuple
+from typing import (
+    TYPE_CHECKING,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    TypeVar,
+    Union,
+    cast,
+)
 
 from cantools import __version__
 
 if TYPE_CHECKING:
-    from cantools.database import Database
+    from cantools.database.can import Database, Message, Signal
 
 HEADER_FMT = '''\
 /**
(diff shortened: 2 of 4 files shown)


src/cantools/database/can/c_source.py
```diff
@@ -1205,148 +1223,153 @@ def _format_choices(signal, signal_name):
...
 
-    for signal in message.signals:
-        minimum = signal.minimum
-        maximum = signal.maximum
+    if maximum is not None:
+        maximum = signal.signal.scaled_to_raw(maximum)
 
-        if minimum is not None:
-            minimum = signal.scaled_to_raw(minimum)
+    if minimum is None and signal.minimum_value is not None:
```
andlaus, line 1264: The names are IMO confusing here: what's the difference between `signal.minimum` and `signal.minimum_value` and `signal.minimum_type_value` without looking up the code of the respective properties? the latter two should probably be renamed to something like e.g. `minimum_raw_value` and `minimum_c_type_value`...

src/cantools/database/can/c_source.py
```diff
@@ -1361,66 +1384,77 @@ def _generate_choices_defines(database_name, messages, node_name):
...
+def _generate_signal_name_macros(database_name: str,
+                                 messages: List["CodeGenMessage"],
+                                 node_name: Optional[str]) -> str:
     result = '\n'.join([
         '#define {}_{}_{}_NAME "{}"'.format(
             database_name.upper(),
             message.snake_name.upper(),
             signal.snake_name.upper(),
-            signal.name)
+            signal.signal.name)
```
andlaus, line 1409: it would be nice if the `CodeGen{Signal,Message}` objects would also be renamed. maybe to something like `cg{signal,message}`? (ideally, the "forwarded" properties to regular `Signal`/`Message` objects should also be removed...)

src/cantools/database/can/c_source.py
```diff
@@ -1606,7 +1657,7 @@ def _generate_helpers_kind(kinds, left_format, right_format):
     return helpers
 
 
-def _generate_helpers(kinds):
+def _generate_helpers(kinds: Tuple[Set[Tuple[str, int]], Set[Tuple[str, int]]]) -> str:
```
andlaus, line 1660: what is the `Set[Tuple[str, int]]` here? introducing some type variable to make it more obvious would be nice:
```suggestion
HelperKind = Tuple[str, int]

def _generate_helpers(kinds: Tuple[Set[HelperKind], Set[HelperKind]) -> str:
```

(only if `Tuple[str, int]` is indeed a `HelperKind`)

andlaus (commented):
looks good

src/cantools/database/can/__init__.py
```diff
@@ -1,5 +1,7 @@
...
-from .database import Database
-from .message import DecodeError, EncodeError, Message
-from .node import Node
-from .signal import Signal
+from .bus import Bus as Bus
+from .database import Database as Database
+from .message import DecodeError as DecodeError
+from .message import EncodeError as EncodeError
+from .message import Message as Message
+from .node import Node as Node
```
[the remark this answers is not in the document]
zariiii9003: I checked csource.py with `mypy --strict` and it complained about that file. These names are exported according to PEP 484:

> Modules and variables imported into the stub are not considered exported from the stub unless the import uses the import ... as ... form or the equivalent from ... import ... as ... form. (UPDATE: To clarify, the intention here is that only names imported using the form X as X will be exported, i.e. the name before and after as must be the same.)

Alternatively, we could put them in `__all__` if you like.

src/cantools/database/can/c_source.py
```diff
@@ -842,41 +858,39 @@ def _format_comment(comment):
...
 
-def _format_range(signal):
-    minimum = signal.minimum
-    maximum = signal.maximum
+def _format_range(signal: "CodeGenSignal") -> str:
+    minimum = signal.signal.minimum
+    maximum = signal.signal.maximum
 
-    phys_to_raw = float if signal.is_float else round
+    def phys_to_raw(x: Union[int, float]) -> float:
```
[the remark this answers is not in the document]
zariiii9003: Done

src/cantools/database/can/c_source.py
zariiii9003: I renamed the properties. I'm not sure, whether that logic here makes sense as it does not consider enums, but 🤷‍♂️

src/cantools/database/can/c_source.py
zariiii9003: I added `cg_` everywhere.

src/cantools/database/can/c_source.py
zariiii9003: Done

src/cantools/database/can/__init__.py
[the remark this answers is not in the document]
andlaus: Ok. I just think that this is a quite clumsy construct. does mypy complain without `--strict`?

andlaus (approved):

src/cantools/database/can/__init__.py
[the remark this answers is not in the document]
zariiii9003: I replaced it with `__all__`. That works too.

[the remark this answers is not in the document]
zariiii9003: Now i removed `__all__`, because it causes issues with these star imports:

```python3
# Remove once less users are using the old package structure.
from .can import *  # noqa: F403
```

Issue #642, opened 2024-01-09
Generated SYM Enum Lines Exceed Max Length


ciepme:
Hi, I'm having an issue working with cantool's generated SYM files where the enumeration definition exceeds the maximum allowable line length for tools like PCAN Symbol Editor (Error 001: Syntax error - Line too long). The issue appears after approx 1,000 characters. Considering using newlines between each value to remedy.

andlaus:
can you provide a patch?

ciepme:
> can you provide a patch?

Yes, I will be working toward this.

swaldhoer:
Should this issue not have been closed when https://github.com/cantools/cantools/pull/643 was merged?

PR #643, opened 2024-01-10
Enum newlines

ciepme:
Fixes issue #642 by creating a newline every 10 choice entries for generated sym files as part of the _dump_choice(signal) function

### DIFF
--- a/src/cantools/database/can/formats/sym.py
+++ b/src/cantools/database/can/formats/sym.py
@@ -781,13 +781,13 @@ def _dump_choice(signal: Signal) -> str:
         return ''
 
     enum_str = f'Enum={_get_enum_name(signal)}('
-    is_first_choice = True
-    for choice_number, choice_value in signal.choices.items():
-        if choice_number % 10 == 0 and not choice_number == 0:
-            enum_str += f'{","+nl if not is_first_choice else ""}{choice_number}="{choice_value}"'
-        else:
-            enum_str += f'{", " if not is_first_choice else ""}{choice_number}="{choice_value}"'
-        is_first_choice = False
+    for choice_count, (choice_number, choice_value) in enumerate(signal.choices.items()):
+        if choice_count % 10 == 0 and not choice_count == 0:
+            enum_str += ","
+            enum_str += nl
+        elif choice_count > 0:
+            enum_str += ", "
+        enum_str += f'{choice_number}="{choice_value}"'
     enum_str += ')'
     return enum_str
 
--- a/src/cantools/database/can/formats/sym.py
+++ b/src/cantools/database/can/formats/sym.py
@@ -775,13 +775,15 @@ def _get_enum_name(signal: Signal) -> str:
 def _dump_choice(signal: Signal) -> str:
     # Example:
     # Enum=DPF_Actv_Options(0="notActive", 1="active", 2="rgnrtnNddAtmtcllyInttdActvRgnrt", 3="notAvailable")
+    new_line = '\n'
+
     if not signal.choices:
         return ''
 
     enum_str = f'Enum={_get_enum_name(signal)}('
     is_first_choice = True
     for choice_number, choice_value in signal.choices.items():
-        enum_str += f'{", " if not is_first_choice else ""}{choice_number}="{choice_value}"'
+        enum_str += f'{",{new_line}" if not is_first_choice else ""}{choice_number}="{choice_value}"'
         is_first_choice = False
     enum_str += ')'
     return enum_str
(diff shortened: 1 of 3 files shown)


src/cantools/database/can/formats/sym.py
```diff
@@ -775,13 +775,18 @@ def _get_enum_name(signal: Signal) -> str:
...
 
     enum_str = f'Enum={_get_enum_name(signal)}('
     is_first_choice = True
     for choice_number, choice_value in signal.choices.items():
-        enum_str += f'{", " if not is_first_choice else ""}{choice_number}="{choice_value}"'
+        if choice_number % 10 == 0 and not choice_number == 0:
+            enum_str += f'{","+nl if not is_first_choice else ""}{choice_number}="{choice_value}"'
+        else:
+            enum_str += f'{", " if not is_first_choice else ""}{choice_number}="{choice_value}"'
         is_first_choice = False
```
andlaus, line 790: I suppose this is slightly wrong because `choice_number` is not necessarily consecutive (?):
```suggestion
        for i, (choice_number, choice_value) in enumerate(signal.choices.items()):
            if i > 0:
                enum_str += ', '
            if i > 0 and i%10 == 0:
                enum_str += '\n'
            enum_str += f'{choice_number}="{choice_value}"'
```

ciepme: Thanks @andlaus that's a great point.

ciepme: I've added a new commit based on your suggestion on non-consecutive enum values @andlaus and a unit test for multiline enum sym files

src/cantools/database/can/formats/sym.py
```diff
@@ -775,14 +775,19 @@ def _get_enum_name(signal: Signal) -> str:
...
 
     enum_str = f'Enum={_get_enum_name(signal)}('
-    is_first_choice = True
-    for choice_number, choice_value in signal.choices.items():
-        enum_str += f'{", " if not is_first_choice else ""}{choice_number}="{choice_value}"'
-        is_first_choice = False
+    for choice_count, (choice_number, choice_value) in enumerate(signal.choices.items()):
+        if choice_count % 10 == 0 and not choice_count == 0:
+            enum_str += ","
+            enum_str += nl
```
andlaus, line 787: this seems to be too indirect:
```suggestion
            enum_str += ',\n'
```

src/cantools/database/can/formats/sym.py
```diff
@@ -775,14 +775,19 @@ def _get_enum_name(signal: Signal) -> str:
 def _dump_choice(signal: Signal) -> str:
     # Example:
     # Enum=DPF_Actv_Options(0="notActive", 1="active", 2="rgnrtnNddAtmtcllyInttdActvRgnrt", 3="notAvailable")
+    nl = '\n'
+
```
andlaus, line 779:
```suggestion
```

src/cantools/database/can/formats/sym.py
```diff
@@ -775,14 +775,19 @@ def _get_enum_name(signal: Signal) -> str:
...
     if not signal.choices:
         return ''
 
     enum_str = f'Enum={_get_enum_name(signal)}('
-    is_first_choice = True
-    for choice_number, choice_value in signal.choices.items():
-        enum_str += f'{", " if not is_first_choice else ""}{choice_number}="{choice_value}"'
-        is_first_choice = False
+    for choice_count, (choice_number, choice_value) in enumerate(signal.choices.items()):
+        if choice_count % 10 == 0 and not choice_count == 0:
```
andlaus, line 785: better use the unequals operator:
```suggestion
        if choice_count % 10 == 0 and choice_count != 0:
```

andlaus (commented):
Once my nitpicks have been dealt with, I think this is ready.

src/cantools/database/can/formats/sym.py
ciepme: My understanding is that this way would be more universal, as "\" characters can't be added to fStrings in Python versions prior to 3.12. I'm totally onboard with updating, just want to make sure we're aligned.

andlaus: if I'm not horribly mistaken, the proposed code is exactly equivalent (albeit slightly more performant) for any python version featuring f-strings: Prior to python 3.12, it is just not possible to nest strings that are using the same delimiter (a la `f"the song goes: {", ".join(["la"]*10)}"` is not valid prior to python 3.12). that should not be done anyway; instead, use different string delimiters (like `f'the song goes: {", ".join(["la"]*10}'`)

ciepme: Thanks for the clarification. Adding this now.

andlaus:
all good now. thanks for your contribution!

swaldhoer:
Should this issue not have been closed when #643 was merged?

PR #645, opened 2024-01-27
slight modification to regex to enable small scaling factors that have scientific numbers in them

RCMast3r:
I ran into issues when I had a really small scaling factor in a sym file and it had a scientific number and this update to the tokenizer fixed the issue.


RCMast3r:
if need be I can add a test for the evaluation of such a sym file.

andlaus:
looks good, thanks!

Issue #667, opened 2024-06-10
Error while trying to run 'monitor' subparser from CLI


jiyojolly:
Hi,
I am getting the following error while trying to run 'monitor' subparser from CLI. 
Error:  "error: Can't instantiate abstract class Monitor with abstract method stop*

Cantools Version: 39.4.5
OS: Ubuntu 22.04.04 

It is a fresh install of Ubuntu. I have tried reinstalling cantools, pip, purging pip cache etc. I can confirm other subparses like list, decode, dump etc. is working as expected. 

Please let me know if you require any more information.

jiyojolly:
I was able to pinpoint the issue to the python-can package dependency. The latest version, python-can 4.4.0, causes an error. I rolled back the package to version 4.3.0, and that fixed the problem. It seems that the definition of the can.Listener base class, from which the derived Monitor class inherits, has changed.

mendola:
Here's a possible workaround for preventing that error while using python-can 4.4.0: https://github.com/cantools/cantools/pull/668

PR #668, opened 2024-06-12
Fix monitor by implementing abstractmethod can.Listener.stop

mendola:
Between `python-can`  [release-4.2](https://github.com/hardbyte/python-can/blob/release-4.2/can/listener.py) and [release-4.4.0](https://github.com/hardbyte/python-can/blob/release-4.4.0/can/listener.py), `can.Listener.stop` became an `abstractmethod`, which caused the `cantools` `monitor` command to fail with:
```
error: Can't instantiate abstract class Monitor with abstract method stop
```

See: https://github.com/cantools/cantools/issues/667

This implements a noop `stop()` method.

### DIFF
--- a/src/cantools/subparsers/monitor.py
+++ b/src/cantools/subparsers/monitor.py
@@ -558,6 +558,9 @@ class Monitor(can.Listener):
 
     def on_message_received(self, msg):
         self._queue.put(msg)
+    
+    def stop(self):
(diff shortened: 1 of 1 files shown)


zariiii9003:
Fixed upstream. I removed the abstractmethod in 4.4.1

PR #681, opened 2024-08-13
Add support for ranged enums in SYM databases

Kimmax:
SYM databases support ranged enum values, at least in version 6.0.  
They look like this:  
`Enum=EnumName(0="Value0", 1="Value 1", 2..5="Ranged Value 2-5")`

This commit adds support by expanding the 2..5 range (where the last number is included, as one might expect it to work like `range()`) into individual values.  

It might not be the best way to do this. I also don't know how to write a proper test for this.

### DIFF
--- a/src/cantools/database/can/formats/sym.py
+++ b/src/cantools/database/can/formats/sym.py
@@ -98,12 +98,14 @@ class Parser60(textparser.Parser):
         }
 
         re_string = r'"(\\"|[^"])*?"'
+        re_number = r'-?\d+(\.\d+)?([eE][+-]?\d+)?'
 
         token_specs = [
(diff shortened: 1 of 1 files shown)


andlaus:
Thanks! The `ruff` complaints of the CI seem to be unrelated to your changes and patch itself looks good as far as I can tell. (I'm not familiar with that file format, though.) Before merging, I would like to see a the unit test with ranged enums; E.g. add a `test_ranged_enum_sym_parsing` function analogous to [`test_multiline_enum_sym_parsing`](https://github.com/cantools/cantools/blob/master/tests/test_database.py#L1939) to `tests/test_database.py`.

Kimmax:
I'll try to cover this using a test

PR #684, opened 2024-08-14
fix errors raised by recent versions of `ruff`

andlaus:
this makes the CI pass again. (at least for non-MacOS systems.)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -55,7 +55,7 @@ master_doc = "index"
 
 # General information about the project.
 project = "cantools"
-copyright = "2015-2023, Erik Moqvist"
+project_copyright = "2015-2023, Erik Moqvist"
 author = "Erik Moqvist"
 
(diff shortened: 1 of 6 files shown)


andlaus:
@zariiii9003: do you have any idea how the macos issues can be fixed? To me, it seems like `pip` cannot install the `kiwisolver` module (why is this pulled in in the first place?)

zariiii9003:
It's a dependency of matplotlib. But i'm don't know why it fails to build. Did we test pypy on Apple silicon before?

andlaus:
> Did we test pypy on Apple silicon before?

Hm, okay. I'm not aware that anything CI-wise has changed over the last year  from our side. Maybe the issue goes away with the next release of kiwisolver. If you agree and also don't mind the changes of this PR, I'd be obliged to you if you pressed the Green Button :)

PR #685, opened 2024-08-14
Add support for -b modifier on variables in SYM

Kimmax:
This commit adds support for the -b modifier on variables in SYM databases


Kimmax:
Tests follow

PR #694, opened 2024-09-16
Add CLI flag to disable snake case conversion and use original casing found in the dbc

Paolo-Beci:
This PR introduces a new CLI option that allows users to disable the automatic conversion of `camel_to_snake_case` and instead use the original casing as found in the DBC file. The following changes are included:

- 	New Feature: Added a CLI flag `--original-casing` to toggle between snake_case conversion (flag omitted) and retaining the original casing (flag applied).

- 	Tests: Created a test to ensure that the new option works as expected: `test_generate_c_source_original_casing`.

- 	Documentation: Updated the README to reflect the new CLI option, providing instructions on how to use it.

These changes will give users more flexibility in how they interact with DBC files, especially in scenarios where preserving the original casing is important.

Let me know if further modifications are needed!

### DIFF
--- a/src/cantools/database/can/c_source.py
+++ b/src/cantools/database/can/c_source.py
@@ -605,6 +605,7 @@ class CodeGenSignal:
     def __init__(self, signal: "Signal") -> None:
         self.signal: Signal = signal
         self.snake_name = camel_to_snake_case(signal.name)
+        self.name = signal.name
 
     @property
     def unit(self) -> str:
@@ -804,6 +805,7 @@ class CodeGenMessage:
 
     def __init__(self, message: "Message") -> None:
         self.message = message
+        self.name = message.name
         self.snake_name = camel_to_snake_case(message.name)
         self.cg_signals = [CodeGenSignal(signal) for signal in message.signals]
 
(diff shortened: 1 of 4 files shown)


src/cantools/database/can/c_source.py
```diff
@@ -605,6 +605,7 @@ class CodeGenSignal:
     def __init__(self, signal: "Signal") -> None:
         self.signal: Signal = signal
         self.snake_name = camel_to_snake_case(signal.name)
+        self.name = signal.name
```
zariiii9003, line 608: This should be a `@property`

src/cantools/database/can/c_source.py
```diff
@@ -804,6 +805,7 @@ class CodeGenMessage:
 
     def __init__(self, message: "Message") -> None:
         self.message = message
+        self.name = message.name
```
zariiii9003, line 808: This should be a `@property`

src/cantools/database/can/c_source.py
```diff
@@ -899,7 +901,7 @@ def _generate_signal(cg_signal: "CodeGenSignal", bit_fields: bool) -> str:
                                       scale=scale,
                                       offset=offset,
                                       type_name=cg_signal.type_name,
-                                      name=cg_signal.snake_name,
+                                      name=cg_signal.snake_name if not original_casing else cg_signal.signal.name,
```
zariiii9003, line 904:
```suggestion
                                      name=cg_signal.name  if original_casing else cg_signal.snake_name,
```

src/cantools/database/can/c_source.py
Paolo-Beci: why?

src/cantools/database/can/c_source.py
Paolo-Beci: could you please submit a proposal? thanks!

src/cantools/database/can/c_source.py
zariiii9003: My intention was to have only a single source of truth (signal.name). But the same could be said about the other attributes of CodeGenSIgnal. So nevermind, it's fine as it is.

PR #695, opened 2024-09-18
Add test and type annotations for LongNamesConverter

### LINKED ISSUE #693: Bug in short name conversion
I think the following line needs protection ahead of it to verify the cut name is not already present.

https://github.com/cantools/cantools/blob/37a04b2c0d08a07512fc00416d90555d47046c77/src/cantools/database/can/formats/dbc.py#L429

Consider the following test data:

```
name                                 short_name                          cut_name                        result
SSSSSSSSSSSSSSSSSSSSSSSSSSSXLLLLA    SSSSSSSSSSSSSSSSSSSSSSSSSSSXLLLL    SSSSSSSSSSSSSSSSSSSSSSSSSSS     SSSSSSSSSSSSSSSSSSSSSSSSSSSXLLLL OK
SSSSSSSSSSSSSSSSSSSSSSSSSSSXLLLLB    SSSSSSSSSSSSSSSSSSSSSSSSSSSXLLLL    SSSSSSSSSSSSSSSSSSSSSSSSSSS     SSSSSSSSSSSSSSSSSSSSSSSSSSS_0000 OK
SSSSSSSSSSSSSSSSSSSSSSSSSSSYLLLLA    SSSSSSSSSSSSSSSSSSSSSSSSSSSYLLLL    SSSSSSSSSSSSSSSSSSSSSSSSSSS     SSSSSSSSSSSSSSSSSSSSSSSSSSSYLLLL OK
SSSSSSSSSSSSSSSSSSSSSSSSSSSYLLLLB    SSSSSSSSSSSSSSSSSSSSSSSSSSSYLLLL    SSSSSSSSSSSSSSSSSSSSSSSSSSS     SSSSSSSSSSSSSSSSSSSSSSSSSSS_0000 FAIL (should be SSSSSSSSSSSSSSSSSSSSSSSSSSS_0001)
```

zariiii9003:
Continuation of #660
Closes #693

### DIFF
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -1,10 +1,10 @@
 # Load and dump a CAN database in DBC format.
 
 import re
+import typing
 from collections import OrderedDict, defaultdict
 from copy import deepcopy
 from decimal import Decimal
-from typing import List
 
 import textparser
 from textparser import (
@@ -405,15 +405,14 @@ class Parser(textparser.Parser):
                 bs,
                 version))
 
+
 class LongNamesConverter:
+    def __init__(self) -> None:
+        self._next_index_per_cut_name: typing.DefaultDict[str, int] = defaultdict(int)
+        self._short_names: typing.Set[str] = set()
 
-    def __init__(self, database):
-        self._database = database
-        self._next_index_per_cut_name = defaultdict(int)
-        self._short_names = set()
-
-    def convert(self, name):
-        short_name = None
+    def convert(self, name: str) -> typing.Optional[str]:
+        short_name: typing.Optional[str] = None
 
         if len(name) == 32:
             self._short_names.add(name)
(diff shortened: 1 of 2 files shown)


andlaus (approved):
I would prefer to import the classes from `typing` instead of just the module because when raising the minimum python version 3.10, these type annotations can basically be lowercased, but I'm fine with the patch.

zariiii9003:
> I would prefer to import the classes from `typing` instead of just the module because when raising the minimum python version to 3.10, these type annotations can basically be lowercased, but I'm fine with the patch.

there's a conflict of `typing.Optional` and `textparser.Optional`. We can change when 3.10 is the minimum supported version.

andlaus:
okay. I would go to import `Optional` from `typing` and use `textparser.Optional` where the other `Optional` is needed. Anyway, the proposed solution is fine IMO, so I'll merge...

PR #696, opened 2024-09-18
monitor: fix crash while decoding message with bad length

trnila:
Decoding message with bad length (and using `--no-strict`) raises exception in decoding that is caught in `format_message` and returned as a string that causes uncaught exception when `--single-line` is not used:
```shell-session
$ ip link add dev vcan0 type vcan
$ ip link set up vcan0
$ cantools -d monitor -c vcan0 tests/files/dbc/bad_message_length.dbc --no-strict
  File "monitor.py", line 540, in update
    modified = self.update_messages()
  File "monitor.py", line 531, in update_messages
    self.try_update_message()
  File "monitor.py",line 417, in try_update_message
    formatted = [f'{timestamp:12.3f}  {lines[1]}']
IndexError: list index out of range
# in another window
$ cansend vcan0 001#11
```

This patch handles all decoding exceptions directly in monitor in a single place.

### DIFF
--- a/src/cantools/subparsers/__utils__.py
+++ b/src/cantools/subparsers/__utils__.py
@@ -1,5 +1,7 @@
 from typing import Iterable
 
+from cantools.database.errors import DecodeError
+
 from ..database.can.database import Database
 from ..database.can.message import Message
 from ..database.namedsignalvalue import NamedSignalValue
@@ -142,7 +144,7 @@ def format_message_by_frame_id(dbase : Database,
                             single_line,
                             allow_truncated=allow_truncated,
                             allow_excess=allow_excess)
-    except Exception as e:
+    except DecodeError as e:
         return f' {e}'
 
 def format_container_message(message : Message,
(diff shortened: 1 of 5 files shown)


src/cantools/subparsers/__utils__.py
```diff
@@ -135,12 +135,15 @@ def format_message_by_frame_id(dbase : Database,
...
-                          allow_truncated=allow_truncated,
-                          allow_excess=allow_excess)
+    try:
+        return format_message(message,
+                            data,
+                            decode_choices,
+                            single_line,
+                            allow_truncated=allow_truncated,
+                            allow_excess=allow_excess)
+    except Exception as e:
```
andlaus, line 145: catching all exceptions here is an awfully wide net. catching `DecodeError` is sufficient?

andlaus (commented):
except for the "catch all" it looks good to me. (I know, this problem is already present in the existing code, but it would be nice to have that fixed as a "drive-by")

andlaus:
all good now, thanks for the patch!

PR #698, opened 2024-09-23
Enable cache via env variable and improve cache key

trnila:
Enable database cache via environment variable `CANTOOLS_CACHE_DIR` so tools like `monitor`, `dump` or any other user scripts can start faster without any source code modification:

```shell-session
$ export CANTOOLS_CACHE_DIR=/tmp
$ cantools dump somebig.arxml
```

Cache key is extended with all `load` arguments like `prune_choices` or `frame_id_mask` as they influence the resulting database. Few unit tests were failing because different database load configurations had same cache key.

Database raw file content is removed from the key and replaced with file path and last modification time. Otherwise cache grows quickly in hundreds of megabytes.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -61,7 +61,7 @@ jobs:
         python -m pip install -e ".[windows-all]"
     - name: Test
       run: |
-        tox -e gh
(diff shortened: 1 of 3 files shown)


andlaus:
looks good, thanks.

PR #701, opened 2024-10-05
ARXML parsing: scientific and hexadecimal numbers

FlorianMickler:
My local arxml dealer got me some stuff with scientific notations and hexadecimal numbers in it. I am not sure if that is pure, but it works with some small tweaks to the parse_number_string function.

### DIFF
--- a/src/cantools/database/can/formats/arxml/utils.py
+++ b/src/cantools/database/can/formats/arxml/utils.py
@@ -15,15 +15,17 @@ def parse_number_string(in_string : str, allow_float : bool=False) \
     - Some ARXML editors seem to sometimes include a dot in integer
       numbers (e.g., they produce "123.0" instead of "123")
     """
-
     # the string literals "true" and "false" are interpreted as 1 and 0
     if in_string == 'true':
         return 1
     elif in_string == 'false':
         return 0
     # the input string contains a dot -> floating point value
-    elif '.' in in_string:
-        tmp = float(in_string)
+    else:
+        if in_string.lower().startswith('0x'):
+            tmp = float.fromhex(in_string)
+        else:
+            tmp = float(in_string)
 
         if not allow_float and tmp != int(tmp):
             raise ValueError('Floating point value specified where integer '
--- a/src/cantools/database/can/formats/arxml/utils.py
+++ b/src/cantools/database/can/formats/arxml/utils.py
@@ -2,7 +2,7 @@
 from typing import Union
 
 
-def parse_number_string(in_string : str, allow_float : bool=False) \
+def parse_number_string(in_string: str, allow_float: bool=False) \
     -> Union[int, float]:
     """Convert a string representing numeric value that is specified
     within an ARXML file to either an integer or a floating point object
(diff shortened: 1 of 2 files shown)


andlaus (commented):
ouch: I was not aware that hexadecimal floating point were even a thing. the patch looks not too shabby, but I'm pretty sure that lines 39 to 47 are unreachable after this and, as a consequence your proposed version would no longer accept octal numbers in C notation (a la "011" for 9).

FlorianMickler:
I am sorry, i was in a hurry yesterday and didn't check beyond my usecase nor did i run the tests.  I have rewrote this routine now, but am wondering: 

The actual hexstring's I encountered were only integers. The python float hexparsing will do that as a subset. 
If the autosar standard actually forbids this, we should obviously throw something if we encounter this. (Taking out line 29 to 31 would be enough to let the int(*,0) parse the hexstring)

also i am unsure about the empty string or spaces returning 0. but this is the original behavior of that routine, maybe this is mandated by the standard?

src/cantools/database/can/formats/arxml/utils.py
```diff
@@ -15,31 +15,51 @@ def parse_number_string(in_string : str, allow_float : bool=False) \
...
+            ret = 1
+
+        if in_string == 'false':
+            ret = 0
+
+        # parse hex strings
+        if in_string.startswith('0x'):
+            ret = float.fromhex(in_string)
+
+        # allow octal notation without an "o" after the leading 0
```
andlaus, line 33: I guess that comment is not correct. `int(x, 0)` does this, but it would complain about `"011"`. (Note: I like the python notation much better than the C one, but IIRC AUTOSAR goes with the latter...)

andlaus (commented):
besides the nitpick, the patch looks good to  me...

andlaus:
> I am sorry, i was in a hurry yesterday and didn't check beyond my usecase nor did i run the tests. 

no worries, finding such issues is the job of maintainers. (i.e., me and a few others in the case of cantools.)

> If the autosar standard actually forbids this, we should obviously throw something if we encounter this. (Taking out line 29 to 31 would be enough to let the int(*,0) parse the hexstring)

if the tools used by your local ARXML dealer eat this, we should try to support this atrocity IMO...

> also i am unsure about the empty string or spaces returning 0. but this is the original behavior of that routine, maybe this is mandated by the standard?

Maybe it is, maybe it is no: The current approach of cantools w.r.t. ARXML is "whack-a-mole": if something is encountered in the wild and the fix does not break anything important, we aim at supporting it, even if it is not standard compliant. The fact that the routine did something originally just means that nobody complained about it yet ;)

src/cantools/database/can/formats/arxml/utils.py
andlaus: @FlorianMickler: do you indent to fix the comment or do you want the PR merged as-is?

andlaus: I just noticed: This should be an `elif` instead of `if` because strings starting with `0x` also fulfill that condition, so in effect all strings prefixed with a `0` would be interpreted as octal...

FlorianMickler: Hi, sorry for replying late again. I don't get your complaint about the comment. Maybe I am being stupid, but octal notation without an o (letter o) after the leading 0 (zero) is what this check is for.  Or  not?

FlorianMickler: > @FlorianMickler: do you indent to fix the comment or do you want the PR merged as-is?

Well, I'd like to understand the problem first before making that decision ..

andlaus: ok, I had a sign error here: I missed the "not" ;)

andlaus:
ok, all good. merging...

PR #708, opened 2024-10-20
Use tox and CMake to test generated C code on Windows

zariiii9003:
@andlaus I'd like your opinion on this.

I want to make testing the C code easier and platform independent via `tox -e codegen`.

Summary of the changes:
- move C test files from /tests to /tests/test_c_code
- delete all MakeFiles and create CMakeLists.txt
- change test framework from nala to unity, because nala does not compile with MSVC. The syntax is pretty much the same, but there is no automatic test discovery, so i added `test_runner.c` and header files.

There is one failing test: `test_sizeof_types`. The code was previously compiled with the `-fpack-struct` flag, but I don't think MSVC can replicate the same bitfield layout as gcc with `-fpack-struct`. I want to remove this test entirely, the size of the structs seems irrelevant, it's just an internal representation and does not affect the data layout on the bus.


.github/workflows/pythonpackage.yml
```diff
@@ -7,32 +7,20 @@ env:
 
 jobs:
   codegen:
-    runs-on: ubuntu-latest
+    runs-on:  ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.8", "3.12"]
+        os: [ "ubuntu-latest", "windows-latest", "macos-latest" ]
```
andlaus, line 13: it would be great if we could keep testing with the oldest supported python version as well, i.e. something like

```suggestion
        os: [ "ubuntu-latest", "windows-latest", "macos-latest" ]
        python-version: ["3.8", "3.13"]
```

(not sure if github-actions likes this...)

tests/test_c_code/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+
+
+# fetch unity test framework
+include(FetchContent)
+FetchContent_Declare(
+    unity
+    GIT_REPOSITORY https://github.com/ThrowTheSwitch/Unity.git
+    GIT_TAG v2.6.0
+)
+FetchContent_MakeAvailable(unity)
```
andlaus, line 47: hm, I'm not sure if it is a good idea to (potentially) retrieve stuff from the internet as part of the `cmake` config process. IMO, the clean solution would be to install the dependency before and `Find` it using cmake. That said, given that this might become quite iffy and the unity testing framework is comprised of only three files, I think that simply grabbing the latest copy and shipping it as part of cantools is a valid option as well...

tests/test_c_code/test_basic.c
```diff
@@ -0,0 +1,619 @@
...
+#include "min_max_only_6_0.h"
+#include "abs.h"
+
+#include <stdio.h>
+#include <stdint.h>
+#include <string.h>
+#include <math.h>
+#include "unity.h"
+
+static bool fequal(double v1, double v2)
```
andlaus, line 22: better call this `fsimilar`. (and possibly declare it `static inline`?)

tests/test_c_code/test_basic.c
```diff
@@ -0,0 +1,619 @@
...
+#include "signed.h"
+#include "my_database_name.h"
+#include "min_max_only_6_0.h"
+#include "abs.h"
+
+#include <stdio.h>
+#include <stdint.h>
+#include <string.h>
+#include <math.h>
+#include "unity.h"
```
andlaus, line 20: IMO, that header name has quite a bit potential for confusion. is is possible to include it from a sub-directory a la `unity-unit-test-framework/unity.h`

andlaus (commented):
besides the issue of having to be connected to the internet while running cmake, this looks pretty good to me.

about the issue with the packed structures, I agree that structure size should be considered to be an internal implementation detail that should not be tested (i.e., removing the test is fine with me). IMO, we could consider to add support for [decorating the generated structures with `__attribute__((packed))`](https://stackoverflow.com/questions/1537964/visual-c-equivalent-of-gccs-attribute-packed) but I'd rather not do that because even then it would be pretty difficult to guarantee that size of the frame structs is the size of the frame on the wire.

.github/workflows/pythonpackage.yml
zariiii9003: Done

tests/test_c_code/test_basic.c
zariiii9003: That existed before my change, but unity provides functions for comparing doubles. I'll replace this.

zariiii9003:
CI fails because of a bug in Python 3.13.0: https://github.com/python/cpython/issues/125235

tests/test_c_code/CMakeLists.txt
zariiii9003: I thought about this and i'm not sure if the other options are really better. I've found a few example projects which use git submodules for Unity. That would work but would require the user to clone recursively.

andlaus: why not vendor it? it is set to the v2.6.0 tag anyway and I think we should not assume that the machine which runs the test has access to the internet at configure time...

zariiii9003: Done :)

andlaus (approved):
looks good to me. merge at will if you're happy with it...

PR #711, opened 2024-11-27
fix floating-point precision issue in arxml 'parse_number_string'

sfpetersen:
Currently, the arxml `parse_number_string` utils function parses numeric values as a `float` first, and converts to integer if an integer is required. For very large integers, floating-point error causes the integer to be parsed incorrectly.

```
>>> int(2**64 -1)
18446744073709551615
>>> int(float(2**64 -1))
18446744073709551616
```

This pull request orders the parsing attempts to prevent using `float` at all for parsing integer strings when possible.


andlaus:
I have no objections and the tests pass. merging...

PR #712, opened 2024-11-28
Encode the given NamedSignalValue 'value' when explicitly given

sfpetersen:
Some signal definitions may have multiple 'named signal values' with the same name but different raw numeric values (things like "reserved"). Currently, when the `encode()` function is called with signal values of type `NamedSignalValue`, the `.name` attribute is used to look up the appropriate raw numeric value. In the case of multiple 'named signal values' with the same name, this may not be encoding what the caller wants.

This PR changes the `encode()` behavior to encode the NamedSignalValue `.value` attribute, while still checking that the given name/value pair exists in the signal definition.

### DIFF
--- a/src/cantools/database/utils.py
+++ b/src/cantools/database/utils.py
@@ -89,8 +89,18 @@ def _encode_signal_values(signals: Sequence[Union["Signal", "Data"]],
             raw_values[name] = value if conversion.is_float else round(value)
             continue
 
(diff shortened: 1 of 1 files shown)


andlaus:
I'm not sure if input files that map multiple values to the same string is valid, but I don't see any harm in merging this. thanks for the patch!

PR #713, opened 2024-12-02
When parsing DBC, use long names in SignalGroup.signal_names to match Signal.name

sfpetersen:
Currently, when loading from a DBC with a `SystemSignalLongSymbol` attribute defined for a given signal, the signal name used in the `Signal` object will be full "long name" of the signal. But when loading signal groups, the "short names" are used in the `SignalGroup` object. It is difficult for the calling code to reconcile this discrepancy because the "short name" is not maintained anywhere in the `Signal` object.

This PR makes the `SignalGroup` objects use the "long names" of signals when loading from DBC, copying the same logic used when loading `Signal` objects.


zariiii9003 (approved):

sfpetersen:
Would it be possible to make a release to get this in the pypi package?

andlaus:
done

PR #714, opened 2024-12-02
Sym generator does not produce unique mux names

kodfadern:
DBC files support re-using the same Mux signal name for all multiplexed signals within that frame. Sym generator does not support this case since Sym files require mux names to be unique. 
A simple fix is suggested to create a unique mux-name-tag in case this happens.

### DIFF
--- a/src/cantools/database/can/formats/sym.py
+++ b/src/cantools/database/can/formats/sym.py
@@ -865,7 +865,7 @@ def _dump_signals(database: InternalDatabase, sort_signals: TypingOptional[Calla
         return ''
 
 def _dump_message(message: Message, signals: List[Signal], min_frame_id: TypingOptional[int], max_frame_id: TypingOptional[int] = None,
-                  multiplexer_id: TypingOptional[int] = None, multiplexer_signal: TypingOptional[Signal] = None) -> str:
+                  multiplexer_id: TypingOptional[int] = None, multiplexer_signal: TypingOptional[Signal] = None, multiplexer_signal_names: TypingOptional[List[str]] = None) -> str:
     # Example:
     # [TestMessage]
     # ID=14A30000h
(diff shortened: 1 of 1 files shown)


andlaus (commented):
#715 seems to have caused a conflict here. since it only affects the SYM format (which I'm not familiar with) and the current master version does not work anyway for the cases this PR fixes, I see no harm in merging this once the conflicts are resolved...

andlaus:
@kodfadern: can you please resolve the merge conflict here? It seems like this is a useful fix for people who use SYM files with multiplexed messages...

kodfadern:
> @kodfadern: can you please resolve the merge conflict here? It seems like this is a useful fix for people who use SYM files with multiplexed messages...

I could fix it but I realized that there are more significant problems with the current SYM generator. 
For example I want to be able to generate files between different formats and compare them for differences but whenever I go through SYM format there is information lost.
I think the main problem is the structure of the SYM file that is currently being generated. 
One problem is that the way it is currently formatted will require that signal names are globally unique (this is not required for other formats, and it is actually not required for SYM files either if the formatting is a bit different). If the format is fixed properly then it will not even be necessary with the fix I have provided in this PR.
I think the format should be as follows:

FormatVersion=6.0 // Do not edit this line!
FloatDecimalPlaces=0
Title="Example_AllSymbols"

{ENUMS}
Enum=ExampleEnum_m(1="Enum1", 0="Enum0")

{SENDRECEIVE}

[ExampleFrame]
ID=12h
Len=8
Var=ExampleSignal1 unsigned 8,4
Var=ExampleSignal0 unsigned 4,4 /e:ExampleEnum_m

If the signals are defined inside the frames in this manner then a lot of the problems with the current generator goes away.
Do you think it would make sense to do this rewrite of the whole structure or should I go ahead with my quick-fix in this PR?

andlaus:
> Do you think it would make sense to do this rewrite of the whole structure or should I go ahead with my quick-fix in this PR?

if you're willing to invest the work, I'm happy to support you where I can: I'm not familiar with the SYM format, so having somebody who knows what he's talking about is highly appreciated :). (just make sure not to change the generic infrastructure of cantools in incompatible ways...)

PR #715, opened 2024-12-04
Raise Minimum Python Version to 3.9

zariiii9003:


andlaus (approved):
LGTM

andlaus:
@zariiii9003: merge at will. (I suppose we should bump the version to `40.0.0` after that!?)

PR #716, opened 2024-12-20
Add option for displaying unit from dbc in the plot legend

joooeey:
This pull request adds a command line option to show the unit of each signal in the legend.

Excerpt from the help (The last line is new):

```
  The following options can be used to configure the subplots/axes.
  If they shall apply to a specific subplot/axis they must be placed among the signals for that subplot/axis and a -- must mark the end of the global optional arguments.
  Otherwise they are used as default value for each subplot/axis.

  --title TITLE
  --color COLOR         The color to be used for the y-label and the signals
                        (unless a different color is given for the signal).
                        All string formats explained in the following link are
                        allowed:
                        https://matplotlib.org/tutorials/colors/colors.html
                        (default: None)
  --xlabel XLABEL
  --ylabel YLABEL
  --ymin YMIN
  --ymax YMAX
  -u, --show-units      Add units to the legend labels. (default: False)
```

Result:

```
python -m cantools plot -u tests/files/dbc/abs.dbc
(1609779922.655421) vcan0 00000343#B204B9049C049C04
(1609779922.655735) vcan0 0000024A#120527052E051905
(1609779923.657524) vcan0 00000343#C404C404CB04C404
(1609779923.658086) vcan0 0000024A#8B058B058B059205
(1609779924.659912) vcan0 00000343#5C04790479045504
(1609779924.660471) vcan0 0000024A#44064B0659064406
(1609779925.662277) vcan0 00000343#15040704F203F203
(1609779925.662837) vcan0 0000024A#8B069906A706A706
(1609779926.664191) vcan0 00000343#BC03B503A703BC03
(1609779926.664751) vcan0 0000024A#A006A706C406C406
```
unit_example

### DIFF
--- a/src/cantools/subparsers/plot.py
+++ b/src/cantools/subparsers/plot.py
@@ -666,7 +666,7 @@ class Signals:
                         splot.axes.xaxis.set_major_formatter(lambda x,pos: str(datetime.timedelta(seconds=x)))
                     axis_format_uninitialized = False
                 plt_func = getattr(splot, sgo.plt_func)
-                if self.global_subplot_args.show_units or self.subplot_args[(sgo.subplot, sgo.axis)].show_units:
+                if self.global_subplot_args.show_units and self.subplot_args[(sgo.subplot, sgo.axis)].show_units:
                     unit = self.get_signal_unit(signal_name)
                     signal_name = f"{signal_name} [{unit}]"
                 container = plt_func(x, y, sgo.fmt, label=signal_name)
@@ -946,7 +946,8 @@ def add_subplot_options(arg_group):
     arg_group.add_argument('--ymin', type=float)
     arg_group.add_argument('--ymax', type=float)
     arg_group.add_argument(
-        '-u', '--show-units',
-        action='store_true',
-        help='Add units to the legend labels.')
+        '--no-units',
+        dest='show_units',
+        action='store_false',
+        help='Do not show units in legend labels.')
     return arg_group
--- a/src/cantools/subparsers/plot.py
+++ b/src/cantools/subparsers/plot.py
@@ -413,7 +413,7 @@ class Plotter:
         self.ignore_unknown_frames = args.ignore_unknown_frames
         self.ignore_invalid_data = args.ignore_invalid_data
         self.output_filename = args.output_file
-        self.signals = Signals(args.signals, args.case_sensitive, args.break_time, args, args.auto_color_ylabels)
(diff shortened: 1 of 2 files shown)


andlaus:
why make this an option? is there a reason not to show the units in the labels?

joooeey:
> why make this an option? is there a reason not to show the units in the labels?

I wouldn't be surprised if there are DBCs out there with wrong units. So I chose to make it an option to ensure full backwards-compatibility of the plot output.

I see three ways to do this:
1. Opt-in to get the units with the `--show-units` option as proposed here.
2. Opt-out of seeing the units with a `--no-units` option.
3. Don't have an option and just show the units to keep the interface simple.

Now that I think of it, 2 seems the most sensible to me. Options 2 and 3 may warrant bumping the version number to 41 because it changes the way plots look by default which could be relevant when running old scripts.

andlaus:
>  Now that I think of it, 2 seems the most sensible to me. 

I think so as wel. Once you pushed the fix, I'll merge... 

> Options 2 and 3 may warrant bumping the version number to 41 because it changes the way plots look by default which could be relevant when running old scripts.

If we strictly sticked to semantic versioning this is true. However, I think that bumps in the major version are only warranted by major incompatible API changes or version bumps of important prerequisites (basically the minimum python version required). Since I would be surprised if the format of axis labels of plots is depended upon by anyone and it is IMO beneficial in the vast majority of cases to specify the units, I propose to increase the release number to 40.1. (which signals new but generally compatible features.)

joooeey:
> I think so as well. Once you pushed the fix, I'll merge...

I've implemented the 2nd way now, with the `--no-units` option.

I've changed most tests to either pass the `--no-units` option or to expect the unit in the legend labels. Which of these two ways to make the tests pass I chose was mostly arbitrary. However, I have made sure that one of the tests with multiple subplots has the unit off for one subplot. Likewise for one test with a secondary y-axis, the units are off on the secondary axis.

The tests listed in the [README](https://git.rwth-aachen.de/smart-forestry/stihl-showcase/-/blob/main/stihl_files/ms500idatalogger/docs/a3556f95-10df-4fe0-a80e-89d5164e808a.PNG?ref_type=heads) pass on my machine. I'm not sure what's up with the tests that fail in CI.

andlaus:
nice, thank you!

joooeey:
> nice, thank you!

I thank you for the quick turnaround!

PR #717, opened 2024-12-26
monitor: signals filtering

trnila:
Some databases contains a lot of messages, up to 64 1bit signals per CAN-FD message and sometimes you dont know how is the message named as you are mainly interested in the signals.

This pull request adds signal filtering so the new filtering behavior will be: 
- show all signals if a filter matches a message name
- **show only signals in a message that matches a filter**

Additionally, decoding errors are counted as `errors` instead of `discarded` messages.


src/cantools/subparsers/monitor.py
```diff
@@ -338,74 +350,68 @@ def process_user_input_filter(self, key):
...
             if message.is_container:
                 self._try_update_container(message, timestamp, data)
-                return
+            else:
+                if len(data) < message.length:
+                    self._update_message_error(timestamp, name, data, f'{message.length - len(data)} bytes too short')
+                    return MessageFormattingResult.DecodeError
 
+                if message.is_multiplexed():
+                    name = format_multiplexed_name(message,
```
andlaus, line 385: if you change `format_message()` to take the decoded data (good idea, btw), you should probably also adapt `format_multiplexed_name()` accordingly. (maybe this function should be renamed as well. `format_multiplexed_message()`?)

andlaus (commented):
It seems like this contains some slightly off-topic cleanups, but it generally is fine with me. I'll merge once the the issue about API consistency of the `format_*_message()` functions has been resolved...

src/cantools/subparsers/monitor.py
trnila: I have adapted `format_multiplexed_name` and kept the function name, as the function is returning `name` - is it okay?

Just found out that decoding containers was not showing correct data even on master - already decoded `cdata` should be used instead of decoding that was hidden in `format_message`. Fixed newly introduced test and added one more test to check encoding of multiplexed messages in a container.

Currently we are having little-bit duplicated code for formatting normal messages and containerized ones. The container version is also not respecting single-line configuration. I could refactor it and union behavior in following PR if the single-line should be respected.

andlaus: > Currently we are having little-bit duplicated code for formatting normal messages and containerized ones. The container version is also not respecting single-line configuration. I could refactor it and union behavior in following PR if the single-line should be respected.

that would be great. (when I implemented support for containers I simply did not test the single line stuff because without signal filtering this is basically useless for me...)

src/cantools/subparsers/monitor.py
```diff
@@ -338,74 +350,69 @@ def process_user_input_filter(self, key):
...
+                    decode_choices=True,
+                    allow_truncated=True,
+                    allow_excess=True
+                )
 
-            if len(data) < message.length:
-                self._update_message_error(timestamp, name, data, f'{message.length - len(data)} bytes too short')
-                return
+                if message.is_multiplexed():
+                    name = format_multiplexed_name(message, decoded_signals)
```
andlaus, line 391: shouldn't multiplexed messages be filtered for signals as well?

andlaus: ok, I got what this function does wrong. All is good...

andlaus:
ok, I think this is ready. let's merge. Thanks for the contribution!

AkhilTThomas:
Thank you for this feature ❤️

PR #719, opened 2025-01-15
Fix CI

andlaus:
the CI system recently started failing on us:

- A new version of ruff has been released which seems to complain about naming conflicts between custom submodules and standard python modules. We disable that check because renaming `signal.py` would probably leave a path of destruction in user code.
- pypy currently seems to be unable to install the (indirect) "pillow" dependency. For this reason, we let's disable testing using pypy for now...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,6 +101,7 @@ lint.ignore = [
     "E501",   # line too long
     "F541",   # f-string-missing-placeholders
     "RUF012", # mutable-class-default
+    "A005",   # Module shadows a Python standard-library module (needed for signal)
 ]
 lint.isort.known-first-party = ["cantools"]
 lint.per-file-ignores."__init__.py" = ["F401"]
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       matrix:
         os: [ "ubuntu-latest", "windows-latest", "macos-latest" ]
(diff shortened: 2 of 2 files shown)


andlaus:
@zariiii9003: would be great if you knew a way to get pypy working again instead of disabling it...

zariiii9003:
Pillow does not build wheels for 3.9 anymore. I'd try pypy-3.10 🤷‍♂️

pyproject.toml
```diff
@@ -101,6 +101,7 @@ lint.ignore = [
     "E501",   # line too long
     "F541",   # f-string-missing-placeholders
     "RUF012", # mutable-class-default
+    "A005",   # Module shadows a Python standard-library module (needed for signal)
```
zariiii9003, line 104: You could move this to per-file-ignores if it's just one file

andlaus:
> Pillow does not build wheels for 3.9 anymore. I'd try pypy-3.10 🤷‍♂️

ok: pypi-3.10 seems to work with pillow on windows, but it bails out due to some other reason which I could not grok from the error message. I disabled testing with pypy on windows, i.e., pypy is still tested on ubuntu...

pyproject.toml
andlaus: good point. fixed.

zariiii9003:
It seems like there's something missing in the matplotlib wheel for pypy...

andlaus:
can we fix this on our side? if not, let's just disable that config; After all, pypy and windows are still tested, just not both together...

zariiii9003 (approved):

PR #722, opened 2025-01-25
monitor: container supports singleline formatting

trnila:
Support singleline formatting for container messages.

Also unifies rendering of multiplexed messages of containers and standard one - internal name like `multiplexed_message__SELECT_HELLO` is shown as `multiplexed_message`. Is it fine?

Container name is indented by one space to the left. Is it intended behavior?
Maybe we can remove 'OneToContainThemAll ::` prefix and intend inner messages instead? But it may not work correctly with filtering (parent container can be hidden....). Also current indention is not so visible in multi-line rendering...
image


tests/test_monitor.py
```diff
@@ -1153,13 +1153,13 @@ def test_container_multiplex(self,
                 call(10, 0, '                  signal1: 0 m,'),
                 call(11, 0, '                  signal5: 4.0'),
                 call(12, 0, '              )'),
-                call(13, 0, '       0.000  OneToContainThemAll :: multiplexed_message__SELECT_HELLO('),
+                call(13, 0, '       0.000  OneToContainThemAll :: multiplexed_message('),
```
andlaus, line 1156: I'm not sure if removing the values of the selector signals here is a good idea. maybe it ought to be formatted more nicely, e.g., `OneToContainThemAll :: multiplexed_message :: SELECT_HELLO`?

andlaus (commented):
Looks (almost) good to me...

andlaus:
> Container name is indented by one space to the left. Is it intended behavior?

Good catch; it isn't intended behavior. Can you add a patch to fix this?

tests/test_monitor.py
trnila: multiple numeric selectors doesn't look good: `Extended :: 0 :: 0 :: 1`.
Maybe we can print selector name in case of multiple numeric values like `Extended :: S0=0 :: S1=0 :: S2=1`... but theres risk it may not fit to screen

trnila:
> > Container name is indented by one space to the left. Is it intended behavior?
> 
> Good catch; it isn't intended behavior. Can you add a patch to fix this?

Fixed and unified all other places (errors and undecoded container messages)

tests/test_monitor.py
```diff
@@ -800,12 +800,12 @@ def test_filter_muxed_signal(self,
             [
                 call(0, 0, 'Received: 2, Discarded: 0, Errors: 0'),
                 call(1, 0, '   TIMESTAMP  MESSAGE                                           ', 'green'),
-                call(2, 0, '       0.000  BATTERY_VT('),
+                call(2, 0, '       0.000  BATTERY_VT :: 0('),
```
andlaus, line 803: hm, this is rather more unintuitive than I anticipated. do you think it would be better to include the selector signal name in the multiplexer name? something akin
```
BATTERY_VT :: BATTERY_VT_INDEX=0 (
 [...]
)
```

tests/test_monitor.py
andlaus: I agree. let's try and if we come up with something better, we can change it later...

tests/test_monitor.py
trnila: Added multiplexor names

andlaus:
great! let's merge.

andlaus:
I just added you the the list of collaborators of the project. Welcome!

(Please do not push to master directly, but keep opening PRs, even if you merge them yourself.)

trnila:
I appreciate it, thank you :)

PR #724, opened 2025-02-09
Update CDD parser to load all possible data types

walshb421:
This allows a CDD with types not explicitly specified to be loaded in the intermediate list of data types. Only data types used/referenced by DIDs are actually included in the internal database.


andlaus:
Thanks. While I can't estimate the pros and cons of this change, I don't really mind merging it. That said, if you want to do non-trivial diagnostics, I would *strongly* advice to export your dataset to ODX using candela's ODX-Studio plugin, and continue using [odxtools](https://pypi.org/project/odxtools/) on the python side...

PR #726, opened 2025-02-15
add a strict par to fix #725

### LINKED ISSUE #725: EncodeError when init value is bigger than maximum value
### The issue:
When init value is bigger than maximum value, it will raise an EncodeError in tester

Image

Image

### Analysis:
in _prepare_initial_signal_values if initial exist, it will directly use it without any check.
`
    def _prepare_initial_signal_values(self):
        initial_sig_values = {}

        # Choose a valid set of mux settings
        mplex_settings = {}
        for m0 in reversed(self._mplex_map.values()):
            for m1 in m0:
                mplex_settings.update(m1)

        for signal in self.database.signals:
            minimum = 0 if not signal.minimum else signal.minimum
            maximum = 0 if not signal.maximum else signal.maximum
            if signal.initial:
                # use initial signal value (if set)
                initial_sig_values[signal.name] = signal.initial
            elif signal.is_multiplexer:
                initial_sig_values[signal.name] = mplex_settings.get(signal.name, 0)
            elif minimum <= 0 <= maximum:
                # use 0 if in allowed range
                initial_sig_values[signal.name] = 0
            else:
                # set at least some default value
                initial_sig_values[signal.name] = minimum
        return initial_sig_values
`
in _update_can_message it did not pass "strict" to database.encode, so encode will use default which is True
then it will raise an error in this encode function.

```
def _update_can_message(self):
        arbitration_id = self.database.frame_id
        extended_id = self.database.is_extended_frame
        pruned_data = self.database.gather_signals(self.data)
        data = self.database.encode(pruned_data,
                                    self.scaling,
                                    self.padding)
        self._can_message = can.Message(arbitration_id=arbitration_id,
                                        is_extended_id=extended_id,
                                        data=data)

    def encode(self,
               data: EncodeInputType,
               scaling: bool = True,
               padding: bool = False,
               strict: bool = True,
               ) -> bytes:

```

### Proposal:

After check the code, at least we have two proposals, pls let me know your comments, tks

#### Add check in _prepare_initial_signal_values

Image

#### Add an optional parameter "strict" 

this will require to add this parameter to below class/function, so that user can have control to encode function

1. cantools.tester.Tester
2. Message
3. Message.update
4. Message._update_can_message

### Why need this change
Sometimes in the DBC file there will be  a lot of such kind of issue, fix one by one will take time, so it would be better if user can have control whether raise error for such issue.

sgnes:
add a strict parameter for Tester and Message, so that user can have control whether raise EncodeError for #725

### DIFF
--- a/src/cantools/tester.py
+++ b/src/cantools/tester.py
@@ -122,7 +122,7 @@ class Message(UserDict):
                  decode_choices,
                  scaling,
                  padding,
-                 strict: bool = True):
+                 strict = True):
         super().__init__()
         self.database = database
         self._mplex_map = invert_signal_tree(database.signal_tree)
--- a/src/cantools/tester.py
+++ b/src/cantools/tester.py
@@ -121,7 +121,8 @@ class Message(UserDict):
                  input_queue,
                  decode_choices,
                  scaling,
-                 padding):
+                 padding,
+                 strict: bool = True):
         super().__init__()
         self.database = database
         self._mplex_map = invert_signal_tree(database.signal_tree)
@@ -130,6 +131,7 @@ class Message(UserDict):
         self.decode_choices = decode_choices
         self.scaling = scaling
         self.padding = padding
+        self.strict = strict
         self._input_list = input_list
         self.enabled = True
         self._can_message = None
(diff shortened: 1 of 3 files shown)


andlaus:
I'm unsure if it is a good idea to make it work for invalid input files, but I see no harm in merging this once the CI system passes...

sgnes:
> I'm unsure if it is a good idea to make it work for invalid input files, but I see no harm in merging this once the CI system passes...

one case is there will be alot of such issue in the customer  dbc, would be great if user have access to whether raise exception. 
this invalid initial value should not affect the functionality, as user can set whatever value they want to send.
Can we merge this chane, thanks!

PR #727, opened 2025-02-19
Added enum support to code generation

Ozuba:
I managed to add enum support to code generation by using the
`--use-enum-choices` flag with the `generate_c_source`
It should be an advantage for autocompletion using linters and debugging. However the reduction of tag lengths
is impossible do to not being able to guarantee tag uniqueness as pure c lacks scoped enums.

I still have to implement some tests but it seems to generate valid compilable units.


src/cantools/database/can/c_source.py
```diff
@@ -1357,14 +1382,23 @@ def _generate_choices_defines(database_name: str,
...
-            signal_choices_defines = '\n'.join([
-                f'#define {database_name.upper()}_{cg_message.snake_name.upper()}_{choice}'
-                for choice in choices
-            ])
-            choices_defines.append(signal_choices_defines)
-
-    return '\n\n'.join(choices_defines)
+            signal_choices_declaration = ""
+            if cg_signal.msg_parent.use_enum_choices:  # Generate choice instead of define
+                choices = _format_enum_choices(cg_signal, cg_signal.msg_parent.database_prefix +
```
trnila, line 1387: would be better to use f-strings instead of joining the string (or maybe `'_'.join([...])` would also work in this case)

src/cantools/database/can/c_source.py
```diff
@@ -1218,6 +1227,22 @@ def _format_choices(cg_signal: "CodeGenSignal", signal_name: str) -> list[str]:
     return choices
 
 
+def _format_enum_choices(cg_signal: "CodeGenSignal", signal_name: str) -> list[str]:
+    choices = []
+
+    for value, name in sorted(cg_signal.unique_choices.items()):
+        if cg_signal.signal.is_signed:
```
trnila, line 1234: we could use this to reduce code duplication:
```
suffix = '' if cg_signal.signal.is_signed else 'u'
fmt = '{signal_name}_{name} = ({value}{suffix})'
```

src/cantools/database/can/c_source.py
```diff
@@ -1357,14 +1382,23 @@ def _generate_choices_defines(database_name: str,
...
-                for choice in choices
-            ])
-            choices_defines.append(signal_choices_defines)
-
-    return '\n\n'.join(choices_defines)
+            signal_choices_declaration = ""
+            if cg_signal.msg_parent.use_enum_choices:  # Generate choice instead of define
+                choices = _format_enum_choices(cg_signal, cg_signal.msg_parent.database_prefix +
+                                               '_' + cg_signal.msg_parent.snake_name + '_' + cg_signal.snake_name)
+                signal_choices_declaration = 'typedef enum {' + (", ").join(
```
trnila, line 1389: would be nice to have each enum variant on single line for better clarity

src/cantools/subparsers/generate_c_source.py
```diff
@@ -55,7 +56,8 @@ def _do_generate_c_source(args):
         with open(fuzzer_path_c, 'w') as fout:
             fout.write(fuzzer_source)
 
-        fuzzer_path_mk = os.path.join(args.output_directory, fuzzer_filename_mk)
+        fuzzer_path_mk = os.path.join(
```
trnila, line 59: (would be good to minimize cosmetic/formatting changes)

trnila (commented):
Useful feature, but some tests should be added to ensure its not broken in the future.

Just thinking if there can be any issue with `enum` as the underlying datatype is `int`/compiler-specific? Like having 32bit enum signal, but compiler with `typedef int16_t int;`...

Ozuba:
> Useful feature, but some tests should be added to ensure its not broken in the future.
> 
> Just thinking if there can be any issue with `enum` as the underlying datatype is `int`/compiler-specific? Like having 32bit enum signal, but compiler with `typedef int16_t int;`...

Quick test tells me that gcc spits out a warning if having non representable integer constant
imagen
However i dont know if there would be any kind of realistic problem in a normal use case. Not at least any problem worse than a typical overflow.

src/cantools/database/can/c_source.py
Ozuba: This breaks the tests hardly for some newline characters, im trying to find a fix

src/cantools/subparsers/generate_c_source.py
Ozuba: Its autopep8 formatting

src/cantools/database/can/c_source.py
Ozuba: Implemented this in last commit

src/cantools/database/can/c_source.py
Ozuba: Implemented it like this on last commit

Ozuba:
proof of concept tests are written but unpacking to enum types results into -fpermissive errors, should i explitely add casts?

zariiii9003:
Please add some simple C test in `tests/test_c_code` with enum encoding/decoding.

Ozuba:
I added a basic encode decode test

tests/test_c_code/test_runner.c
```diff
@@ -57,5 +60,8 @@ int main(void) {
     RUN_TEST(absence_of_initial_value);
     RUN_TEST(nullptr_ptr_initial_value);
 
+    // test_enum.h
+    RUN_TEST(test_encode_decode);
```
zariiii9003, line 64: You're actually calling `test_encode_decode` from `test_basic.c` here. You need to make a few changes:

- use a more specific name like `test_enum_encode_decode`
- add the `test_enum.c` to `TEST_SOURCES` in `CMakeLists.txt`, otherwise you'll get a linker error
- i cannot find `choices_use_enum_choices.h` anywhere. You can generate a new file via `execute_process` in `CMakeLists.txt`

```CMake
execute_process(COMMAND cantools generate_c_source --database-name motohawk_use_enum_choices --use-enum-choices ../files/dbc/motohawk.dbc 
                WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
```

Ozuba: Thanks for reviewing, this is actually one of my  first pull requests, sory for the inexperience, should be fixed now i think!

zariiii9003: You forgot to replace the function name in `RUN_TEST(test_encode_decode);` 😄

tests/files/c_source/choices_use_enum_choices.h
```diff
@@ -0,0 +1,151 @@
...
+#define CHOICES_USE_ENUM_CHOICES_FOO_LENGTH (8u)
+
+/* Extended or standard frame types. */
+#define CHOICES_USE_ENUM_CHOICES_FOO_IS_EXTENDED (0)
+
+/* Frame cycle times in milliseconds. */
+
+
+/* Signal choices. */
+typedef enum {CHOICES_USE_ENUM_CHOICES_FOO_FOO_A_NEGATIVE_VALUE = (-5), CHOICES_USE_ENUM_CHOICES_FOO_FOO_WITH_SPACE = (0), CHOICES_USE_ENUM_CHOICES_FOO_FOO_______ = (1), CHOICES_USE_ENUM_CHOICES_FOO_FOO_UNUSED_2_ = (2), CHOICES_USE_ENUM_CHOICES_FOO_FOO_UNUSED_3 = (3), CHOICES_USE_ENUM_CHOICES_FOO_FOO_UNUSED_2 = (4), CHOICES_USE_ENUM_CHOICES_FOO_FOO_RESERVED_5 = (5), CHOICES_USE_ENUM_CHOICES_FOO_FOO_RESERVED_6 = (6)} choices_use_enum_choices_foo_foo_t;
```
zariiii9003, line 60: Could you improve the formatting and add line breaks?

Ozuba: You mean to have enums in this fashion:
```
typedef enum {
   BAR = 1,
   FOO = 3,
   BARFOO = 4
} barfoo_t;
```

tests/test_c_code/test_runner.c
Ozuba: Oh gosh i wasn't understanding you, there was a function already named like that!!
late hour, programming lol 🤣

tests/files/c_source/choices_use_enum_choices.h
Ozuba: Should be implemented by now

src/cantools/database/can/c_source.py
Ozuba: Finally implemented enums in multiline view for better clarity

Ozuba:
All should be good by now

andlaus:
@zariiii9003, @trnila: I defer the merge decision to you (in case this is not clear)

tests/files/c_source/choices_use_enum_choices.h
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * All signal values are as on the CAN bus.
+ */
+struct choices_use_enum_choices_foo_t {
+    /**
+     * Range: -128..127 (-128..127 -)
+     * Scale: 1
+     * Offset: 0
+     */
+    choices_use_enum_choices_foo_foo_t foo;
```
zariiii9003, line 88: I don't think the data type is correct. Signals can have enum values and a physical range at the same time. 
In this case, signal `foo` can have any value from -128 to 127. That's incompatible with the enum type 🤔

Ozuba: I why is that incompatible with enum type?, its just a signed integer underlaying no?

zariiii9003: But the enum type is usually int32 while the signal value could be uint64 or something like that?

trnila: Some signals are having defined enum values only for few signals (like `ERROR` or `INIT`) and the remaining are the values to be used after multiplication by constant and adding some offset...

In C++ we could use `std::variant<SignalEnumType, uint64_t>` to support some `uin64_t` measurement with special enum cases like `INIT, ERROR, PAUSED`...
In C++ we could also define enum underlying size `enum my_enum : uint64_t { VALUE1, VALUE2 };` - but thats not possible in C.

Maybe we can keep the signal datatype inside the struct and generate new function to return casted `enum` if its in the enum range? But it could be problematic for enum values that are not continuous...

Ozuba: Doing a bit of research it seems that both gcc and clang, decide the underlying type to fit all the tags in the enum
```
#include <stdio.h>
#include <stdint.h>

typedef enum {
    SMALL = 1,
    MEDIUM = 2147483647,   // INT_MAX (32-bit)
    LARGE = 9223372036854775807ULL  // 64-bit value, exceeds INT_MAX
} BigEnum;

int main() {
    // Check the size of the enum type
    printf("Size of BigEnum: %zu bytes\n", sizeof(BigEnum));
    
    // Check if the enum type has been allocated 64 bits
    BigEnum value = LARGE;
    printf("Value of LARGE: %lld\n", (long long)value);

    return 0;
}
```

This simple test on both gcc and clang shows that the underlying type is 64 bit
```
Size of BigEnum: 8 bytes
Value of LARGE: 9223372036854775807
```
Other compilers might not respect this as ive seen its not part of any C standard by now, but i think this feature would be useful in 99% of cases.

zariiii9003: The enum values might be 0 and 1, which results in a signed 32bit enum, while the signal data type is uint32 or a 64bit integer

Ozuba: If this is something to worry about, we could introduce a stub tag in order to force the enum to have a given size,
however, i mostly implemented this in order to get better integration of value tables. I didnt consider their usage as special values of a real signal. The entire cantools toolset is awesome but i miss some deeper integration in the c code generator such as automatic signal scaling and a universal decode function for a given subset of messages into a message struct. Mimicking the decoding functionality of its python usage, separating the memory representation of signals from the scaled/Value table representation. I think this pull request intent would fit better under that kind of approach.

olowo726:
@trnila "Some signals are having defined enum values only for few values (like ERROR or INIT) and the remaining are the values to be used after multiplication by constant and adding some offset..."

I think this is not too common. In the cases I've seen value table are not used for this purpose, instead comments are used to reserve a few values. Is it worth the effort to try to find a solution which both provides C enums and "normal" access? For me the C enum feature would be useful, it's a clean addon, and perhaps those few using both valuetables and normal signal values for the same signal will have to live without C enums for value tables, as they've done up to now.

trnila:
@olowo726 I have seen it multiple times - not sure if its permitted or it is just "misused". The advantage over comments is showing these labels in GUIs like canoe instead of showing some raw random constants. In rust it would actually make sense - and its much better then just comment:
```rust
enum SomeSignal {
    Init,
    Error,
    Value(u16)
}
```

Anyway it looks very nice and I think we could merge @zariiii9003? We can think about other use-cases of DBC once somebody will start using it...

zariiii9003:
> I think this is not too common. 

Very common. Looking through a few OEM dbc files: All signals, that represent some sensor value, contain at least init and fault values. Values like "sensor_not_installed" or various supply voltage faults are also very common.

In my opinion this is not a correct implementation yet.

PR #730, opened 2025-03-15
Fix ruff complaints

zariiii9003:


andlaus:
thanks. These new complaints seem a bit weird to me because the types which are being cast to ought to be known at the given location...

zariiii9003:
Here is the explanation for this rule: https://docs.astral.sh/ruff/rules/runtime-cast-value/

PR #731, opened 2025-03-18
fix(dump_string): remove unnecessary space to maintian proper format

detoxifiedplant:
remove unnecessary space to maintain proper format while dumping into string

### DIFF
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -663,13 +663,13 @@ def _dump_attribute_definitions(database: InternalDatabase) -> list[str]:
             choices = ','.join([f'"{choice}"'
                                 for choice in definition.choices])
             ba_def.append(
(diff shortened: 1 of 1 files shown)


juleq:
Hi.

Have you considered that these spaces might be in there for a purpose? Thats how Vector formats and for proper roundtrips, sticking to the Vector formatting is the way to go as they are authoritative for this format.

That's off the top of my head, I did not check with candb++, but you could do so if you want to see what I mean. Just open a dbc there and save it. Then do a roundtrip (load/dump) to a second file using cantools instead. Diff the two and they should match. Remove the spaces and they won't.

Best

detoxifiedplant:
@juleq Hi,
Thank you for the input and sorry for the late reply,
i am not aware of the purpose for trailing space, as I can't think of any, please enlighten me.
for Vector formats, it can be eliminated.
as i was implementing the dump and came across it while comparing with cantool's output [and standard J1939 or OBD].

Thank You for the awesome library 🥳

juleq:
Vector tools output it like that and Vector is the owner of the format.

PR #732, opened 2025-03-21
add extra_args for interface specific configuration arguments

miniehue:
Using the interface "socketcand" i found it is not possible to pass inerface specific arguments.

This suggestion is similar to the argument parsing of the can.viewer from the python-can package.
Interface specific arguments can be added and directly passed to the python-can bus configuration.

### DIFF
--- a/src/cantools/subparsers/__utils__.py
+++ b/src/cantools/subparsers/__utils__.py
@@ -237,7 +237,8 @@ def cast_from_string(string_val: str) -> Union[str, int, float, bool]:
 def parse_additional_config(unknown_args: Sequence[str]) -> TAdditionalCliArgs:
     for arg in unknown_args:
         if not re.match(r"^--[a-zA-Z][a-zA-Z0-9\-]*=\S*?$", arg):
-            raise ValueError(f"Parsing argument {arg} failed")
+            raise ValueError(f'Parsing argument {arg} failed, use --param=value'
+                             ' to pass additional args to CAN Bus.')
 
     def _split_arg(_arg: str) -> tuple[str, str]:
         left, right = _arg.split("=", 1)


src/cantools/subparsers/monitor.py
```diff
@@ -76,6 +77,9 @@ def create_bus(self, args):
         if args.fd:
             kwargs['fd'] = True
 
+        if hasattr(args, "extra_args") and args.extra_args:
```
trnila, line 80: is `hasattr(...)` really needed? If it was added due to failing tests, we should modify `Args` in `tests/test_monitor.py`.
Can we also add a test (`tests/test_monitor.py`) to verify these arguments are really passed to `Bus`?

Regarding `argparse.REMAINDER` - I hope there will be no name clash between monitor and `Bus`.
Other alternatives are:
- use `parse_known_args` and pass extra arguments after `--`, ie. `cantools monitor -c can0 bus.dbc -- --host=192.168.0.10 --port=29536`
- all arguments passed as single argument, ie. ` --bus_args "--host=192.168.0.10 --port=29536"`

miniehue: I removed hasattr(...) and added extra_args to Args in tests/test_monitor.py.,

I also added a test for parsing extra arguments.

I chose this way of passing extra arguments because users may be familiar with it from python-can.

zariiii9003:
Using private functions of python-can is a bad idea. I'd suggest copying the code from python-can instead.
Another option would be a change of the python-can API...

andlaus (approved):

miniehue:
Are further changes neccessary?

src/cantools/subparsers/__utils__.py
```diff
@@ -207,3 +210,40 @@ def format_multiplexed_name(message : Message,
...
+        return string_val.lower() == "true"
+
+    # value is string
+    return string_val
+
+
+def parse_additional_config(unknown_args: Sequence[str]) -> TAdditionalCliArgs:
+    for arg in unknown_args:
+        if not re.match(r"^--[a-zA-Z][a-zA-Z0-9\-]*=\S*?$", arg):
+            raise ValueError(f"Parsing argument {arg} failed")
```
trnila, line 240: can we write some hint please? something like `Parsing argument {arg} failed, use --param=value to pass additional args to CAN Bus`?

trnila (approved):

src/cantools/subparsers/__utils__.py
miniehue: Sure.

miniehue:
@trnila  Are you happy with tis PR ?

trnila:
I apologize for my late response.
I have just retested it and yes, its awesome - thank you! :)

PR #733, opened 2025-04-16
Fix frame collisions between extended and standard frames

### LINKED ISSUE #654: Frame IDs collisions between extended format frames can and standard format frames
I think CAN IDs are being handled incorrectly. I noticed this when converting a DBC file to C code and I think it applies to everything as the issue is in `cantools/database/can/database.py`

#  Reproduce issue
I have the following test case DBC file which seems valid and passes the consistency checks of CANdb++ (Other than receiver and sender nodes). I've attached it as a `.txt` because github doesn't attach `.dbc` files
[ExtenedMsgTest.dbc.txt](https://github.com/cantools/cantools/files/14876590/ExtenedMsgTest.dbc.txt)

Below is the command to convert it to C code:
`python3 -m cantools generate_c_source -o ./dbc ExtenedMsgTest.dbc`

I end up with the following output:
```
WARNING:cantools.database.can.database:Overwriting message 'New_Message_X001' with 'New_Message_001' in the frame id to message dictionary because they have identical masked frame ids 0x1.
Successfully generated ./dbc/extened_msg_test.h and ./dbc/extened_msg_test.c.
```
I don't think the warning is valid as the IDs are different in the DBC and both messages can appear on the CAN bus. This doesn't change the c code source generation, but could when trying to do other tasks.

The warning is raised due to the following code snippit:
```
masked_frame_id = (message.frame_id & self._frame_id_mask)

        if masked_frame_id in self._frame_id_to_message:
            LOGGER.warning(
                "Overwriting message '%s' with '%s' in the frame id to message "
                "dictionary because they have identical masked frame ids 0x%x.",
                self._frame_id_to_message[masked_frame_id].name,
                message.name,
                masked_frame_id)

        self._name_to_message[message.name] = message
        self._frame_id_to_message[masked_frame_id] = message
```
At the above point the the extended ID bit flag has been removed from the identifier, leaving only the base and extension identifier parts to form the frame_id. When look ups are done, for example in the `get_message_by_frame_id` function, the extended boolean flag is never put back in to the frame ID to form the complete identifier. This means collisions can occur even though the extended and standard frame formats are supposed to have separate _namespaces_.

This stack overflow answer has the relavent bits of the standard:
https://stackoverflow.com/a/63216382

Lauszus:
If an extended frame was having an ID that would be <= 0x7FF, then it would collide with the ID of a standard frame

This is fixed by setting the top bit i.e. OR'ing the frame by 0x80000000 similar to what is done in DBC files and in the Linux kernel

Fixes #654

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,7 +64,7 @@ bus using the `python-can`_ package.
    >>> import can
    >>> can_bus = can.interface.Bus('vcan0', bustype='socketcan')
    >>> data = example_message.encode({'Temperature': 250.1, 'AverageRadius': 3.2, 'Enable': 1})
-   >>> message = can.Message(arbitration_id=example_message.frame_id, data=data)
+   >>> message = can.Message(arbitration_id=example_message.frame_id, is_extended_id=example_message.is_extended_frame, data=data)
    >>> can_bus.send(message)
 
 Alternatively, a message can be encoded using the `encode_message()`_
--- a/examples/motor_tester/motor.py
+++ b/examples/motor_tester/motor.py
@@ -11,7 +11,7 @@ import can
 
 def create_message(speed, load):
     return can.Message(arbitration_id=0x010,
-                       extended_id=False,
+                       is_extended_id=False,
                        data=struct.pack('<HB', speed, load))
 
 
--- a/src/cantools/database/can/database.py
+++ b/src/cantools/database/can/database.py
@@ -472,12 +472,12 @@ class Database:
 
         return self._name_to_message[name]
 
-    def get_message_by_frame_id(self, frame_id: int, is_extended_frame: Optional[bool] = None) -> Message:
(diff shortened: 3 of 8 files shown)


zariiii9003:
Since `is_extended_frame=False` does nothing, maybe `force_extended_id` or something like that would be better? Then it can be False by default and doesn't have to be optional.

You could also edit one of the test dbc files or add a new one to include a small extended ID.

Lauszus:
> Since `is_extended_frame=False` does nothing, maybe `force_extended_id` or something like that would be better? Then it can be False by default and doesn't have to be optional.

Yeah that is better. I have renamed it now :)
 
> You could also edit one of the test dbc files or add a new one to include a small extended ID.

There is already a small extended ID in `attributes.dbc` i.e. 0x39 (2147483705 in the DBC file).

zariiii9003 (approved):

PR #740, opened 2025-06-08
Removed Special Chars for dbc

FedericoSpada:
When saving a dbc file, I've made sure that all Nodes, Messages and Signals names contain just valid characters, so that the file can be opened without Syntax Errors via the DBC Reader app.

I'm not sure if this problem also affects other Database formats, but unfortunately, I don't have the tools to test them.

### DIFF
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -438,6 +438,13 @@ def get_dbc_frame_id(message):
 
     return frame_id
 
+def get_dbc_name(name):
+    #replace special chars with "_"
+    name = re.sub(r"\W", "_", name)
+    #multiple "_" are condensated to just 1 "_"
+    name = re.sub(r"_+", "_", name)
+    return name
+
 
 def _get_node_name(attributes, name):
     try:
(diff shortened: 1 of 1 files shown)


andlaus:
just to further my own understanding: this is required if you want to convert a non-DBC dataset to DBC if the former features strings containing such characters?

@juleq: do you have any thoughts?

FedericoSpada:
I've created my own database format with additional features. Sometimes I have to save it in standard formats so that I can distribute it to other colleagues. To do so, I create a Database object and fill it with the info coming from my format. After that, I use dump_file(), which saves the file with the name and format requested by the user.

For the DBC file, I've encountered this problem where Vector Tools didn't recognise the file, so I thought to fix it.
I made the conversion always run because I can't imagine a situation where you wouldn't want to do it: if you open an existing DBC file and then re-save it, my changes have no effect. If instead you create a database with an invalid name in it, the saved DBC file is useless and (probably) you can't even re-open it with this same library.

juleq:
Could you provide a few names, before and after?

What happens if the user purposefully uses double underscores in his name?

FedericoSpada:
Any character that is not a-z, A-Z, 0-9 or "_" is replaced with "_". For example, I had a signal called "Motor Speed" and it was converted to "Motor_Speed". A message called "PDO3TX TRAC" was converted to "PDO3TX_TRAC (in my specific case, the only char that has been replaced was in fact the space).
Something more strange like "Hey, pal! How 'u doing?" would be converted to "Hey_pal_How_u_doing_"

The multiple underscore convertion is just optional, but I thought that it was nicer "Hey_pal_How_u_doing_" than "Hey__pal__How__u_doing_". As a matter of fact, we can remove it so that the double underscore can be used freely.

I think that I've tried other characters, but in the end it seemed that the names must be valid a Variable name as in most programming languages. I've also tried to use " as prefix and suffix that in other sections allows to specify any text, but it hasn't been accepted by Vector Tools.

andlaus:
> If you have no objections I would merge after.

go ahead :)

juleq:
Great, thanks. Can you make the linter happy? I'll merge then.

andlaus:
@FedericoSpada: most likely you don't even need to fix the `ruff` complaints manually: `ruff check --fix` will take care of the "obvious" fixes...

PR #742, opened 2025-07-20
Logreader: Parse more information and all CAN bus messsages + Bugfixes

erzoe:
New features:
- logreader: added `DataFrame.is_extended_frame`
  commit a57a7084441a67cd2759ffaf5cad33afa4b6bb01
- logreader: parse remote frames and added `DataFrame.is_remote_frame`
  commit 9636c98e20b2f458f3cdc79529bdc21d493423e0
- logreader: added optional timezone argument for Parser in case a user wants to parse the timestamps of CAN bus messages recorded in a different time zone
  commit d496bd6f15912d1e94dec954ac67bc829f83843b
- logreader: added `DataFrame.__repr__` method
  commit 289b4746790a3f32bd7c116ae21d0fa0d2301751
- logreader: more typing
  commit 9e968ade6ef44c3caae5dbc88d9e311534dc8a7f and c05b1a008705d4ef9b8e8541a6ab4dfadc2cbb46
- logreader: extended tests to make sure that all real CAN bus messages are now parsed
  commit 4e5859df6d9445dd687f97a4c4e539c92cfe32b8

Bugfixes:
- logreader: use system timezone instead of utc to parse timestamps by default
  commit 66a8d25c1a441af41be61e9df82b5774f6759fd6
- logreader: ignore error frames and warnings consistently (Before this commit the behavior was inconsistent.
Some formats parsed errors as DataFrame objects and some ignored errors because the regex did not match.
Now, no errors are parsed as DataFrame objects anymore. The raw lines indicating errors and warnings are still available via `iterlines(keep_unknowns=True)`.)
  commit 9acf8db4ebf15ba6c14c34a876affe15425ebab3
- logreader: more typing and fixed wrong typing of DataFrame.timestamp (`timestamp` was typed as `datetime.datetime`, although it can also be a `datetime.timedelta` or `None` - depending on `timestamp_format`.)
  commit b2ea03c36b36cf75c37ff3c2bb74c562633412fc
  
(These changes also include some refactorings of the internal pattern classes in order to implement the new features. They are now instantiated and inherit from eacht other, see commits d496bd6f15912d1e94dec954ac67bc829f83843b, f343f03d48ecefe8045a42eb91d20f6afefd065a and e67769324683baebc73ae2b523194195f5204165.)


erzoe:
I have seen that the static type checks with mypy have failed on Github because Github is using an older Python version (3.9). I am using TypeAlias which has been added in Python 3.10. On my computer mypy says "Success: no issues found in 61 source files" (Python 3.13.3, mypy 1.16.1).

The typing does not affect actually running the code. All dynamic tests have passed.

src/cantools/logreader.py
```diff
@@ -32,269 +42,244 @@ def __init__(self, channel: str,
...
-class PCANTracePatternV11(BasePattern):
+    def parse_data(self, match_object: 're.Match[str]') -> 'tuple[bytes, bool]':
+        data = match_object.group('can_data')
+        if data == 'RTR':
+            is_remote_frame = True
+            data = bytes(0)
+        else:
+            is_remote_frame = False
+            data = data.replace(' ', '')
+            data = binascii.unhexlify(data)
```
andlaus, line 199: probably this can be replaced by
```suggestion
            data = bytes.fromhex(data)
```

tox.ini
```diff
@@ -9,6 +9,7 @@ deps =
     coverage==7.3.*
     parameterized==0.9.*
     windows-curses; platform_system=="Windows" and platform_python_implementation=="CPython" and python_version<"3.13"
+    freezegun>=1.5.2,<1.6.0
```
andlaus, line 12: why is the freezegun module necessary? the patterns for the logreaders should not depend on the current date and time?!

andlaus (commented):
looks pretty good. besides the usual nits and the issue with type checking on python 3.9, I'm wondering why you used string-type type hints instead of the types directly (e.g. `'int|str|None'` instead of `int|str|None` or (for 3.9) `Optional[Union[str, int]]`)

erzoe:
@andlaus 

> I'm wondering why you used string-type type hints instead of the types directly (e.g. 'int|str|None' instead of int|str|None or (for 3.9) Optional[Union[str, int]])

To keep backward compatibility with older Python versions. In older versions using `int|str|None` without quotes raised an exception when trying to run the program. Therefore I have built up a habit of always quoting type hints if they are not trivial.

I dislike the syntax `Optional[Union[str, int]])` because `'int|str|None'` is easier to type and IMHO easier to read. I have used it, anyway, where ruff told me to remove the quotes around type hints.

andlaus:
> it could be typed in older versions but using TypeAlias (which has been added in Python 3.10) makes it easier and improves maintainability.

as @zariiii9003 notes, `TypeAlias` is available via `typing_extensions` in python 3.9. Also, I think it is possible to simply use an assignment, i.e., 
```python
foo = Union[float, int, str]
```
(maybe I forgot something here.)

> Keeping Python 3.9 compatibility in type checking does not seem worth it.

Let's agree to disagree here: IMO supporting older python versions is important in an industry as conservative as automotive. In particular, IMO we should support any version which still receives some upstream TLC...  

> Therefore I have built up a habit of always quoting type hints if they are not trivial.

I have a slight preference for the unquoted style precisely because it makes the python interpreter complain if a type was forgotten to be imported. not a big deal, though...

> I dislike the syntax Optional[Union[str, int]]) because 'int|str|None' is easier to type and IMHO easier to read.

I also prefer the new syntax, but the difference is IMO not bad enough to break 3.9 compatibility prematurely...

tox.ini
erzoe: Oh, now I see. I have missed something important there.

For generating the test cases I ran `candump -tA vcan0` and `candump -ta vcan0` and used the time stamps printed by `-tA` as the expected values for `-ta`, too, and the tests failed. What I failed to see was that the decoded time stamp contained the information that it was for a different timezone and that when it would have been converted to the correct time zone it would have given the correct result. So it was not wrong how it was before, just very surprising for me that I got the result for a different time zone than my own, especially considering that for `-tA` (which is tagged as the same TimestampFormat, namely ABSOLUTE) the result was for my own time zone.

Nevertheless I think we should not make a breaking change if the behavior was not wrong, especially because the new behavior would probably be wrong if the log was recorded in a different time zone than were it is parsed.

So I am thinking to change the default behavior back to returning a time stamp in UTC but keep the time zone argument to the parser. When explicitly passing `tz=None` the parser yields naive datetimes (as it currently does with this merge request). When passing a time zone different from UTC it converts the datetime from UTC to the desired timezone.

Thank you for your feedback.

erzoe:
```
foo = Union[float, int, str]
```

True, without quotes that works. I have fixed that.

But the other changes regarding the time zone I will unfortunately probably not be able to finish before my vacation.

erzoe:
@andlaus Thank you for your contributions. Better duplicate tests than missing tests.

Why do you want to avoid freezegun? Why should we decrease test coverage if we already have the tests? I think it is important to test as much as possible even if the functionality seems trivial. freezegun is only a testing dependency, it is nothing that users of the library need to install.

I totally agree now that keeping the possibility to run static type checking on older python versions is important. What I missed when I claimed that dropping support for static type checking in Python 3.9 would be acceptable was that this is a library and not a program. So type hints are important for the users as well, not just for the contributors. Static type checking works on Python 3.9 since I have removed the TypeAlias, even without your changes. But I am willing to accept your changes of the type hints if you prefer this syntax.

tox.ini
erzoe: The merge request as it is already complies with the desired behavior I have described in my previous comment when explicitly passing an argument for the time zone parameter. The only thing which would need to be changed is the default argument for the time zone parameter. But after some more thought I would like to propose to keep the change of the default.

I see three possibilities how changing the default from UTC to system default can affect users:
- The users were aware of getting the datetime object in UTC and converted it with `t.astimezone(datetime.timezone(datetime.timedelta(hours=...)))`. They will still get the exact same result after calling astimezone.
- The users were in UTC and did not need to convert the datetime object. They will still get the datetime object in UTC. There is a subtle difference though: The datetime object would now be naive. This does not make a difference for converting it with astimezone or formatting it (without %z) but it does make a difference for comparing it with other datetime objects. While I usually prefer naive datetime objects because they are the default and are what you get when calling `datetime.datetime.now()` I think that adding a conversion to a timezone-aware object with `t.astimezone()` would be better. Not only because this case would then yield the exact same result as before but also as a distinction to `candump -tA` where we return naive datetime objects which represent the (unknown) time zone where the log has been recorded.
- The users were not aware of getting the datetime object in UTC and used it as is outside of UTC. (This is the category I would have fallen into if I had not accidentally stumbled upon this). They were getting the datetime object for a wrong time zone and will now get the datetime object according to their locale which makes a lot more sense IMO. Converting to other time zones and comparing to other date time objects would still work the same like before (assuming we add a conversion to a timezone aware object) but formatting the object would give the result in the system time zone.

In my previous comment I voiced the concern that "the new behavior would probably be wrong if the log was recorded in a different time zone than were it is parsed." That was under the assumption that a user would want the datetime object for the time zone where the log was recorded. But returning the object for the system time zone seems like a good default, too. And in some cases this will be the same, anyway.

@andlaus @zariiii9003 What do you think? Is a breaking change acceptable here?

andlaus:
cool, thanks; let's get this merged :). (if you find further fixes, just open a new PR.)

PR #743, opened 2025-08-01
Make Exceptions Picklable

zariiii9003:
@andlaus unrelated, but the `from .can import *` should be removed. The comment says `# Remove once less users are using the old package structure` for seven years now


andlaus:
looks good. merging...

PR #744, opened 2025-08-01
Add cantools.db DeprecationWarning

zariiii9003:
I removed the star imports and added missing imports where tests required it.

I decided against removing `cantools.db` completely, after seeing how often it was used in our own tests.
Instead, a DeprecationWarning was added to warn against use of `cantools.db`


andlaus:
Neat! thanks for springing to action as quickly as a bolt of lightning :D

PR #754, opened 2025-09-29
fix(database): broken caching when loading dbc file

ckagerer:
Caching DBC files has not been working for me for some time now. I debugged it and found that it is incorrectly checking whether the cache object was created correctly.
This is probably due to a change in diskcache.
The problem only occurs when the cache is empty or has been newly created. Then the original query generates false and therefore never caches.

### DIFF
--- a/src/cantools/database/__init__.py
+++ b/src/cantools/database/__init__.py
@@ -132,7 +132,7 @@ def load_file(filename: StringPathLike,
     db: Union[can.Database, diagnostics.Database]
 
     with diskcache.Cache(cache_dir) if cache_dir else nullcontext() as cache:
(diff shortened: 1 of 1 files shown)


trnila (approved):
Good catch!
Since the `diskcache.Cache` doesnt implement `__bool__`, the `if cache` calls [`__len__`](https://github.com/grantjenks/python-diskcache/blob/ebfa37cd99d7ef716ec452ad8af4b4276a8e2233/diskcache/core.py#L2363) instead. And its `0` when theres no cache yet...

christian-hollmayer-horsch (approved):
Great!

PR #758, opened 2025-11-01
Update supported Python versions

ilanbiala:
* Removes support for Python 3.9 and adds support for Python 3.14. Python 3.9 is officially EOL as of Oct 31, 2025 (https://devguide.python.org/versions/#unsupported-versions).
* Bumps PyPy to 3.11 since that is the current supported version (Pillow doesn't build wheels for 3.10 now).
* Windows on PyPy still doesn't build, only due to a Matplotlib import error. Installing vcredist packages through `choco` didn't fix, and [`msvc-runtime`](https://pypi.org/project/msvc-runtime/) isn't available on PyPy.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -32,10 +32,10 @@ jobs:
     strategy:
       matrix:
         os: [ "ubuntu-latest", "windows-latest", "macos-latest" ]
-        python-version: ["3.10", "3.14", "pypy-3.10"]
+        python-version: ["3.10", "3.14", "pypy-3.11"]
         exclude:
           - os: windows-latest
-            python-version: pypy-3.10
+            python-version: pypy-3.11
       fail-fast: false
     steps:
     - uses: actions/checkout@v4


tox.ini
```diff
@@ -8,7 +8,7 @@ deps =
     pytest-cov==4.1.*
     coverage==7.3.*
     parameterized==0.9.*
-    windows-curses; platform_system=="Windows" and platform_python_implementation=="CPython" and python_version<"3.13"
+    windows-curses; platform_system=="Windows" and platform_python_implementation=="CPython" and python_version<"3.14"
```
ilanbiala: Doesn't yet support Python 3.14 (https://github.com/zephyrproject-rtos/windows-curses/issues/76)

ilanbiala:
@andlaus @zariiii9003 could you take a look please? I'm not sure how releases are handled, but this would likely necessitate at least a minor version bump.

tox.ini
andlaus: do you think it will be added within a reasonable time span? (<= 4-5 weeks)

andlaus:
okay, that sounds like we should merge to me :)

PR #759, opened 2025-11-17
Add support for CAN FD messages in Tester

### LINKED ISSUE #757: is_fd flag not passed to underlying can message
https://github.com/cantools/cantools/blob/5c193b1b98f462af84712de5c62ce533367bc3b7/src/cantools/tester.py#L248-L258

py-can message not passed the `is_fd` flag in the `_update_can_message` method which causes socketcan to not include the `CAN_FD_FRAME` tag on messages with fd-length data fields. the `is_fd` flag should be set equal to the `Message` object's `database.is_fd` flag.

ilanbiala:
Fixes #757

### DIFF
--- a/src/cantools/database/can/message.py
+++ b/src/cantools/database/can/message.py
@@ -274,7 +274,7 @@ class Message:
         self._is_extended_frame = value
 
     @property
-    def is_fd(self):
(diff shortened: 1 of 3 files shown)


andlaus:
looks good; let's merge! Thanks!

PR #767, opened 2025-12-15
Remove heavy-handed Mypy import ignore and replace with targeted ignores

ilanbiala:


ilanbiala:
Yeah, in this case this is mostly just shifting the ignores to specific lines. Generally I'd rather have less Mypy configuration though and let the tool run by itself. Eventually, I'd like to shift away from some of these packages as well, but that's further down the line.

PR #770, opened 2025-12-17
Print detailed database difference report in tests for easier debugging

zariiii9003:
I added a new test with some `# TODO` comments, some of which i will address in a follow-up PR.

### DIFF
--- a/src/cantools/database/can/database.py
+++ b/src/cantools/database/can/database.py
@@ -185,7 +185,8 @@ class Database:
                 add_diff(path_parts, type(a), type(b), 'type-mismatch')
                 return
 
-            if a is None:
+            if callable(a) or a is None:
+                # nothing to compare
                 return
 
             if isinstance(a, (int, str, set, bool)):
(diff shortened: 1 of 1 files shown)


src/cantools/database/can/database.py
```diff
@@ -146,86 +146,124 @@ def is_similar(self,
...
+
+                if name.startswith('_'):
+                    # skip private attributes
+                    continue
+
+                a_attr = getattr(a, name)
+                b_attr = getattr(b, name)
+                if callable(a_attr) or callable(b_attr):
+                    continue
+                if type(a_attr) is not type(b_attr):
```
andlaus: isn't that already covered by the if statement on line 184?

src/cantools/database/can/database.py
```diff
@@ -146,86 +146,124 @@ def is_similar(self,
...
+                    compare(a._signal_dict, b._signal_dict, ["_signal_dict"])
+                    continue
+
+                if name.startswith('_'):
+                    # skip private attributes
+                    continue
+
+                a_attr = getattr(a, name)
+                b_attr = getattr(b, name)
+                if callable(a_attr) or callable(b_attr):
```
andlaus: I suppose that if one of these attributes is callable, but the other is not, a `diff` should be added.

andlaus (approved):
LGTM

src/cantools/database/can/database.py
zariiii9003: Yes, that can be removed

PR #772, opened 2025-12-18
Fix long symbol handling for environment variables and signal groups

### LINKED ISSUE #766: Signals with SystemSignalLongSymbol getting removed from their signal group
Hi,
While loading a DBC file with signals that has the attribute SystemSignalLongSymbol, the signal is removed from its signal group once it's dumped to a new DBC file. This is due to the changes done in [#713 ](https://github.com/cantools/cantools/pull/713) making the signal names in the signal group to have long names instead of the short ones. I used the tests/files/dbc/long_names.dbc to run a script to load and dump the file and the signals with SystemSignalLongSymbol got removed from its signal group. Here is the difference between the two files:

Image

I ran the same script using the version before this PR #713 , and it works fine. After this commit, the behavior changes.
Is there any known workaround to resolve this?

zariiii9003:
Closes #766

### DIFF
--- a/src/cantools/database/can/environment_variable.py
+++ b/src/cantools/database/can/environment_variable.py
@@ -13,7 +13,8 @@ class EnvironmentVariable:
                  env_id,
                  access_type,
                  access_node,
-                 comment):
+                 comment,
+                 dbc_specifics=None):
         self._name = name
         self._env_type = env_type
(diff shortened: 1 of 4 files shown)


zariiii9003:
@andlaus last one :smile:

PR #783, opened 2026-03-21
feat: add is_in_phys_range() to generated C source

mon-martins:
Title: feat: add is_in_range_phys() to generated C source

Closes #781 
Related to #782 

## Solution

A new `is_in_range_phys()` function is generated alongside each signal's 
existing functions. It checks the physical value directly against the signal's 
defined min/max limits, before any encoding or casting occurs.

Generated output example (signal with scale=0.5, range=[-40, 87.5]):

    bool db_sensor_msg_temp_is_in_range_phys(double value)
    {
        return ((value >= -40.0) && (value <= 87.5));
    }

## Usage

    double value = get_sensor();

    if (!db_msg_speed_is_in_range_phys(value)) {
        handle_error();
        return;
    }

    msg.speed = db_msg_speed_encode(value);

## Changes

- `src/cantools/database/can/c_source.py`: added `SIGNAL_DECLARATION_IS_IN_RANGE_PHYS_FMT`, 
  `SIGNAL_DEFINITION_IS_IN_RANGE_PHYS_FMT` and `_generate_is_in_range_phys()`
- `tests/test_command_line.py`: added `test_generate_c_source_is_in_range_phys` 
  and `test_generate_c_source_is_in_range_phys_use_float`
- `tests/files/dbc/is_in_range_phys.dbc`: new fixture for the tests above
- `tests/files/c_source/`: reference files updated with `is_in_range_phys()` 
  additions (see #782 for context on why this update was necessary)


andlaus (commented):
looks good. I'd prefer calling it `is_in_phys_range()`, but I am willing to let myself be convinced for `is_in_range_phys()`.

PR #785, opened 2026-04-17
Add signal param to dump CLI tool.

KacperChmielewski:
I often use `cantools dump` to check one specific signal. To do so, I used to use `| less -p <mySignalName>`. 
Now you can use the `-s` or `--signal` flag.

Maybe there was a reason why this feature wasn't implemented earlier?

### DIFF
--- a/src/cantools/subparsers/dump/__init__.py
+++ b/src/cantools/subparsers/dump/__init__.py
@@ -104,12 +104,12 @@ def _dump_can_message(message, with_comments=False, name_prefix='', WIDTH=None):
                               name_prefix=f'{message.name} :: ')
 
 
-def _dump_can_signal(dbase: CanDatabase, signal_name: str, with_comments: bool) -> None:
+def _dump_can_single_message(dbase: CanDatabase, message_name: str, with_comments: bool) -> None:
     WIDTH = 80
     try:
-        message = dbase.get_message_by_name(signal_name)
+        message = dbase.get_message_by_name(message_name)
     except KeyError:
-        raise KeyError(f'Unknown signal {signal_name}') from KeyError
+        raise KeyError(f'Unknown message {message_name}') from KeyError
 
     print('================================= Message =================================')
     _dump_can_message(message,
@@ -161,12 +161,12 @@ def _do_dump(args):
                                encoding=args.encoding,
                                prune_choices=args.prune,
                                strict=not args.no_strict)
-    signal_name = args.signal
+    message_name = args.message
     if isinstance(dbase, CanDatabase):
-        if signal_name is None:
-            _dump_can_database(dbase, args.with_comments)
+        if message_name is not None:
+            _dump_can_single_message(dbase, message_name, args.with_comments)
         else:
-            _dump_can_signal(dbase, signal_name, args.with_comments)
+            _dump_can_database(dbase, args.with_comments)
     elif isinstance(dbase, DiagnosticsDatabase):
         _dump_diagnostics_database(dbase)
     else:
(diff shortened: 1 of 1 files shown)


src/cantools/subparsers/dump/__init__.py
```diff
@@ -102,6 +103,20 @@ def _dump_can_message(message, with_comments=False, name_prefix='', WIDTH=None):
                               WIDTH=WIDTH,
                               name_prefix=f'{message.name} :: ')
 
+
+def _dump_can_signal(dbase: CanDatabase, signal_name: str, with_comments: bool) -> None:
+    WIDTH = 80
+    try:
+        message = dbase.get_message_by_name(signal_name)
```
andlaus: this seems to be incorrect: it returns the CAN *frame* named `signal_name`, **not** the signal.

As I see it, you need to switch to a fundamentally different approach, because a signal with a given name can be featured by multiple messages...

andlaus (changes requested):
AFAICS this PR does not do what it says it does. that said, I'm not fundamentally opposed to this kind of functionality.

KacperChmielewski:
Damn, I mixed up "signal_name" with "message_name". Will fix this in a moment. Sorry for the confusion.

KacperChmielewski:
I guess it's ok now.

Below you can see example usage of this feature.

```
cantools dump tests/files/dbc/vehicle.dbc -m RT_SB_GPS_Speed
================================= Message =================================

  Name:           RT_SB_GPS_Speed
  Id:             0x9567c22
  Length:         8 bytes
  Cycle time:     - ms
  Senders:        -
  Layout:

                          Bit

             7   6   5   4   3   2   1   0
           +---+---+---+---+---+---+---+---+
         0 |   |   |   |   |   |   |<-x|<-x|
           +---+---+---+---+---+---+---+---+
                                     |   +-- Validity_GPS_Speed_2D
                                     +-- Validity_GPS_Speed_3D
           +---+---+---+---+---+---+---+---+
         1 |<-----------------------------x|
           +---+---+---+---+---+---+---+---+
             +-- Accuracy_GPS_Speed
           +---+---+---+---+---+---+---+---+
     B   2 |------------------------------x|
     y     +---+---+---+---+---+---+---+---+
     t   3 |-------------------------------|
     e     +---+---+---+---+---+---+---+---+
         4 |<------------------------------|
           +---+---+---+---+---+---+---+---+
             +-- GPS_Speed_2D
           +---+---+---+---+---+---+---+---+
         5 |------------------------------x|
           +---+---+---+---+---+---+---+---+
         6 |-------------------------------|
           +---+---+---+---+---+---+---+---+
         7 |<------------------------------|
           +---+---+---+---+---+---+---+---+
             +-- GPS_Speed_3D

  Signal tree:

    -- {root}
       +-- Validity_GPS_Speed_2D
       +-- Validity_GPS_Speed_3D
       +-- Accuracy_GPS_Speed
       +-- GPS_Speed_2D
       +-- GPS_Speed_3D

  ------------------------------------------------------------------------
```

```
cantools dump tests/files/dbc/vehicle.dbc -m RT_SB_GPS      
error: 'Unknown message RT_SB_GPS'
```

The thing I'm worried about is that atm there are 3 functions:
```
_dump_can_message # # original function to print single message within database
_dump_can_single_message  # newly created function to print only one message
_dump_can_database # original function to print the whole database
```

In my opinion it will look better if I rename these functions like this
```
_dump_can_message # -> _print_can_message
_dump_can_single_message # -> _dump_can_message
_dump_can_database # this one stays the same
```

I'm still not convinced. Do you have any thoughts on this?

andlaus:
you can make a single function for printing multiple messages, something like
```python
def _dump_can_messages(msg_list: Sequence[Message]) -> None:
    for msg in msg_list:
        ... # dump msg to stdout
```

for your feature, you need to filter externally, i.e.,
```python
_dump_can_messages([ database.get_message_by_name(args.message_name) ])

andlaus:
looks good now, merging. thanks for your contribution!

Issue #784, opened 2026-04-06
BA_REL_ signal keys not updated when long signal names are shortened


saeedsalih1997:
When a DBC file containing signals with names longer than 32 characters is loaded and dumped by cantools, the BA_REL_ entries (e.g. GenSigTimeoutTime) are written with stale signal name keys instead of the correct shortened names. This causes tools like CANdb++ to report "unknown symbol" errors when opening the output file.

<img width="387" height="84" alt="Image" src="https://github.com/user-attachments/assets/3e03fed6-7b78-4c2d-ad92-ffd49b948591" />

During export, [**make_signal_names_unique**](vscode-file://vscode-app/c:/Users/A508210/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in [**dbc.py**](vscode-file://vscode-app/c:/Users/A508210/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) renames each signal using [LongNamesConverter](vscode-file://vscode-app/c:/Users/A508210/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and stores the original full name in the SystemSignalLongSymbol attribute. However, the [attributes_rel](vscode-file://vscode-app/c:/Users/A508210/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) dictionary — which backs the BA_REL_ output in [_dump_attributes_rel](vscode-file://vscode-app/c:/Users/A508210/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — is still keyed by the old signal names (as they appeared in the input file). These old keys are never updated to match the new short names, so [_dump_attributes_rel](vscode-file://vscode-app/c:/Users/A508210/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) writes them as it is into the output DBC.

andlaus:
thanks for the analysis. Can you make a fix and open a pull request?

zariiii9003:
I created a test file:

<details>

[...]

PR #787, opened 2026-05-06
fix: preserve DBC signal short names and use global signal name uniqueness

saeedsalih1997:
Related to #784 
- Store original DBC short name on load so dump restores it exactly
- Use a single DBC wide LongNamesConverter instead of per message to ensure globally unique short names
- Update attributes_rel keys when signal/node names are shortened

### DIFF
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -411,10 +411,14 @@ class Parser(textparser.Parser):
 
 
 class LongNamesConverter:
-    def __init__(self, long_names: list[str]) -> None:
+    def __init__(self, long_names: list[str], reserved_short_names: set[str] | None = None) -> None:
 
         self.long_to_short: dict[str, str] = {}
         self.short_to_long: dict[str, str] = {}
+
+        if reserved_short_names:
+            for name in reserved_short_names:
+                self.short_to_long[name] = '\x00'
 
         for long_name in sorted(long_names, key=lambda s: (len(s), s)):
             short_name = long_name[:32]
@@ -1578,8 +1582,9 @@ def _load_signals(tokens,
     signals = []
 
     for signal in tokens:
+        token_short = signal[1][0]
         signals.append(
-            Signal(name=get_signal_name(frame_id_dbc, signal[1][0]),
+            Signal(name=get_signal_name(frame_id_dbc, token_short),
                    start=int(signal[3]),
                    length=int(signal[5]),
                    receivers=get_receivers(signal[20]),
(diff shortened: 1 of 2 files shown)


src/cantools/database/can/formats/dbc.py
```diff
@@ -411,11 +411,15 @@ def grammar(self):
...
 class LongNamesConverter:
-    def __init__(self, long_names: list[str]) -> None:
+    def __init__(self, long_names: list[str], reserved_short_names: set[str] | None = None) -> None:
 
         self.long_to_short: dict[str, str] = {}
         self.short_to_long: dict[str, str] = {}
 
+        if reserved_short_names:
+            for name in reserved_short_names:
+                self.short_to_long[name] = '\x00'
```
andlaus, line 421: I would be surprised if using a string containg a single null character as long name is a good idea (please convince me otherwise). maybe use `None` or an empty string instead?

src/cantools/database/can/formats/dbc.py
```diff
@@ -1578,35 +1582,44 @@ def get_signal_spn(frame_id_dbc, name):
     signals = []
 
     for signal in tokens:
+        token_short = signal[1][0]
```
andlaus, line 1585: I like that introducing a helper variable makes this more readable, but I don't like the `token_short` name. Maybe:
```suggestion
        signal_name = signal[1][0]
```

src/cantools/database/can/formats/dbc.py
```diff
@@ -1578,35 +1582,44 @@ def get_signal_spn(frame_id_dbc, name):
...
                    multiplexer_signal=get_multiplexer_signal(signal[1],
                                                              multiplexer_signal)))
+        # If the original DBC file shortened this signal, store the original short name as a
+        # hidden Python attribute (not in dbc.attributes, so never written to output).
+        # This lets make_signal_names_unique preserve the original short name in DBC
+        created = signals[-1]
+        if created.name != token_short:
+            if created.dbc is None:
+                created.dbc = DbcSpecifics()
+            created.dbc._orig_dbc_short_name = (created.name, token_short)
```
andlaus, line 1622: the '.dbc._orig_dbc_short_name` should be always present IMO, i.e.,
```suggestion
        created_signal = signals[-1]
        if created_signal.dbc is None:
            created_signal.dbc = DbcSpecifics()
        created_signal.dbc._orig_short_name = token_short
```

src/cantools/database/can/formats/dbc.py
```diff
@@ -1959,6 +1972,14 @@ def make_node_names_unique(database: InternalDatabase, shorten_long_names: bool)
             get_long_node_name_attribute_definition(database))
         node.name = short_name
 
+        # Update attributes_rel: rename node_name keys from long_name to short_name
+        if database.dbc is not None and database.dbc.attributes_rel is not None:
+            for element in database.dbc.attributes_rel.values():
+                if 'signal' in element:
+                    for signal_lst in element['signal'].values():
```
andlaus, line 1979:
```suggestion
                    for signal_list in element['signal'].values():
```

src/cantools/database/can/formats/dbc.py
```diff
@@ -1981,11 +2002,54 @@ def make_message_names_unique(database: InternalDatabase, shorten_long_names: bo
...
+    # Collect short names already locked in by existing signals so the
+    # converter never generates a duplicate for a newly added signal.
+    reserved_short_names = {
+        sig.dbc._orig_dbc_short_name[1]
+        for message in database.messages
+        for sig in message.signals
+        if sig.dbc is not None
+            and sig.dbc._orig_dbc_short_name is not None
+            and sig.dbc._orig_dbc_short_name[0] == sig.name
+    }
```
andlaus, line 2025: with the above suggestion this would become
```suggestion
    reserved_short_names = {
        sig.dbc._orig_short_name
        for message in database.messages
        for sig in message.signals
        if sig.dbc is not None
            and sig.dbc._orig_short_name != sig.name
    }
```

andlaus (commented):
With my limited understanding of the DBC format, the code looks okay to me. That said, I'm not sure why this is needed: if you want to dump exactly the same names as in the original file, wouldn't you just pass `shorten_long_names=False` to `dump_string()`?

PR #788, opened 2026-05-11
Add typing for src/cantools/database/... and other related areas

ilanbiala:
Some areas that raised errors have been ignored for now to allow this improvement to move forward.

### DIFF
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -1,10 +1,10 @@
 # Load and dump a CAN database in DBC format.
 
 import re
+import typing
 from collections import OrderedDict, defaultdict
 from copy import deepcopy
 from decimal import Decimal
-from typing import cast
 
 import textparser  # type: ignore
 from textparser import (
@@ -23,6 +23,7 @@ from textparser import (
 )
 
 from cantools.database.can.internal_database import InternalDatabase
+from cantools.typechecking import Comments
 
 from ...conversion import BaseConversion
 from ...namedsignalvalue import NamedSignalValue
@@ -30,12 +31,16 @@ from ...utils import (
     SORT_SIGNALS_DEFAULT,
     sort_signals_by_start_bit,
     sort_signals_by_start_bit_reversed,
+    type_sort_attribute,
     type_sort_attributes,
     type_sort_choices,
     type_sort_signals,
 )
-from ..attribute import Attribute
-from ..attribute_definition import AttributeDefinition, AttributeDefinitionType
+from ..attribute import Attribute, AttributeType
+from ..attribute_definition import (
+    AttributeDefinition,
+    AttributeDefinitionType,
+)
 from ..bus import Bus
 from ..environment_variable import EnvironmentVariable
 from ..message import Message
@@ -177,17 +182,17 @@ ATTRIBUTE_DEFINITION_GENMSGCYCLETIME = AttributeDefinition(
 
 ATTRIBUTE_DEFINITION_GENSIGSTARTVALUE = AttributeDefinition(
     name='GenSigStartValue',
-    default_value=0,
+    default_value=0.0,
     kind='SG_',
     type_name='FLOAT',
     minimum=0,
     maximum=100000000000)
 
 
-def to_int(value):
+def to_int(value: typing.Any) -> int:
     return int(Decimal(value))
 
-def to_float(value):
+def to_float(value: typing.Any) -> float:
     return float(Decimal(value))
 
 def _get_value(definition: AttributeDefinitionType, value: int | float | None) -> str:
@@ -198,6 +203,14 @@ def _get_minimum(definition: AttributeDefinitionType) -> str:
 
 def _get_maximum(definition: AttributeDefinitionType) -> str:
     return _get_value(definition, definition.maximum)
+
+def _get_attribute_value(attribute: AttributeType) -> str | int | float:
+        result = attribute.value
+
+        if attribute.definition.type_name == "STRING":
+            result = f'"{attribute.value}"'
+
+        return result
 
 class Parser(textparser.Parser):
 
@@ -438,7 +451,7 @@ class LongNamesConverter:
             self.short_to_long[short_name] = long_name
 
 
-def get_dbc_frame_id(message):
+def get_dbc_frame_id(message: Message) -> int:
     frame_id = message.frame_id
 
     if message.is_extended_frame:
@@ -446,7 +459,7 @@ def get_dbc_frame_id(message):
 
     return frame_id
 
-def get_dbc_name(name):
+def get_dbc_name(name: str) -> str:
     #replace special chars with '_'
     name = re.sub(r'\W', '_', name)
     #append '_' if it starts with a number
@@ -739,45 +752,29 @@ def _dump_attribute_definition_defaults_rel(database):
     return ba_def_def_rel
 
 
-def _dump_attributes(database, sort_signals, sort_attributes):
-    attributes = []
+def _dump_attributes(database: InternalDatabase, sort_signals: type_sort_signals, sort_attributes: type_sort_attributes) -> list[str]:
+    attributes: list[type_sort_attribute] = []
 
-    def get_value(attribute):
-        result = attribute.value
+    for attribute in database.dbc.attributes.values():
+        attributes.append(('dbc', attribute, None, None, None, None))
 
-        if attribute.definition.type_name == "STRING":
-            result = f'"{attribute.value}"'
-
-        return result
-
-    if database.dbc is not None:
-        if database.dbc.attributes is not None:
-            for attribute in database.dbc.attributes.values():
-                attributes.append(('dbc', attribute, None, None, None, None))
-
-        for envvar in database.dbc.environment_variables.values():
-            if envvar.dbc is not None:
-                if envvar.dbc.attributes is not None:
-                    for attribute in envvar.dbc.attributes.values():
-                        attributes.append(('envvar', attribute, None, None, None, envvar))
+    for envvar in database.dbc.environment_variables.values():
+        for attribute in envvar.dbc.attributes.values():
+            attributes.append(('envvar', attribute, None, None, None, envvar))
 
     for node in database.nodes:
-        if node.dbc is not None:
-            if node.dbc.attributes is not None:
-                for attribute in node.dbc.attributes.values():
-                    attributes.append(('node', attribute, node, None, None, None))
+        for attribute in node.dbc.attributes.values():
+            attributes.append(('node', attribute, node, None, None, None))
 
     for message in database.messages:
         # retrieve the ordered dictionary of message attributes
-        msg_attributes = OrderedDict()
-        if message.dbc is not None and message.dbc.attributes is not None:
-            msg_attributes.update(message.dbc.attributes)
+        msg_attributes = message.dbc.attributes
 
         # synchronize the attribute for the message cycle time with
         # the cycle time specified by the message object
-        gen_msg_cycle_time_def: AttributeDefinition  # type: ignore[annotation-unchecked]
+        gen_msg_cycle_time_def: AttributeDefinition[int] | None
         msg_cycle_time = message.cycle_time or 0
-        if gen_msg_cycle_time_def := database.dbc.attribute_definitions.get("GenMsgCycleTime"):
+        if gen_msg_cycle_time_def := typing.cast("AttributeDefinition[int] | None", database.dbc.attribute_definitions.get("GenMsgCycleTime")):
             if msg_cycle_time != gen_msg_cycle_time_def.default_value:
                 msg_attributes['GenMsgCycleTime'] = Attribute(
                     value=msg_cycle_time,
@@ -789,7 +786,7 @@ def _dump_attributes(database, sort_signals, sort_attributes):
             del msg_attributes['GenMsgCycleTime']
 
         # if bus is CAN FD, set VFrameFormat
-        v_frame_format_def: AttributeDefinition  # type: ignore[annotation-unchecked]
+        v_frame_format_def: AttributeDefinitionType | None
         if v_frame_format_def := database.dbc.attribute_definitions.get("VFrameFormat"):
             if message.protocol == 'j1939':
                 v_frame_format_str = 'J1939PG'
@@ -801,14 +798,14 @@ def _dump_attributes(database, sort_signals, sort_attributes):
                 v_frame_format_str = 'ExtendedCAN'
             else:
                 v_frame_format_str = 'StandardCAN'
-            v_frame_format_def = _get_enum_vframeformat_attribute(v_frame_format_def)
+            v_frame_format_def = _get_enum_vframeformat_attribute(v_frame_format_def)  # type: ignore
             # only set the VFrameFormat if it valid according to the attribute definition
             if (
                 v_frame_format_str in v_frame_format_def.choices
                 and v_frame_format_str != v_frame_format_def.default_value
             ):
                 msg_attributes['VFrameFormat'] = Attribute(
-                    value=v_frame_format_def.choices.index(v_frame_format_str),
+                    value=str(v_frame_format_def.choices.index(v_frame_format_str)),
                     definition=v_frame_format_def,
                 )
 
@@ -818,15 +815,13 @@ def _dump_attributes(database, sort_signals, sort_attributes):
             attributes.append(('message', attribute, None, message, None, None))
 
         # handle the signals contained in the message
-        if sort_signals:
+        if callable(sort_signals):
             signals = sort_signals(message.signals)
         else:
             signals = message.signals
         for signal in signals:
             # retrieve the ordered dictionary of signal attributes
-            sig_attributes = OrderedDict()
-            if signal.dbc is not None and signal.dbc.attributes is not None:
-                sig_attributes = signal.dbc.attributes
+            sig_attributes = signal.dbc.attributes
 
             # synchronize the attribute for the signal start value with
             # the start value specified by the message object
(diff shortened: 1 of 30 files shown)


src/cantools/database/can/formats/dbc.py
```diff
@@ -2103,8 +2075,8 @@ def dump_string(database: InternalDatabase,
                           sig_mux_values='\r\n'.join(sig_mux_values))
 
 
-def get_definitions_dict(definitions, defaults):
-    result = OrderedDict()
+def get_definitions_dict(definitions, defaults) -> OrderedDict[str, AttributeDefinitionType]:  # type: ignore[no-untyped-def]
```
andlaus: in the worst case, `Any` can be used as the argument types to get rid of the `# type: ignore`. That said, I'm pretty sure that something more specific can be found...

src/cantools/database/can/message.py
```diff
@@ -110,7 +112,7 @@ def __init__(self,
         else:
             self._signals = signals
         self._signal_dict: dict[str, Signal] = {}
-        self._contained_messages = contained_messages
```
andlaus, line 113: I'm not so sure if we really want do this: if `.contained_messages` is `None`, it currently means that the message is an ordinary message, whilst if it is an empty list, it means that it is a container message without any potentially contained children. I guess the distinction is not really relevant in practice, though....

src/cantools/database/can/message.py
```diff
@@ -128,12 +130,12 @@ def __init__(self,
         self._senders = senders or []
         self._send_type = send_type
         self._cycle_time = cycle_time
-        self._dbc = dbc_specifics
-        self._autosar = autosar_specifics
+        self._dbc = dbc_specifics or DbcSpecifics()
```
andlaus: IMO this is a thing which we almost certainly should not do: `self._dbc` should only be available if the database was created from a DBC file (same for the object for autosar specifics)

src/cantools/database/can/message.py
```diff
@@ -547,7 +547,7 @@ def gather_signals(self,
         for mux_signal_name, mux_nodes in node['multiplexers'].items():
             mux_num = self._get_mux_number(input_data, mux_signal_name)
             mux_node = mux_nodes.get(mux_num)
-            if mux_num is None or mux_node is None:
```
andlaus, line 550: nice catch!

src/cantools/database/can/message.py
```diff
@@ -1178,7 +1178,7 @@ def decode_container(self,
     def get_contained_message_by_header_id(self, header_id: int) \
         -> Optional['Message']:
 
-        if self.contained_messages is None:
+        if len(self.contained_messages) == 0:
```
andlaus: if we decide to keep `.contained_messages` to be always a list, this `if` statement is not necessary...

src/cantools/database/can/message.py
```diff
@@ -1194,7 +1194,7 @@ def get_contained_message_by_header_id(self, header_id: int) \
     def get_contained_message_by_name(self, name: str) \
         -> Optional['Message']:
 
-        if self.contained_messages is None:
+        if len(self.contained_messages) == 0:
```
andlaus: ditto

src/cantools/database/can/message.py
```diff
@@ -1227,17 +1227,17 @@ def is_multiplexed(self) -> bool:
...
         return bool(self._codecs['multiplexers'])
 
-    def _check_signal(self, message_bits, signal):
-        signal_bits = signal.length * [signal.name]
+    def _check_signal(self, message_bits: MutableSequence[str | None], signal: Signal) -> None:
+        signal_bits: list[str | None] = cast('list[str | None]', signal.length * [signal.name])
 
         if signal.byte_order == 'big_endian':
-            padding = start_bit(signal) * [None]
+            padding = cast('list[str | None]', start_bit(signal) * [None])
```
andlaus, line 1230: shouldn't that be 

```suggestion
            padding = cast(list[None], start_bit(signal) * [None])
```

also, is the cast really necessary? `start_bit()` returns `int` after all...

(same for the other casts below)

src/cantools/database/can/message.py
```diff
@@ -1270,21 +1270,21 @@ def _check_mux(self, message_bits, mux):
         for multiplexer_id in sorted(children):
             child_tree = children[multiplexer_id]
             child_message_bits = deepcopy(children_message_bits)
-            self._check_signal_tree(child_message_bits, child_tree)
+            self._check_signal_tree(child_message_bits, cast('SignalTreeType', child_tree))
```
andlaus: hm, this seems a bit weird to me: why  is `child_message_bits` not of `SignalTreeType`?

src/cantools/database/can/node.py
```diff
@@ -29,14 +29,14 @@ def __init__(self,
         self._comments: Comments | None
         if isinstance(comment, str):
             # use the first comment in the dictionary as "The" comment
-            self._comments = {None: comment}
+            self._comments = cast("Comments", {None: comment})
```
andlaus: why is the cast required here? (`Comments` is defined as `dict[str | None, str]`)

tests/test_list.py
```diff
@@ -58,6 +58,8 @@ def test_dbc(self):
   Size: 8 bytes
   Is extended frame: False
   Is CAN-FD frame: False
+  Is network management frame: False
```
andlaus: this is due to the fact that with your changes, the `.autosar` object is always available, even if the database was loaded from a file format which does not know about the concept of "network management" frames...

tests/test_list.py
```diff
@@ -144,6 +146,7 @@ def test_arxml3(self):
             self.assertEqual(actual_output, expected_output)
 
     def test_arxml4(self):
+        self.maxDiff = None
```
andlaus: that seems like a debugging artifact

andlaus (commented):
thanks, this looks pretty good!

Besides the usual nits, the only two things I'm not convinced about is the approach to replace `None` objects with empty lists (`.contained_messages`, ...) and to always have format-specific objects for every format irrespective of which format was used to load the database. (I'm open for good arguments besides the fact that it reduces the necessary legwork...)

andlaus (commented):
(for browser tab reasons, github did not include all comments in my last review)

src/cantools/database/can/formats/dbc.py
ilanbiala: I can add `Any`, but it doesn't actually provide additional info to users, and I think it becomes a bit harder to track whether that's actually acceptable usage or just a placeholder type. I don't have a strong preference here though.

src/cantools/database/can/message.py
ilanbiala: If there is a use case for a container message without any children, I agree. From what I've seen though, that isn't a use case I've found online (or at least there isn't a test case for it), so I think this is a worthwhile change.

src/cantools/database/can/node.py
ilanbiala: Because the literal `{None: comment}` is not typed to that extent

tests/test_list.py
ilanbiala: Agreed, will remove

tests/test_list.py
ilanbiala: Correct, but this list output seems like a negligible difference compared to being able to simplify the library implementation?

src/cantools/database/can/formats/dbc.py
andlaus: I would strongly prefer to find out the "real" types which these parameters can assume, but if this is too much effort I slightly prefer `Any` over `# type: ignore` because the latter has occasionally exploded on me in the past when a new mypy version renamed the error name (for `no-untyped-def` I don't believe it will happen, but still...)

src/cantools/database/can/message.py
ilanbiala: How/where is `Message._dbc` used to determine if the database was created from a DBC? And why would that not be possible to support with a `DbcSpecifics.filename` and/or `DbcSpecifics.empty` attribute added to the class? Generally, having optionally instantiated variables makes implementation behavior quite messy rather than just adding a flag to know whether the instance is valid.

src/cantools/database/can/message.py
andlaus: I don't think there's any practical use case, but autosar allows it, so it must be found somewhere in the wild. For container messages I think that the "empty list means normal message" is fine...

src/cantools/database/can/formats/dbc.py
ilanbiala: I'm fine with `Any`. And agreed, I generally aim for the real types as much as possible. In some cases these are quite tedious to accurately find, and also end up not providing much useful info `dict[str, dict[str, ... ]]]]]`, so I'm hoping after we get quite a bit of typing we can look at some refactors.

src/cantools/database/can/node.py
andlaus: I see. the problem is that `Comment` has been changed to `OrderedDict` from `dict`? how about `self._comments = Comments( [(None, comment)] )` ?

src/cantools/database/can/message.py
ilanbiala: Understood, and I think `Message.is_container()` still supports what's needed?

src/cantools/database/can/internal_database.py
```diff
@@ -24,5 +24,5 @@ def __init__(self,
         self.nodes = nodes
         self.buses = buses
         self.version = version
-        self.dbc = dbc_specifics
```
andlaus, line 27: this is probably not a good idea, because if the format specifics attribute for every format is always available leads to these objects to be used without them actually being defined (see the additional "is network management message" lines in output of `test_list` below...

ilanbiala: Duplicate of https://github.com/cantools/cantools/pull/788#discussion_r3218537534?

tests/test_list.py
andlaus: my main issue is more that if the user wants to access format specifics, she should explicitly test for them, e.g.

```python
if message.dbc is not None:
    ...
elif message.autosar is not None:
    ...
```

If I did not get this wrong such a test is basically impossible with the PR in its current form?

src/cantools/database/can/node.py
ilanbiala: `dict` vs `OrderedDict` shouldn't matter here (I switched to `OrderedDict` to maintain consistency with other data I/O processing in the library). `{None: comment}` as a literal will be inferred as either having a dict with literals, or still `dict[None, str]` which isn't broad enough for the key. I'll give the constructor a try though, that might work.

src/cantools/database/can/message.py
andlaus: yeah. the practical difference is negligible: Slightly "less correct in a corner case" but quite a bit simpler to (mis)use. IMO your argument that the tradeoff is worthwhile is valid...

src/cantools/database/can/message.py
andlaus: > How/where is Message._dbc used to determine if the database was created from a DBC?

I use `msg.[dbc|autosar]` quite often in my application code which has to work with both ARXML and DBC files... (mainly for autosar stuff like E2E protection, SecOC, etc.)

> And why would that not be possible to support with a DbcSpecifics.filename and/or DbcSpecifics.empty attribute added to the class?

because letting these objects become `None` is cleaner and simpler ;)

> Generally, having optionally instantiated variables makes implementation behavior quite messy rather than just adding a flag to know whether the instance is valid.

agreed but if you access a format specific your code should be explicit about it, so in this case I would like to have an `if` around that code.

src/cantools/database/can/internal_database.py
andlaus: yes, sorry...

src/cantools/database/can/formats/dbc.py
andlaus: agreed. Thanks for your effort, BTW :)

src/cantools/database/can/message.py
ilanbiala: Type hints did all the heavy lifting!

src/cantools/database/can/message.py
ilanbiala: Unfortunately, I think so:
```
src/cantools/database/can/message.py:1234: error: Incompatible types in assignment (expression has type "list[None]", variable has type "list[str | None]")  [assignment]
src/cantools/database/can/message.py:1234: note: "list" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
src/cantools/database/can/message.py:1234: note: Consider using "Sequence" instead, which is covariant
```
`Sequence[]` doesn't support `+`-style list operations, so it can't be used as a workaround. Setting the initial type to something like `list[str | None] | list[None]` might be a fix, but that looks awfully ugly. Let me know if you have any better suggestions than leaving the `cast()` statements.

src/cantools/database/can/message.py
ilanbiala: It's not `child_message_bits` that is getting casted, it is `child_tree`. Do you still see an issue?

tests/test_list.py
ilanbiala: Yes, we'd add a new attribute so the conditional could be something like:
```
if not message.dbc.empty:
    ...
elif not message.autosar.empty:
    ...
```
And e.g. `DbcSpecifics.empty` / `AutosarSpecifics.empty` would be set in the class `__init__()` if all of the arguments are `None`.

Thoughts?

ilanbiala:
@andlaus everything is passing here now, so if you'd be able to take a look at the open comments and see if you have additional feedback. Otherwise, I think this is overall a big step in the right direction.

I also haven't squashed subsequent commits yet, so you can take a look at those diffs if you see anything new that's unexpected. I'll squash down to one commit and merge after you approve.

src/cantools/database/can/message.py
andlaus: its not really a technical one, but I think that casts are pretty ugly and somewhat distracting and am surprised that this cast is necessary in the first place. That said, if there's no better alternative, I don't mind keeping the cast...

tests/test_list.py
andlaus: simply keeping the status-quo and letting `message.[dbc|autosar]` become `None` has the same effect and is more fool proof because users will not access non-existing properties. Further, the proposed approach would break a ton of existing user code because there ought to be heaps of `if msg.autosar` statements which would all break,  and finally `mypy` can complain about accessing potential `None` objects and IMO we should not worry too much about users who don't want to run `mypy` on their stuff.

andlaus:
> @andlaus everything is passing here now, so if you'd be able to take a look at the open comments and see if you have additional feedback. Otherwise, I think this is overall a big step in the right direction.

I *really* want to keep the current approach to the format-specific property objects (i.e., setting them to `None` if they do not apply) because changing this would IMO constitute a major API break. Besides that, this PR looks great :)

> I also haven't squashed subsequent commits yet, so you can take a look at those diffs if you see anything new that's unexpected. I'll squash down to one commit and merge after you approve.

IMO squashing them is not necessary, but if you prefer to do so, I don't mind either...

src/cantools/database/can/formats/dbc_specifics.py
```diff
@@ -34,9 +34,9 @@ def attributes(self) -> OrderedDict[str, AttributeType]:
 
         return self._attributes
 
-    @attributes.setter
```
andlaus, line 49: is there a reason why you removed the setter? if yes, it is IMO better to remove the code completely instead of commenting it out...

ilanbiala:
> I really want to keep the current approach to the format-specific property objects (i.e., setting them to None if they do not apply) because changing this would IMO constitute a major API break. Besides that, this PR looks great :)

Yeah, this would be a major version bump, which we should feel free to to do in order to simplify the internal library quite a bit. As it stands now, users would already be doing:
```
if msg.dbc:
    ...
```
so I don't see a reason not to simplify things internally and just add a release note that says to migrate such calls to:
```
if not msg.dbc.empty:
    ...
```

A big driver of adding typing to the library is to uncover exactly these areas where the library's functionality is not very well defined, working with uninitialized data, or otherwise could be improved and documented through typing.

src/cantools/database/can/formats/dbc_specifics.py
ilanbiala: Yes, I was checking and testing within the library. This setter isn't used anywhere, and it's the only setter specified. Either we should have all setters defined or no setters, so I'm removing.

src/cantools/database/can/message.py
ilanbiala: I managed to simplify this quite a bit, but overall `Message._check_mux()` has room to be rewritten and simplified

andlaus:
> Yeah, this would be a major version bump, which we should feel free to to do in order to simplify the internal library quite a bit.

I agree that major version bumps coming with some pain for users are in the cards, but IMO only if there is a major technical reason beyond "let's make things slightly simpler internally" (which I don't even fully agree with in this case, because the proposed change caused some minor bugs, cf. the changed output of the unit tests.)

andlaus:
I made a PR for your branch which fixes the issue with the format specifics: https://github.com/ilanbiala/cantools/pull/2

after this (or something equivalent) is included, this PR will be merged.

andlaus:
@ilanbiala: what's the status? are you going to merge https://github.com/ilanbiala/cantools/pull/2 ?

andlaus:
@ilanbiala: If I won't hear back from you, I'll merge this PR tomorrow and open https://github.com/ilanbiala/cantools/pull/2 as a separate top-level PR to go in immediately thereafter.

ilanbiala:
Sorry, was working on some personal things, merged your PR into my fork and merging here.

andlaus:
no worries. thanks for merging :)

PR #790, opened 2026-06-01
monitor: remove the `-f` alias for `--filter-regex`

andlaus:
on my machine, this cause conflicts with the `--fd` option:

```
  File "cantools", line 6, in <module>
    sys.exit(_main())
             ~~~~~^^
  File "cantools/src/cantools/__init__.py", line 102, in _main
    _load_subparser(subparser_name, subparsers)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cantools/src/cantools/__init__.py", line 67, in _load_subparser
    result.add_subparser(subparsers)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "cantools/src/cantools/subparsers/monitor.py", line 663, in add_subparser
    monitor_parser.add_argument(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        '-f', '--fd',
        ^^^^^^^^^^^^^
        action='store_true',
        ^^^^^^^^^^^^^^^^^^^^
        help='(Deprecated) Python CAN CAN-FD bus.')
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/argparse.py", line 1517, in add_argument
    return self._add_action(action)
           ~~~~~~~~~~~~~~~~^^^^^^^^
  File "/usr/lib64/python3.13/argparse.py", line 1906, in _add_action
    self._optionals._add_action(action)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/usr/lib64/python3.13/argparse.py", line 1724, in _add_action
    action = super(_ArgumentGroup, self)._add_action(action)
  File "/usr/lib64/python3.13/argparse.py", line 1531, in _add_action
    self._check_conflict(action)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/usr/lib64/python3.13/argparse.py", line 1673, in _check_conflict
    conflict_handler(action, confl_optionals)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/argparse.py", line 1682, in _handle_conflict_error
    raise ArgumentError(action, message % conflict_string)
argparse.ArgumentError: argument -f/--fd: conflicting option string: -f
```

### DIFF
--- a/src/cantools/subparsers/monitor.py
+++ b/src/cantools/subparsers/monitor.py
@@ -644,7 +644,7 @@ def add_subparser(subparsers):
         action='store_true',
         help='Skip database consistency checks.')
     monitor_parser.add_argument(
-        '-f', '--filter-regex',
+        '--filter-regex',
         help='Use given filter regex.')
 
     if _check_legacy_args():


zariiii9003 (approved):

zariiii9003:
By the way, why didn't that error show up in CI?

andlaus:
I don't know: It only appeared on import in one of my scripts (which didn't even use any of the CLI stuff), but not when invoking the `cantools` command. I wasn't in the mood to investigate but it might be that my local version of python-can is too old or too new?

zariiii9003:
cantools pins the latest python-can version so i don't think that's the issue. But i guess it's fine now.

trnila:
It conflicts with legacy args:
https://github.com/cantools/cantools/blob/d5b414519bb586ebe22df8ad2527cc1d1bb7a23f/src/cantools/subparsers/monitor.py#L650-L664

I would rather drop legacy `-f` aliasing `--fd` as its not too big effort to write extra `-d` instead of rewriting full `--filter-regex`.

Maybe your script is using program arguments like `-i`, `--interface`, `-b`, `--interface`, `--bus-type`, ... and hence triggering detection of legacy args?
https://github.com/cantools/cantools/blob/d5b414519bb586ebe22df8ad2527cc1d1bb7a23f/src/cantools/subparsers/monitor.py#L589
... but all subparses seems to be called only from `_main` / `_load_subparser`

andlaus:
> I would rather drop legacy -f aliasing --fd as its not too big effort to write extra -d instead of rewriting full --filter-regex.

the alias for `--fd` was there first (and IIRC it applies to all CLI tools), so I am not so sure about this. That said, I agree that the "legacy flag" should be dropped, but IMO it should be dropped *in addition* because people (read: me) tend to interpret a `-f` flag as `--force`...

> Maybe your script is using program arguments like -i, --interface, -b, --interface, --bus-type, ... and hence triggering detection of legacy args?

weirdly enough, the script in question did not use *any* CLI parsing infrastructure, the issue happened on `import cantools`. I should probably have a look because the argument parser should IMO not be set up at that point...

PR #791, opened 2026-06-04
`database.dump_file()`: always write the baudrate of the bus into DBCs

andlaus:
This only affects database objects which were not created by loading a DBC file: If the database file to be dumped to a DBC file was loaded from another DBC, the `Bitrate` attribute was passed through. (That said, if the bitrate of the high-level bus object was modified by the application, the unmodified value was written to the file.)

### DIFF
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -679,7 +679,14 @@ def _dump_attribute_definitions(database: InternalDatabase) -> list[str]:
 
     # create 'Baudrate' attribute definition
     if len(database.buses) == 1 and database.buses[0].baudrate is not None:
-        definitions['Baudrate'] = ATTRIBUTE_DEFINITION_BAUDRATE
+        # add a definition of the `Baudrate` attribute if it is needed
+        # and not yet present
+        if 'Baudrate' not in definitions:
+            definitions['Baudrate'] = ATTRIBUTE_DEFINITION_BAUDRATE
+    # remove the definition of the `Baudrate` attribute if it is
+    # not needed
+    elif 'Baudrate' in definitions:
+        del definitions['Baudrate']
 
     # create 'VFrameFormat' and 'CANFD_BRS' attribute definitions if bus is CAN FD
     if _bus_is_canfd(database):
@@ -781,11 +788,9 @@ def _dump_attributes(database: InternalDatabase, sort_signals: type_sort_signals
 
     attributes: list[type_sort_attribute] = []
 
-    # we use a copy of the original cantools database object with the
-    # .dbc objects added...
-    assert database.dbc is not None
-
-    # remove the old 'Baudrate' attribute if it exists
+    # remove the old 'Baudrate' attribute if it exists. We synchronize
+    # that DBC attribute with the `.baudrate` attribute of the
+    # high-level bus description object below.
     if database.dbc.attributes.get('Baudrate') is not None:
         del database.dbc.attributes['Baudrate']
 
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -146,6 +146,13 @@ ATTRIBUTE_DEFINITION_LONG_SIGNAL_NAME = AttributeDefinition(
     kind='SG_',
     type_name='STRING')
 
+ATTRIBUTE_DEFINITION_BAUDRATE = AttributeDefinition(
+    name='Baudrate',
+    default_value=125_000,
+    type_name='INT',
+    minimum=0,
+    maximum=10*1024*1024)
+
 ATTRIBUTE_DEFINITION_VFRAMEFORMAT = AttributeDefinition(
     name='VFrameFormat',
     default_value='StandardCAN',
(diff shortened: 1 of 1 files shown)


src/cantools/database/can/formats/dbc.py
```diff
@@ -770,16 +781,38 @@ def _dump_attributes(database: InternalDatabase, sort_signals: type_sort_signals
 
     attributes: list[type_sort_attribute] = []
 
+    # we use a copy of the original cantools database object with the
+    # .dbc objects added...
+    assert database.dbc is not None
```
zariiii9003: I don't like assertions in production code. Either this is always true and can be omitted or there should be a proper error message.

In this case a few lines earlier there already is a check for 'if database.dbc is None'.
If you addedthe assertion for mypy, then i'd prefer assigning the dbc to a variable like
```python
database_dbc = database.dbc
if database_dbc is None:
    return ba
```

src/cantools/database/can/formats/dbc.py
```diff
@@ -670,6 +677,10 @@ def _dump_attribute_definitions(database: InternalDatabase) -> list[str]:
     if 'GenSigStartValue' not in definitions and _need_startval_def(database):
         definitions['GenSigStartValue'] = ATTRIBUTE_DEFINITION_GENSIGSTARTVALUE
 
+    # create 'Baudrate' attribute definition
+    if len(database.buses) == 1 and database.buses[0].baudrate is not None:
```
zariiii9003, line 681: Maybe use the same form as the other definitions:
`if 'Baudrate' not in definitions and _need_baudrate_def(database):`.

In `_need_baudrate_def` you could check whether any of the buses in the database has a baudrate value.

src/cantools/database/can/formats/dbc.py
```diff
@@ -770,16 +781,38 @@ def _dump_attributes(database: InternalDatabase, sort_signals: type_sort_signals
 
     attributes: list[type_sort_attribute] = []
 
+    # we use a copy of the original cantools database object with the
+    # .dbc objects added...
+    assert database.dbc is not None
+
+    # remove the old 'Baudrate' attribute if it exists
```
zariiii9003: This needs a bit more explanation. It took me a while to understand why this is necessary.

My first suggestion would have been to update the dbc baudrate value in the `Bus.baudrate` setter. Then i noticed the Bus object does not contain dbc attributes and the baudrate is saved in the database object instead.

src/cantools/database/can/formats/dbc.py
zariiii9003: hmm no, there is no way to define multiple buses in dbc anyway. So that case should probably fail earlier.

andlaus: ok. leave it as-is?

src/cantools/database/can/formats/dbc.py
andlaus: the point of this is to make mypy happy. I think that returning silently is not loud enough, because IMHO it is a bad implementation mistake and it should go up in flames. `assert` raises `AssertationError`, we could possibly raise something else?

src/cantools/database/can/formats/dbc.py
andlaus: right. something like
```suggestion
    # remove the old 'Baudrate' attribute if it exists. We synchronize that DBC attribute
    # with the `.baudrate` attribute of the high-level bus description object
    # below.
```

 I just noticed that the definition for the Baudrate attribute is added unconditionally, even if we do not write a Baudrate attribute. I think this is okay, but if you have a different opinion, I'll change it...

src/cantools/database/can/formats/dbc.py
zariiii9003: If the attribute definition already exists, it should stay unmodified in my opinion, so i'd still prefer the `if 'Baudrate' not in definitions` check.

src/cantools/database/can/formats/dbc.py
zariiii9003: Yes looks good to me.

src/cantools/database/can/formats/dbc.py
zariiii9003: But `if database.dbc is None` is already checked on line 779, so if you prefer an exception instead of returning an empty list, it should be raised there.

andlaus: right. I removed the assert. thanks!

src/cantools/database/can/formats/dbc.py
andlaus: I added it conditionally, but I do not like trivial helper functions which are called exactly once, so I used some ifs for that...

zariiii9003: I guess i'm pedantic here, but i asked to add the check to avoid noise in a roundtrip scenario (read, edit, write). Now you added the check but also added the deletion 😄

Anyhow, feel free to merge.

zariiii9003 (approved):

src/cantools/database/can/formats/dbc.py
andlaus: I suppose I missed that. thanks for the approval :)

PR #792, opened 2026-06-06
Fix restrictive ifname pattern in logreader.CandumpPattern

### LINKED ISSUE #519: logreader can't parse possible network interface names
I found out by trying to decode on an interface named "vcan-i". The dash made the regex fail, I couldn't find a lot of informations on what is allowed to be a network device name on linux ([some infos here](https://unix.stackexchange.com/questions/451368/allowed-chars-in-linux-network-interface-names)).
As this is a virtual device I'm working around by changing the name so that it has no dashes.

nelfin:
[net/core/dev.c:dev_valid_name](https://github.com/torvalds/linux/blob/8e65320d91cdc3b241d4b94855c88459b91abf66/net/core/dev.c#L1325) only limits interface names to "must be valid file name and not contain spaces", so e.g. "can-123" is a valid interface name.

    $ sudo ip link add vcan.0 type vcan
    $ sudo ip link set dev vcan.0 up
    $ cansend vcan.0 123#0001020304050607
    $ candump vcan.0
      vcan.0  123   [8]  00 01 02 03 04 05 06 07

Fixes #519

---

I went with "non-blank", which is slightly less restrictive, but I figured that it wasn't worth trying to match the semantics exactly (since we don't do things like compare against IFNAMSIZ etc.)

### DIFF
--- a/src/cantools/logreader.py
+++ b/src/cantools/logreader.py
@@ -99,7 +99,7 @@ class CandumpDefaultPattern(CandumpBasePattern):
     # vcan0  1F0   [8]  00 00 00 00 00 00 1B C1   '.......Á'
     #(Ignore anything after the end of the data to work with candump's ASCII decoding)
     pattern = re.compile(
-        r'^\s*?(?P<channel>[a-zA-Z0-9]+)\s+(?P<can_id>[0-9A-F]+)\s+\[\d+\]\s*(?P<can_data>remote request|[0-9A-F ]*).*?$')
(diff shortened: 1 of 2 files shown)


andlaus:
> I went with "non-blank", which is slightly less restrictive, but I figured that it wasn't worth trying to match the semantics exactly (since we don't do things like compare against IFNAMSIZ etc.)

blank names are [not allowed](https://github.com/torvalds/linux/blob/8e65320d91cdc3b241d4b94855c88459b91abf66/net/core/dev.c#L1327-L1328) as are ".", ".." or names containing a slashes, colons or spaces.

That said, I think it is okay to match a superset of what the kernel accepts, as long as it always matches the whole valid interface name...

PR #793, opened 2026-06-06
Clean up monitor script and remove deprecated functionality

ilanbiala:
It's been about a year since https://github.com/cantools/cantools/pull/745 landed, and there's some extra cruft in `monitor.py` that can be removed to simplify it a bit. The changes in #732 can also be removed since python-can incorporates that functionality in `bus_kwargs`.

The change for user scripts would be from:
```
cantools monitor ... -c can0 -b socketcand --host=192.168.0.10 --port=29536
```
to:
```
cantools monitor ... -c can0 -i socketcand --bus_kwargs host=192.168.0.10 port=29536
```

### DIFF
--- a/src/cantools/subparsers/__utils__.py
+++ b/src/cantools/subparsers/__utils__.py
@@ -1,5 +1,4 @@
-import re
-from collections.abc import Iterable, Sequence
+from collections.abc import Iterable
 
 from cantools.database.errors import DecodeError
 
@@ -10,7 +9,6 @@ from ..typechecking import (
     ContainerDecodeResultType,
     ContainerUnpackResultType,
     SignalDictType,
-    TAdditionalCliArgs,
 )
 
 MULTI_LINE_FMT = '''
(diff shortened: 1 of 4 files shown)


andlaus:
hm, I'm not sure as this might break user scripts. I tend to agree with the change because it has been almost a year since the legacy arguments produced deprecation warnings. maybe @zariiii9003 has some insights.

ilanbiala:
@andlaus I just pushed some additional changes, take a look. The gist of it is that `python-can` added `bus_kwargs`, which essentially took over `cantools`'s `extra_args` functionality. I added details of what the change to a user's script would be. If we have any part of release notes that we call out how to migrate and handle any breaking changes, this should be very easy for users to update.

zariiii9003 (approved):
Personally i'd keep these a little longer, but it's okay to break existing code if we bump the major version.
Changes look good to me.

ilanbiala:
Is there a specific timeline you'd have in mind for keeping this functionality? In my mind the migration is very simple and it's contained to just `cantools monitor`, so even though it'd be a breaking change and a major version bump users can easily make any needed updates.

andlaus:
> Personally i'd keep these a little longer, but it's okay to break existing code if we bump the major version.
Changes look good to me.

I'm pretty much indifferent. @ilanbiala: is there a reason why you want to have these arguments removed quickly?

ilanbiala:
`monitor.py` had a few areas where adding typing revolves around this deprecated code, and eventually I'd also like to refactor and replace `argparse` with something better. Simplifying all the CLIs helps with making that effort easier to tackle.

On a sidenote, Python 3.10 will be EOL in October, so I expect there will be a major version bump sometime in the next few months. We can wait until then when it's officially EOL, to minimize version churn for users. Please add other breaking changes to the v42 milestone and we can merge them all at that one point so we can keep `master` clean for features/bug fixes.

andlaus:
> @andlaus I just pushed some additional changes, take a look. The gist of it is that python-can added bus_kwargs, which essentially took over cantools's extra_args functionality.

yeah. it might be that newer versions of python-can add a `-f` flag to the `ArgumentParser` object. @zariiii9003: do you have any info here?

> If we have any part of release notes that we call out how to migrate and handle any breaking changes, this should be very easy for users to update.

I can add it to the release notes if I don't forget. (alternatively, you can tag the next release...)

andlaus:
> Please add other breaking changes to the v42 milestone and we can merge them all at that one point so we can keep master clean for features/bug fixes.

I don't have anything related to python 3.10 which merrits a major release at the moment. (this PR only merrits a minor release IMO (i.e., ´41.5.0´) even if this is not exactly semversion compliant.) That said, in the long run, I'd like to remove the diagnostics functionality (its users should migrate to `odxtools`) and flatten the somewhat byzantine directory structure. @zariiii9003, @juleq, @ilanbiala: any thoughs?

zariiii9003:
> yeah. it might be that newer versions of python-can add a `-f` flag to the `ArgumentParser` object. @zariiii9003: do you have any info here?

python-can uses `-f` for file paths, so that shouldn't be an issue here.

Regarding the version: I'd be not amused if a lib breaks code in a minor version, so let's do the major version bump even if there are no other big changes :eyes:

ilanbiala:
> Regarding the version: I'd be not amused if a lib breaks code in a minor version, so let's do the major version bump even if there are no other big changes 
Agreed here.

andlaus:
ok, let's get over this and start the 42-era...

PR #795, opened 2026-06-12
Add dependabot configuration

zariiii9003:
Just trying to trigger readthedocs while adding something useful...

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -5,6 +5,9 @@ on: [push, pull_request]
 env:
   PY_COLORS: "1"
 
+permissions:
+  contents: read
+
 jobs:
   codegen:
     runs-on:  ${{ matrix.os }}
@@ -15,6 +18,9 @@ jobs:
       fail-fast: false
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # 6.0.2
+      with:
+        fetch-depth: 0
+        persist-credentials: false
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # 6.2.0
       with:
@@ -39,6 +45,9 @@ jobs:
       fail-fast: false
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # 6.0.2
+      with:
+        fetch-depth: 0
+        persist-credentials: false
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # 6.2.0
       with:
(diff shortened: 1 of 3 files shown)


zariiii9003:
I resynchronized the hooks on readthedocs and deleted the dead webhooks in this repo.
The [PR build](https://cantools--795.org.readthedocs.build/en/795/) works now.

juleq:
Great, thanks.

zariiii9003:
@juleq since you are owner of the PyPI repo, could you set up trusted publishing?

.github/workflows/pythonpackage.yml
```diff
@@ -62,16 +71,19 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # 6.0.2
+      with:
+        fetch-depth: 0
+        persist-credentials: false
     - name: Set up Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # 6.2.0
```
andlaus, line 86: what's the reason for specifying the commit instead of the branch for the actions?

andlaus (approved):
LGTM

.github/workflows/pythonpackage.yml
zariiii9003: It's [generally recommeded](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions) due to recent security incidents.

juleq:
Have added a trusted publishers on PyPi. Looking into the workflow to find out what to do next ;).

zariiii9003:
> Have added a trusted publishers on PyPi. Looking into the workflow to find out what to do next ;).

I added the `pypi` environment name. So in PyPI you need set the name of the yaml file and the name of the environment.

juleq:
Ok, have removed and readded using the pypi env name:

- Repository: [cantools/cantools](https://github.com/cantools/cantools)
- Workflow: pythonpackage.yml
- Environment name: pypi

We happy?

zariiii9003:
I think so, we'll see on release :smile:

zariiii9003:
Forgot to adapt the workflow triggers, now the dependabot PR runs the checks twice -.-

juleq:
I could invite you to pypi as a backup. Or @andlaus. Or both ^^. Otherwise we would have to bug Erik again when I fail to show up.

zariiii9003:
If you like, my username is the same as here

juleq:
Did so for cantools. Cannot do for textparser, insufficient permissions.

zariiii9003:
> Did so for cantools. Cannot do for textparser, insufficient permissions.

Got the invitation, thank you!

andlaus:
> I could invite you to pypi as a backup. Or @andlaus. Or both ^^. Otherwise we would have to bug Erik again when I fail to show up.

You can invite me, but I currently cannot access my `andlaus` pypi account. You should better invite `hexdev` on pypi which belongs to our supplier who does the distribution of odxtools on pypi...

PR #798, opened 2026-06-12
Adapt workflow triggers

zariiii9003:
Avoid dependabot PRs running tests twice

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -1,6 +1,12 @@
 name: Python package
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    tags:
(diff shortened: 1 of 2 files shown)


zariiii9003:
@andlaus hope you don't mind the xfail

zariiii9003:
@andlaus i think the 42.0.0 release wasn't triggered because `v*` does not match our tags :/
I think `*.*.*` might work.

andlaus:
what's the problem? pypi currently features 42.0.1: https://pypi.org/project/cantools/

andlaus:
sorry for the noise: my SMTP server is not of the fast kind, so I only got the github notification email now. I see that you fixed this in #802...

PR #800, opened 2026-06-21
fix mypy errors with textparser 0.26.*

zariiii9003:


andlaus (approved):
LGTM

zariiii9003:
Just in time for numpy 2.5.0 to break it again :smiling_face_with_tear:

andlaus:
it's a hamster's wheel :D

andlaus:
@zariiii9003: do you see any way of fixing the numpy issue other than fixing matplotlib to an old version?

zariiii9003:
We could set a numpy version limit in the mypy tox environment and then use the tox env in Github actions.

andlaus:
we can also specify the matplotlib version depending on the python version in `pyproject.toml`:
```toml
[...]
plot = [
     "matplotlib == 3.10; python_version <= '3.10'",
     "matplotlib; python_version > '3.10'",
]
```

unfortunately this does not work because the python version which runs mypy is 3.14 that and it only pretends to be 3.10...

PR #801, opened 2026-06-23
Use tox for mypy and ruff in CI, fix git tag trigger

zariiii9003:


.github/workflows/pythonpackage.yml
```diff
@@ -95,18 +95,13 @@ jobs:
...
-        pip install mypy ruff
-        pip install -e .[plot]
+        pip install tox
     - name: Static type checking with mypy
       run: |
-        mypy --python-version 3.10 .
-        mypy --python-version 3.11 .
-        mypy --python-version 3.12 .
-        mypy --python-version 3.13 .
-        mypy --python-version 3.14 .
```
andlaus, line 106: this means that we only do type checking on the latest python version, right?

edit: no, this moved to the testenv...

tox.ini
```diff
@@ -43,8 +43,15 @@ commands =
 [testenv:mypy]
 deps =
 	mypy
+    numpy<2.5
```
andlaus, line 46: we probably will need to remove the version specifier once we drop support for 3.10...

andlaus:
nice, CI works again! Thanks!

tox.ini
zariiii9003: I think 3.11 will cause the same issue with numpy 2.5

PR #807, opened 2026-07-10
rename `{Attribute,AttributeDefinition}` to `Dbc{Attribute,AttributeDefinition}`

andlaus:
This is the first part of a pretty long series of pull requests about refactoring the DBC format handling code in cantools.

We start simple and rename the `Attribute` class to `DbcAttribute` and `AttributeDefinition` to `DbcAttributeDefinition`. The reason for this is that the fact that these classes are highly DBC-specific should be immediately clear from their names...


zariiii9003:
I think all dbc stuff can be moved into a subpackage `cantools.database.formats.dbc` similar to arxml, but I don't see the need to rename classes 🤔

Since this is just the first part, what else is planned?

andlaus:
so far, the the series contains the move to a separate subdirectory (quite at the end, though), getting rid of the dictionaries with magical entries (for attributes, relation attributes, environment variables, ...), full type annotations (making `mypy --strict` work for the DBC stuff) and a few scatter-shot cleanups. since it is fully bisectable, I can adapt the size of the PRs to your liking :)

andlaus:
okay, I opened a pull request with all the DBC changes which I would like to get in: #808. (as a draft PR. #808 is not meant to be merged but is intended as a sneak peek of what is to come.)

andlaus:
> I think all dbc stuff can be moved into a subpackage cantools.database.formats.dbc similar to arxml, but I don't see the need to rename classes 🤔

IMO reading code which contains `DbcAttribute` instead of `Attribute` makes it much clearer what it does. This is analogous to `DbcSpecifics`: using `Specifics` as the class name can IMO be quite confusing if there are specifics classes for other formats (currently there's just `AutosarSpecifics`).

andlaus:
ok, I moved this one to the end of the refactoring series. this PR has been superseeded by #809, closing.

PR #809, opened 2026-07-14
DBC cleanup, part1 (v2)

andlaus:
this is the "new" first part of the DBC cleanup PR series after I moved renaming `Attribute` to `DbcAttribute` to the end. (A lot of "fun" with merge conflicts was involved.) This PR makes the `_get_{value,minimum,maximum}()` helper functions class members of `AttributeDefinition`, changes `DbcParser.grammar()` return a `Grammar` object and changes all strings in `dbc.py` which use double quotes to single-quote strings for consistency.

### DIFF
--- a/src/cantools/database/can/attribute_definition.py
+++ b/src/cantools/database/can/attribute_definition.py
@@ -1,8 +1,9 @@
 from typing import Generic, TypeVar
 
-AttributeValueType = TypeVar('AttributeValueType', str, int, float)
+AttributeValue = str | int | float
+AttributeValueTypeVar = TypeVar('AttributeValueTypeVar', str, int, float)
 
-class AttributeDefinition(Generic[AttributeValueType]):
+class AttributeDefinition(Generic[AttributeValueTypeVar]):
     """A definition of an attribute that can be associated with attributes
     in nodes/messages/signals.
 
@@ -10,14 +11,14 @@ class AttributeDefinition(Generic[AttributeValueType]):
 
     def __init__(self,
                  name: str,
-                 default_value: AttributeValueType | None = None,
+                 default_value: AttributeValueTypeVar | None = None,
                  kind: str | None = None,
                  type_name: str | None = None,
                  minimum: int | float | None = None,
                  maximum: int | float | None = None,
                  choices: list[str] | None = None) -> None:
         self._name = name
-        self._default_value: AttributeValueType | None = default_value
+        self._default_value: AttributeValueTypeVar | None = default_value
         self._kind = kind
         self._type_name = type_name
         self._minimum = minimum
@@ -33,7 +34,7 @@ class AttributeDefinition(Generic[AttributeValueType]):
         return self._name
 
     @property
-    def default_value(self) -> AttributeValueType | None:
+    def default_value(self) -> AttributeValueTypeVar | None:
         """The default value that this attribute has, or ``None`` if
         unavailable.
 
@@ -42,7 +43,7 @@ class AttributeDefinition(Generic[AttributeValueType]):
         return self._default_value
 
     @default_value.setter
-    def default_value(self, value: AttributeValueType | None) -> None:
+    def default_value(self, value: AttributeValueTypeVar | None) -> None:
         self._default_value = value
 
     @property
--- a/src/cantools/database/can/attribute_definition.py
+++ b/src/cantools/database/can/attribute_definition.py
@@ -75,6 +75,13 @@ class AttributeDefinition(Generic[AttributeValueType]):
         self._minimum = value
 
     @property
+    def formatted_minimum(self) -> str:
+        """The defined minimum value of the attribute definition as it appears in the DBC file
+
+        """
+        return '' if self._minimum is None else f' {self._minimum}'
+
+    @property
     def maximum(self) -> int | float | None:
         """The maximum value of the attribute, or ``None`` if unavailable.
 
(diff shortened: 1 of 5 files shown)


src/cantools/database/can/formats/dbc.py
```diff
@@ -50,6 +51,9 @@
 from .dbc_specifics import DbcSpecifics
 from .utils import num
 
+# Attribute value types produced during parsing (before they become Attribute objects)
+AttributeValue = str | int | float
```
ilanbiala: Already defined here: https://github.com/andlaus/cantools/blob/2240f654f3186cea377ee1048538f01314b5900e/src/cantools/database/can/attribute_definition.py#L3

src/cantools/database/can/formats/dbc.py
```diff
@@ -748,7 +737,7 @@ def _dump_attribute_definition_defaults(database: InternalDatabase) -> list[str]
 
     for definition in definitions.values():
         if definition.default_value is not None:
-            if definition.type_name in ["STRING", "ENUM"]:
```
ilanbiala, line 751: [Ruff prefers double quotes by default](https://docs.astral.sh/ruff/settings/#format_quote-style). Is there a reason to churn here?

ilanbiala: In fact, since we [don't specify a preference](https://github.com/andlaus/cantools/blob/2240f654f3186cea377ee1048538f01314b5900e/pyproject.toml#L75-L102), looking at the root cause here of why Ruff isn't complaining about the single-quote strings above is probably more worthwhile, and then fixing that. Ruff should enforce consistency and that should be a separate auto-fix PR, not something that we have to manually change.

andlaus: > [Ruff prefers double quotes by default](https://docs.astral.sh/ruff/settings/#format_quote-style). Is there a reason to churn here?

personally I prefer double quotes as well, but in all of Erik's projects the default seems to be single quotes except for nested quotes. (approximately 90% of the existing cantools strings use single quotes.)

> In fact, since we [don't specify a preference](https://github.com/andlaus/cantools/blob/2240f654f3186cea377ee1048538f01314b5900e/pyproject.toml#L75-L102), looking at the root cause here of why Ruff isn't complaining about the single-quote strings above is probably more worthwhile, and then fixing that. Ruff should enforce consistency and that should be a separate auto-fix PR, not something that we have to manually change.

this change was not motivated by ruff ;)

src/cantools/database/can/formats/dbc.py
andlaus: right. fixed, thanks: https://github.com/cantools/cantools/pull/809/changes/af76ae77d9a281a9940d6a3a128ad2378ea9e139

src/cantools/database/can/attribute_definition.py
```diff
@@ -1,23 +1,24 @@
 from typing import Generic, TypeVar
 
-AttributeValueType = TypeVar('AttributeValueType', str, int, float)
+AttributeValue = str | int | float
```
zariiii9003, line 3: Is `AttributeValue` necessary? Can't we use `AttributeValueTypeVar` instead?

src/cantools/database/can/attribute_definition.py
```diff
@@ -74,6 +75,13 @@ def minimum(self) -> int | float | None:
     def minimum(self, value: int | float | None) -> None:
         self._minimum = value
 
+    @property
+    def formatted_minimum(self) -> str:
```
zariiii9003, line 79: I feel like the formatting does not belong into the dataclass, but 🤷

zariiii9003 (approved):
Approved. Sorry for the slow replies.

src/cantools/database/can/formats/dbc.py
ilanbiala: If the goal is to stick with what has been set, then that should be done in a separate PR by setting the Ruff style rule and running with auto-fix, no? It doesn't need to clutter PRs along the way.

src/cantools/database/can/attribute_definition.py
andlaus: maybe my understanding of this is incomplete, but aren't type variables primarily meant as the arguments for generics, a la `class Attribute(Generic[AttributeValueTypeVar]):`?

src/cantools/database/can/formats/dbc.py
andlaus: > If the goal is to stick with what has been set, then that should be done in a separate PR by setting the Ruff style rule and running with auto-fix, no?

maybe I'll do a quotes consistency PR for the rest of cantools after the DBC refactoring stuff has been merged, but I think for now it is better to limit the efforts to the DBC code. The change is also fairly trivial and IMO reverting this is not worth the squeeze (i.e., rebase conflicts in a dozen or so follow up commits).

>  It doesn't need to clutter PRs along the way.

The reason I'll chopped up the DBC refactoring into small fragments is exactly to make it easier to review since there are not 100 different things in the air at once. Also, I agree that "quotes consistency" is not an important issue, just that it does not hurt to have it and it is simple to review if the PR which contains it is small...

src/cantools/database/can/attribute_definition.py
andlaus: I would agree if `AttributeDefinition` was not a highly DBC-specific affair. (I don't care too much either.)

andlaus:
okay, let's get over this hump...

PR #808, opened 2026-07-13
Overview: DBC cleanup series

andlaus:
This PR is the full series of changes which I have planned for the DBC refactoring. Since it is huge and features many unreleated changes, this pull request is not intended to be merged, but to provide a sneak peek into what's to come for the smaller pull requests that are intended for review...

PR #810, opened 2026-07-16
`dbc.py`: rename a few functions and variables to reduce the mental load

andlaus:
In particular, nested functions like `get_name()` depended on the function which they were nested within. This confused me endlessly because stuff like `get_attributes()` in `load_signals()` did something quite different from `get_attributes()` in `load_messages()` a few dozen lines below...

### DIFF
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -24,7 +24,6 @@ from textparser import (
 )
 
 from cantools.database.can.internal_database import InternalDatabase
-from cantools.typechecking import Comments
 
 from ...conversion import BaseConversion
 from ...namedsignalvalue import NamedSignalValue
@@ -463,19 +462,17 @@ def get_dbc_name(name: str) -> str:
     return name
 
 
-def _get_node_name(attributes, name):
-    try:
-        return attributes['node'][name]['SystemNodeLongSymbol'].value
-    except (KeyError, TypeError):
-        return name
+def _get_node_long_name(attributes, node_short_name):
+    attrib = attributes.get('node', {}).get(node_short_name, {}).get('SystemNodeLongSymbol')
+    if attrib is not None:
+        return str(attrib.value)
+    return node_short_name
 
-
-def _get_environment_variable_name(attributes, name):
-    try:
-        return attributes['envvar'][name]['SystemEnvVarLongSymbol'].value
-    except (KeyError, TypeError):
-        return name
-
+def _get_envvar_long_name(attributes, envvar_short_name):
+    attrib = attributes.get('envvar', {}).get(envvar_short_name, {}).get('SystemEnvVarLongSymbol')
+    if attrib is not None:
+        return str(attrib.value)
+    return envvar_short_name
 
 def _dump_version(database: InternalDatabase) -> str:
     return '' if database.version is None else database.version
@@ -1113,8 +1110,8 @@ def _load_comments(tokens):
             node_name = item[1]
             comments[node_name] = item[2]
         elif kind == 'EV_':
-            environment_variable_name = item[1]
-            comments[environment_variable_name] = item[2]
+            envvar_name = item[1]
+            comments[envvar_name] = item[2]
 
     return comments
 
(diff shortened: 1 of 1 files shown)


src/cantools/database/can/formats/dbc.py
```diff
@@ -463,19 +462,17 @@ def get_dbc_name(name: str) -> str:
...
-    except (KeyError, TypeError):
-        return name
-
-
-def _get_environment_variable_name(attributes, name):
-    try:
-        return attributes['envvar'][name]['SystemEnvVarLongSymbol'].value
-    except (KeyError, TypeError):
-        return name
+def _get_node_long_name(attributes, node_short_name):
```
ilanbiala, line 466: Typing?

src/cantools/database/can/formats/dbc.py
```diff
@@ -1449,27 +1447,27 @@ def _load_signals(tokens,
     except KeyError:
         pass
 
-    def get_attributes(frame_id_dbc, signal):
+    def get_signal_attributes(frame_id_dbc, signal_name):
```
ilanbiala, line 1451: Repeated function above should be extracted and both can be deleted from within function scope.

src/cantools/database/can/formats/dbc.py
```diff
@@ -1449,27 +1447,27 @@ def _load_signals(tokens,
...
         """
 
         try:
-            return comments[frame_id_dbc]['signal'][signal]
+            return comments[frame_id_dbc]['signal'][signal_name]
         except KeyError:
             return None
 
-    def get_choices(frame_id_dbc, signal):
+    def get_signal_choices(frame_id_dbc, signal):
```
ilanbiala, line 1471: These function renames would be more useful when paired with refactors outside of these local scopes and into a reusable class. All of these nested functions are what really increases mental load.

ilanbiala: And lack of typing - reading, maintaining, and building on top of code that operates on dynamic, unstructured and untyped data is another big challenge.

src/cantools/database/can/formats/dbc.py
```diff
@@ -1657,23 +1655,23 @@ def get_attributes(frame_id_dbc: int) -> typing.Any:
         except KeyError:
             return None
 
-    def get_comment(frame_id_dbc: int) -> str | Comments | None:
+    def get_message_comment(frame_id_dbc):
```
ilanbiala: Why delete the type annotation in this PR? That's not the PR scope, and there isn't any code here that fixes it so this is going to result in churn.

src/cantools/database/can/formats/dbc.py
```diff
@@ -2140,8 +2138,8 @@ def dump_string(database: InternalDatabase,
                           sig_mux_values='\r\n'.join(sig_mux_values))
 
 
-def get_definitions_dict(definitions: typing.Any, defaults: typing.Any) -> OrderedDict[str, AttributeDefinitionType]:
```
ilanbiala, line 2143: Again, why remove..?

src/cantools/database/can/formats/dbc.py
andlaus: this will be done in a later PR of the DBC refactoring series ([sneak peek](https://github.com/cantools/cantools/pull/808)). The problem with the full batch is that is way too large to be reasonably reviewable, so I decided to split it into small batches.

src/cantools/database/can/formats/dbc.py
andlaus: I basically agree, but the point of this PR is to make the code more readable whilst introducing as few "drive by fixes" as possible to keep it reviewable...

src/cantools/database/can/formats/dbc.py
andlaus: Again, I agree that nested functions are to be avoided in all but exceptional cases (the current DBC code takes them *way* too far). This kind of refactoring ought to be much easier once the current DBC refactoring is done, though...

src/cantools/database/can/formats/dbc.py
andlaus: for whatever reason mypy started complaining ("returning Any from a typed function"). I agree that it should be kept, though. [fixed](6014f9700464a559f5e50ea92a409c8008500077), thanks for the catch.

src/cantools/database/can/formats/dbc.py
andlaus: same as above: new mypy complaints. (also fixed)

src/cantools/database/can/formats/dbc.py
andlaus: (also, be aware that `get_signal_attributes()` is subtly different from `get_message_attributes()`. this was a big part of my original confusion: reading message parsing code, it is all too easy to end up looking at the function for signal parsing.)

andlaus:
@ilanbiala, @zariiii9003: any objections to merging this?

src/cantools/database/can/formats/dbc.py
andlaus: an issue which complicates unifying the two nested `get_signal_attributes()` inner functions is that they use the `attributes` dictionary argument of the outer function.

zariiii9003 (approved):

PR #812, opened 2026-07-20
DBC cleanup, part 4: do not use exceptions for flow control

andlaus:
using exceptions in non-exceptional circumstances is bad style for performance, debugging and code understandability reasons (spaghetti code). That said, the DBC code IMO did not use them in an outrageously bad way because it did not use it to jump over several levels in the call stack...

### DIFF
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -24,7 +24,6 @@ from textparser import (
 )
 
 from cantools.database.can.internal_database import InternalDatabase
-from cantools.typechecking import Comments
 
 from ...conversion import BaseConversion
 from ...namedsignalvalue import NamedSignalValue
@@ -1404,18 +1403,17 @@ def _load_signal_groups(tokens, attributes):
 
         """
 
-        try:
-            return attributes[frame_id_dbc]['signal'][signal_short_name]
-        except KeyError:
-            return OrderedDict()
+        frame_signals = attributes.get(frame_id_dbc, {})
+        signal_map = frame_signals.get('signal', {})
+        return signal_map.get(signal_short_name, OrderedDict())
 
     def get_signal_long_name(frame_id_dbc, signal_short_name):
         signal_attributes = get_signal_attributes(frame_id_dbc, signal_short_name)
 
-        try:
-            return signal_attributes['SystemSignalLongSymbol'].value
-        except (KeyError, TypeError):
-            return signal_short_name
+        attrib = signal_attributes.get('SystemSignalLongSymbol')
+        if attrib is not None:
+            return attrib.value
+        return signal_short_name
 
     for signal_group in tokens.get('SIG_GROUP_',[]):
         frame_id = int(signal_group[1])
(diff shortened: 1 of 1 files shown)


andlaus:
@zariiii9003, @ilanbiala: any objections to merging this?

src/cantools/database/can/formats/dbc.py
```diff
@@ -52,6 +51,14 @@
 from .dbc_specifics import DbcSpecifics
 from .utils import num
 
+# make mypy complain if we assign a value to the empty dictionary but
+# do not pay the runtime performance penalty of MappingProxyType.
+if typing.TYPE_CHECKING:
```
zariiii9003, line 56: I think something like `EMPTY_DICT = typing.cast(Mapping[typing.Any, typing.Any], {})` would do the same.

src/cantools/database/can/formats/dbc.py
```diff
@@ -464,13 +471,13 @@ def get_dbc_name(name: str) -> str:
 
 
 def _get_node_long_name(attributes, node_short_name):
-    attrib = attributes.get('node', {}).get(node_short_name, {}).get('SystemNodeLongSymbol')
+    attrib = attributes.get('node', empty_dict).get(node_short_name, empty_dict).get('SystemNodeLongSymbol')
```
zariiii9003: I don't really like this chaining of `get()` that much.
I'd prefer assignment to meaningful variable names, checking for None explicitely and returning early whithout doing any unnecessary work.

zariiii9003 (approved):

src/cantools/database/can/formats/dbc.py
andlaus: agreed, but when I tried this, the  result was quite a bit longer and more difficult to read. (IMO python would benefit from a `None`-aware dot operator a la Dart's `?.`. )

That said, the next few PRs in this series will replace these "magic" dictionaries by dataclasses (cf. e.g., https://github.com/cantools/cantools/pull/808/changes/d27ab83b5230394543b8b6be199b0c78b4bf680a ) which -- while not completely eliminating the issue -- significantly reduce its impact.

src/cantools/database/can/formats/dbc.py
```diff
@@ -464,13 +471,13 @@ def get_dbc_name(name: str) -> str:
...
     if attrib is not None:
         return str(attrib.value)
     return node_short_name
 
 def _get_envvar_long_name(attributes, envvar_short_name):
-    attrib = attributes.get('envvar', {}).get(envvar_short_name, {}).get('SystemEnvVarLongSymbol')
```
ilanbiala, line 473: This seems like a change made to appease Mypy? But `.get()` should exist on any dictionary, so I'm a bit surprised this is needed. What is the error the comes up without the change?

src/cantools/database/can/formats/dbc.py
```diff
@@ -1740,16 +1738,12 @@ def get_message_protocol(frame_id_dbc: int) -> str | None:
...
+        if message_attributes is not None and 'SystemMessageLongSymbol' in message_attributes:
+            return str(message_attributes['SystemMessageLongSymbol'].value)
+        return message_short_name
 
-    def get_message_signal_groups(frame_id_dbc: int) -> list[SignalGroup] | None:
-        try:
-            return signal_groups[frame_id_dbc]  # type: ignore[no-any-return]
-        except KeyError:
-            return None
+    def get_message_signal_groups(frame_id_dbc):
```
ilanbiala, line 1744: Add back function signature typing?

src/cantools/database/can/formats/dbc.py
andlaus: changed: 764b1f5b098293b2105d8c0bf52fee202b4cd1b0 . thanks for the proposal!

src/cantools/database/can/formats/dbc.py
andlaus: the point here is not to satisfy mypy, but to not having to create a dict object for the default value regardless of whether it is used or not: since python creates these eagerly, we have to pay a (moderate) performance penalty if the requested entry exists if we are using `{}` instead of `EMPTY_DICT`. A better way would be factory functions for the default, a la `attributes.get('envvar', lambda: {})`, but python's standard dictionary types do not support that as far as I know...)

src/cantools/database/can/formats/dbc.py
andlaus: For some strange reason mypy started to complain:
```
> mypy src/
src/cantools/database/can/formats/dbc.py:1745: error: Returning Any from function declared to return "list[SignalGroup] | None"  [no-any-return]
Found 1 error in 1 file (checked 61 source files)
```

anyway -- adding comprehensive type hints are not too far off...

PR #813, opened 2026-07-22
DBC cleanup, part 5: replace the `comments` dictionary with "magic" entries by a dataclass

andlaus:
This improves things on the clarity of intend, typing and `dict.get()`-chaining fronts. The next two PRs in the series will do the same to attributes and relation attributes.

### DIFF
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -63,10 +63,10 @@ EMPTY_DICT = typing.cast("MappingProxyType[typing.Any, typing.Any]", {})
 @dataclass(slots=True, kw_only=True)
 class DbcComments:
     # comment for the database as a whole
-    database: str|None = None
+    database: str | None = None
 
     # comment for the CAN bus described by the database
-    bus: str|None = None
+    bus: str | None = None
 
     # comments for all nodes on the CAN bus:
     #     nodes[node_name] -> node_comment
@@ -1503,10 +1503,6 @@ def _load_signals(tokens,
         if message_signal_comments is None:
             return None
         return message_signal_comments.get(signal_name)
-
-        frame_cmt = comments.get(frame_id_dbc, EMPTY_DICT)
-        signal_cmt = frame_cmt.get('signal', EMPTY_DICT)
-        return signal_cmt.get(signal_name)
 
     def get_signal_choices(frame_id_dbc, signal):
         """Get choices for given signal.
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -4,6 +4,7 @@ import re
 import typing
 from collections import OrderedDict, defaultdict
 from copy import deepcopy
+from dataclasses import dataclass, field
 from decimal import Decimal
 
 from textparser import (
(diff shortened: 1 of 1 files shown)


andlaus:
@zariiii9003, @ilanbiala: any comments?

src/cantools/database/can/formats/dbc.py
```diff
@@ -1467,6 +1500,10 @@ def get_signal_comment(frame_id_dbc, signal_name):
         """Get comment for given signal.
 
         """
+        message_signal_comments = comments.signals.get(frame_id_dbc)
+        if message_signal_comments is None:
+            return None
+        return message_signal_comments.get(signal_name)
 
         frame_cmt = comments.get(frame_id_dbc, EMPTY_DICT)
```
zariiii9003: this is unreachable now?

src/cantools/database/can/formats/dbc.py
```diff
@@ -1091,34 +1117,41 @@ def _dump_environment_variables(database: InternalDatabase) -> list[str]:
...
-
-    for comment in tokens.get('CM_', []):
-        if not isinstance(comment[1], list):
-            # CANdb++ behaviour: all bus comments are concatenated
-            existing_comment = comments['database'].get('bus', '')
-            comments['database']['bus'] = existing_comment + comment[1]
+    comments = DbcComments()
+
+    for comment_tokens in tokens.get('CM_', []):
+        if not isinstance(comment_tokens[1], list):
```
zariiii9003, line 1122: I know you didn't change this, but is there a reason why we don't check if it's a string here?

src/cantools/database/can/formats/dbc.py
```diff
@@ -58,6 +59,31 @@
 
 EMPTY_DICT = typing.cast("MappingProxyType[typing.Any, typing.Any]", {})
 
+# The type of the object returned by _load_comments()
+@dataclass(slots=True, kw_only=True)
+class DbcComments:
+    # comment for the database as a whole
+    database: str|None = None
```
zariiii9003: I'd prefer spaces left and right of the pipe character 🤷

andlaus: sorry, I always forget. fixed: 9d8e2eca3f1ee905407bda36afd8a7061bc254fc (we should enforce a formatter in the CI system sometime in the future...)

src/cantools/database/can/formats/dbc.py
andlaus: honestly, I don't know: from the [`Grammar`](https://github.com/cantools/cantools/blob/master/src/cantools/database/can/formats/dbc.py#L339-L347) object, it looks like a string is indeed the only alternative to lists, but the `textparser` stuff is almost black magic to me...

src/cantools/database/can/formats/dbc.py
andlaus: oops, right. (I got lost in rebases.) Thanks, fixed: 9d8e2eca3f1ee905407bda36afd8a7061bc254fc

andlaus:
(the CI system currently fails because `ruff` recently started producing quite a few new complaints. #815 fixes that...

andlaus:
@zariiii9003: do you see any further issues with this?

zariiii9003 (approved):

PR #815, opened 2026-07-24
`plot`: make sure that parsing timestamps keeps working for python >= 3.15

andlaus:
The following warning is produced when running the unit tests on the current master version:

```
  src/cantools/subparsers/plot.py:229: DeprecationWarning: Parsing dates involving a day of month without a year specified is ambiguious
  and fails to parse leap day. The default behavior will change in Python 3.15
  to either always raise an exception or to use a different default year (TBD).
  To avoid trouble, add a specific year to the input & format.
  See https://github.com/python/cpython/issues/70647.
    out = datetime.datetime.strptime(user_input, p)
```

this patch works around this by using 1904 as the year for formats which do not specify one.

### DIFF
--- a/examples/dbc_io/main.py
+++ b/examples/dbc_io/main.py
@@ -24,8 +24,8 @@ try:
     print("Input cycle time: ", message.cycle_time)
     message.cycle_time = 2000
     print("Output cycle time:", message.cycle_time)
-except KeyError as e:
-    raise e
+except KeyError:
+    raise
 
 # Manipulate the message frame id.
 print("Input frame id: ", hex(message.frame_id))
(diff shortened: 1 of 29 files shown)


andlaus:
I added a patch fixing the many issues newly raised by `ruff` 0.16.0 which seems to have been released yesterday. This is a lot of code churn (sorry!) and has nothing to do with the subject of this PR, but it is necessary to make the CI system pass.

andlaus:
@zariiii9003, @ilanbiala: any objections?

zariiii9003 (approved):
LGTM 👀

PR #818, opened 2026-07-29
Fix crash on candump ERRORFRAME lines in logreader

nishantshah0:
logreader crashed with binascii.Error: Odd-length string on candump error-frame lines because the data group [0-9A-F ]* consumed the leading E of ERRORFRAME. This killed candump | cantools decode pipelines with error frames enabled, since decode.py reads stdin through this parser.

Changes: the data group now matches complete byte pairs ((?:[0-9A-F]{2} *)*). ERRORFRAME cannot be consumed pair-wise because its second character is not hex. The marker is captured explicitly and CandumpBasePattern.unpack returns None for it, mirroring how PCANTracePatternV11 handles Error rows. groupdict().get() is used because CandumpDefaultLogPattern shares this unpack but has no error_frame group. Tests added for all three view patterns plus a stream test.

Behavior notes: cantools decode now prints error-frame lines verbatim (the keep_unknowns path) instead of dying, so no output is lost. A truncated odd-length data field (00 00 0) previously crashed; it now parses the complete pairs and ignores the dangling nibble.

Open question: skipping matches the PCAN precedent, but an is_error_frame flag on DataFrame would mirror python-can's Message. Happy to rework to that if preferred.

### DIFF
--- a/src/cantools/logreader.py
+++ b/src/cantools/logreader.py
@@ -71,7 +71,12 @@ class BasePattern:
 
 class CandumpBasePattern(BasePattern):
 
-    def unpack(self, match_object: re.Match[str]) -> DataFrame:
+    def unpack(self, match_object: re.Match[str]) -> DataFrame | None:
+        if match_object.groupdict().get('error_frame'):
+            # Error frames carry no payload. Skip them, mirroring how
+            # PCANTracePatternV11 skips 'Error' rows.
+            return None
+
         channel = match_object.group('channel')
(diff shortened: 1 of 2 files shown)


andlaus:
Looks good, merging. Thanks!

> Open question: skipping matches the PCAN precedent, but an is_error_frame flag on DataFrame would mirror python-can's Message. Happy to rework to that if preferred.

no opinions here from my side ;)

PR #816, opened 2026-07-29
DBC cleanup, part 6: replace the `attributes` dictionary with "magic" entries by a dataclass

andlaus:
this PR does exactly the same to attributes as what #813 did to comments and for the same reasons.

PR #819, opened 2026-08-07
DBC cleanup, part 7: avoid a magic dict for relation attributes

andlaus:
this does the same for relation attributes as what #813 did for comments and #816 did for attributes: replace the dictionary containing "magic" entries with by a dataclass.

after this PR, the "magical dictionaries" will fortunately have been dealt with...

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -301,5 +301,9 @@ autodoc_type_aliases = {
 # disable specific warnings
 nitpick_ignore = [
     ("py:class", "TypeAliasForwardRef"),
+    ("py:class", "Choices"),
+    ("py:class", "AttributeType"),
+    ("py:class", "AttributeDefinitionType"),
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -301,5 +301,9 @@
 # disable specific warnings
 nitpick_ignore = [
     ("py:class", "TypeAliasForwardRef"),
+    ("py:class", "Choices"),
+    ("py:class", "AttributeType"),
+    ("py:class", "AttributeDefinitionType"),
```
andlaus, line 306: Note: I have no idea why sphinx suddenly requires these types to be ignored...

zariiii9003: I think some of these are due to imports beeing behind the TYPE_CHECKING condition. But even after removing it i couldn't fix all warnings.... so :shrug:

zariiii9003 (approved):
Looks good to me.

PR #820, opened 2026-08-10
DBC cleanup, part 8: Rename quite a few variable and function names

andlaus:
Unfortunately, this PR is quite large but it does not contain any functional changes and should thus be relatively easy to review: It renames additional functions and variables which I missed in #810. As with #810, the purpose is to make the intend of these variables and functions clearer,


andlaus:
@zariiii9003, @ilanbiala: I intend to merge this one on Friday if no objections are raised...

andlaus:
let's get this merged...

PR #821, opened 2026-08-13
decode: don't print undecodable contained messages twice with --single-line

NotAFlightRisk:
Piping a candump log through `cantools decode --single-line` against an ARXML with container PDUs, and any contained message that fails to decode comes out listed twice, which reads like the container really did carry two copies of it:

```
$ echo "  vcan0  066   [8]  0A 0B 0C 04 E2 D8 7F D6" | cantools decode --single-line tests/files/arxml/system-4.2.arxml
  vcan0  066   [8]  0A 0B 0C 04 E2 D8 7F D6 :: OneToContainThemAll(message1: Undecodable data: e2 d8 7f d6, message1: Undecodable data: e2 d8 7f d6)
```

Without `--single-line` the same frame prints it once. The test in here fails without the change.

Left the format itself alone - the multi-line version prefixes these with `header_id##hexdata ::` and the single-line one doesn't, but that looks deliberate so I've not touched it.

### DIFF
--- a/src/cantools/subparsers/__utils__.py
+++ b/src/cantools/subparsers/__utils__.py
@@ -67,7 +67,6 @@ def _format_container_single_line(message : Message,
         if isinstance(cm, Message):
             if isinstance(signals, bytes):
                 formatted_cm = f'{cm.name}: Undecodable data: {signals.hex(" ")}'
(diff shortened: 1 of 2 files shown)


andlaus:
nice, thanks! merging...

PR #822, opened 2026-08-14
DBC cleanup, part 9: add comprehensive type hints

andlaus:
with this, `mypy --strict` passes for `dbc.py`.

The main mechanism which this commit introduces is the `dbc_assert_type(x, T)` function. It returns `x` if `x` is of type `T` or raises a `ParseError` if this is not the case. For most cases, this allows to tell the type checker the type of `x` without introducing too much code churn.

Besides this I did a few additional renames (much fewer than #820) , but I did not find the nerves of factoring them out into a separate PR.


andlaus:
@ilanbiala: this one should be of special interest to you...

andlaus:
As with #820, I'll merge this on Friday if no objections are raised...

andlaus:
no vetos. Let's merge...

PR #824, opened 2026-08-16
kcd: derive auto message length from the furthest-ending signal

NotAFlightRisk:
`length` is optional on a KCD `<Message>`, and when it's left off the loader sizes the message from whichever signal starts furthest in, then treats that one signal's length as the end. Furthest-starting isnt furthest-ending. An 8-bit signal at offset 40 in one `<MuxGroup>` beats a 56-bit signal at offset 8 in another, so the message gets 6 bytes when it needs 8 and the file won't load at all:

```
$ cantools dump tests/files/kcd/auto_message_length.kcd
error: KCD: "The signal Signal2 does not fit in message Message1."
```

Mux groups are the only way in. Signals in a plain message can't overlap under strict mode, so whichever starts last ends last too. `the_homer.kcd` has two auto-length multiplexed messages and neither trips it, beacuse every mux-group signal in `ABS` and `Radio` is 8 bits at the same offset, so the old formula lands on the right number anyway. I compared old against new for every length-less message under `tests/files/kcd` and `examples/`, and new fixture is the only one that shifts.

New fixture rather than another message in `the_homer.kcd`, since that file's message count and indices get asserted in a lot of places.

Explicit `length=` still wins, including where it's too small for the signals in the message. That file is broken and should keep failing.

### DIFF
--- a/src/cantools/database/can/formats/kcd.py
+++ b/src/cantools/database/can/formats/kcd.py
@@ -222,8 +222,9 @@ def _load_message_element(message, bus_name, nodes, strict, sort_signals):
 
     if length == 'auto':
         if signals:
-            last_signal = max(signals, key=start_bit)
-            length = (start_bit(last_signal) + last_signal.length + 7) // 8
+            last_bit = max(start_bit(signal) + signal.length
+                           for signal in signals)
+            length = (last_bit + 7) // 8
         else:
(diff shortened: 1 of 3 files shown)


andlaus:
good find. Merging, thanks!

PR #825, opened 2026-08-19
Fix for Issue #205 (Overlapping Signals / Multiplexors)

### LINKED ISSUE #205: Overlapping signals causes ValueError: Short data.
I have attached the dbc to recreate this issue at the bottom under details - GitHub won't let me attach it.
It is fairly common for signals to overlap within dbc files. A good example of this is error flagging and fault codes. A byte might be made up of bit flags that will indicate issues while the byte (fault code) can help detemine the broader area at fault. The dbc file will do a better job at explaining it than I can!

The code to reproduce the error:
```
from cantools import database
dbc = database.load_file('example.dbc',strict=False)
dbc.decode_message(287,[0])
```

The above code will cause:

```
  File "C:\Users\User01\AppData\Local\Programs\Python\Python38\lib\site-packages\cantools\database\can\database.py", line 389, in decode_message
    return message.decode(data, decode_choices, scaling)
  File "C:\Users\User01\AppData\Local\Programs\Python\Python38\lib\site-packages\cantools\database\can\message.py", line 789, in decode
    return self._decode(self._codecs, data, decode_choices, scaling)
  File "C:\Users\User01\AppData\Local\Programs\Python\Python38\lib\site-packages\cantools\database\can\message.py", line 748, in _decode
    decoded = decode_data(data,
  File "C:\Users\User01\AppData\Local\Programs\Python\Python38\lib\site-packages\cantools\database\utils.py", line 95, in decode_data
    unpacked.update(formats.little_endian.unpack(bytes(data[::-1])))
ValueError: Short data.
```

**Is there a way around this?**

<details>

VERSION ""

NS_ : 
	NS_DESC_
	CM_
	BA_DEF_
	BA_
	VAL_
	CAT_DEF_
	CAT_
	FILTER
	BA_DEF_DEF_
	EV_DATA_
	ENVVAR_DATA_
	SGTYPE_
	SGTYPE_VAL_
	BA_DEF_SGTYPE_
	BA_SGTYPE_
	SIG_TYPE_REF_
	VAL_TABLE_
	SIG_GROUP_
	SIG_VALTYPE_
	SIGTYPE_VALTYPE_
	BO_TX_BU_
	BA_DEF_REL_
	BA_REL_
	BA_DEF_DEF_REL_
	BU_SG_REL_
	BU_EV_REL_
	BU_BO_REL_
	SG_MUL_VAL_

BS_:

BU_:

BO_ 287 Message01: 1 Vector__XXX
 SG_ Flag1 : 0|1@1+ (1,0) [0|1] "" Vector__XXX
 SG_ Flag2 : 1|1@1+ (1,0) [0|1] "" Vector__XXX
 SG_ Flag3 : 2|1@1+ (1,0) [0|1] "" Vector__XXX
 SG_ Flag4 : 3|1@1+ (1,0) [0|1] "" Vector__XXX
 SG_ Flag5 : 4|1@1+ (1,0) [0|1] "" Vector__XXX
 SG_ Flag6 : 5|1@1+ (1,0) [0|1] "" Vector__XXX
 SG_ Flag7 : 6|1@1+ (1,0) [0|1] "" Vector__XXX
 SG_ Flag8 : 7|1@1+ (1,0) [0|1] "" Vector__XXX
 SG_ Status : 0|8@1+ (1,0) [0|255] "" Vector__XXX

</details>

friessssss:
## Summary

Fixes #205.

cantools still treated overlapping signals as invalid at decode/encode time, even when a database was loaded with `strict=False`. That shows up as `ValueError: Short data` (or a `DecodeError` wrapping it).

This is a real pattern in automotive DBs: bit flags plus a wider status/fault field covering the same bits. The original report used eight 1-bit flags overlapping an 8-bit `Status` on a 1-byte Intel message. The same failure also happened for Motorola signals and for multiplexed SYM frames where a status word shares bits with a flag.

`strict=True` is unchanged: overlapping signals still raise at load time. With `strict=False`, decode and encode now return consistent values for both the wide field and the overlapping bits.

## Approach

`bitstruct` cannot represent overlapping fields in a single format string. Overlapping signals are now split into internally non-overlapping layers, each compiled as its own codec. Decode unpacks every layer and merges the results. Encode packs each layer and ORs them together.

Non-overlapping messages still use a single layer, so the common path is unchanged.

## Tests

- The original #205 DBC (little-endian flags + status byte)
- Overlapping big-endian signals
- A reduced multiplexed SYM frame (`tests/files/sym/issue_205.sym`)

## Notes

Load overlapping databases with `strict=False`, same as before. Encoding also needs `strict=False` so the overlap check in `Message.encode()` does not reject the payload.

### DIFF
--- a/src/cantools/database/utils.py
+++ b/src/cantools/database/utils.py
@@ -14,6 +14,7 @@ from typing import (
 from ..typechecking import (
     ByteOrder,
     Choices,
+    CompiledFormatDict,
     Formats,
     SignalDictType,
     SignalMappingType,
@@ -231,7 +232,7 @@ def _group_non_overlapping_signals(
     return layers
 
 
-def _compile_bitstruct(fmt: str, names: list[str]):
+def _compile_bitstruct(fmt: str, names: list[str]) -> CompiledFormatDict:
     try:
         return bitstruct.c.compile(fmt, names)
     except Exception:
--- a/src/cantools/database/utils.py
+++ b/src/cantools/database/utils.py
@@ -30,10 +30,12 @@ if TYPE_CHECKING:
     from ..database.can.signal import Signal
     from ..database.diagnostics import Data
 
+import bitstruct  # type: ignore
+
 try:
     import bitstruct.c  # type: ignore
 except ImportError:
-    import bitstruct  # type: ignore
+    pass
 
 
 def format_or(items: list[int | str]) -> str:
@@ -104,6 +106,21 @@ def _encode_signal_values(signals: Sequence[Union["Signal", "Data"]],
     return raw_values
 
 
+def _pack_formats(formats: Formats, raw_signal_values: dict[str, int | float]) -> int:
+    packed_union = (
+        int.from_bytes(formats.big_endian.pack(raw_signal_values), "big")
+        | int.from_bytes(formats.little_endian.pack(raw_signal_values), "little")
+    )
+
+    for big_endian, little_endian in formats.extra_layers:
+        packed_union |= (
+            int.from_bytes(big_endian.pack(raw_signal_values), "big")
+            | int.from_bytes(little_endian.pack(raw_signal_values), "little")
+        )
+
+    return packed_union
+
+
 def encode_data(signal_values: SignalMappingType,
                 signals: Sequence[Union["Signal", "Data"]],
                 formats: Formats,
@@ -113,11 +130,7 @@ def encode_data(signal_values: SignalMappingType,
         return 0
 
     raw_signal_values = _encode_signal_values(signals, signal_values, scaling)
-    big_packed = formats.big_endian.pack(raw_signal_values)
-    little_packed = formats.little_endian.pack(raw_signal_values)
-    packed_union = int.from_bytes(big_packed, "big") | int.from_bytes(little_packed, "little")
-
-    return packed_union
+    return _pack_formats(formats, raw_signal_values)
 
 
 def decode_data(data: bytes,
@@ -151,6 +164,9 @@ def decode_data(data: bytes,
             **formats.big_endian.unpack(data),
             **formats.little_endian.unpack(data[::-1]),
         }
+        for big_endian, little_endian in formats.extra_layers:
+            unpacked.update(big_endian.unpack(data))
+            unpacked.update(little_endian.unpack(data[::-1]))
     except (bitstruct.Error, ValueError) as e:
         # bitstruct returns different errors in PyPy and cpython
         raise DecodeError("unpacking failed") from e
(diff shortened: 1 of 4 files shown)


src/cantools/database/utils.py
```diff
@@ -30,10 +31,12 @@
     from ..database.can.signal import Signal
     from ..database.diagnostics import Data
 
+import bitstruct  # type: ignore
+
 try:
     import bitstruct.c  # type: ignore
 except ImportError:
-    import bitstruct  # type: ignore
```
andlaus, line 36: failing completely if `bitstruct.c` is not available is a horrible idea...

src/cantools/database/utils.py
```diff
@@ -104,6 +107,21 @@ def _encode_signal_values(signals: Sequence[Union["Signal", "Data"]],
     return raw_values
 
 
+def _pack_formats(formats: Formats, raw_signal_values: dict[str, int | float]) -> int:
```
andlaus, line 108: a comment what this function is supposed to do would be good.

src/cantools/database/utils.py
```diff
@@ -104,6 +107,21 @@ def _encode_signal_values(signals: Sequence[Union["Signal", "Data"]],
...
+    packed_union = (
+        int.from_bytes(formats.big_endian.pack(raw_signal_values), "big")
+        | int.from_bytes(formats.little_endian.pack(raw_signal_values), "little")
+    )
+
+    for big_endian, little_endian in formats.extra_layers:
```
andlaus, line 124: please add a comment about what extra layers are

src/cantools/database/utils.py
```diff
@@ -189,6 +206,39 @@ def decode_data(data: bytes,
...
+            layer_ends.append(start + signal.length)
+
+    return layers
+
+
+def _compile_bitstruct(fmt: str, names: list[str]) -> CompiledFormatDict:
+    try:
+        return bitstruct.c.compile(fmt, names)
+    except Exception:
+        return bitstruct.compile(fmt, names)
```
andlaus, line 248: without importing `bitstruct`,  this will crash as well. please use the previous approach and simply call `bitstruct.compile()`

andlaus (commented):
I did not go into this too deeply, but overall it looks good. I'm wondering if this works with multiplexers, though...

src/cantools/database/utils.py
friessssss: Agreed ! Good catch, this is fixed now, was an artifact of the way I was testing initially. Reverted to the original strategy

src/cantools/database/utils.py
friessssss: I moved this to a function for ease of re-use. Returned the initial import to the original approach, so this should be much more valid now.

src/cantools/database/utils.py
friessssss: Added !

src/cantools/database/utils.py
friessssss: Commented !

friessssss:
@andlaus I added an explicit multiplex test to the test suite to make sure that this works as expected. Also the `issue_205.sym` is stripped down version of the symbol file that caused me to stumble upon this issue in the first place, and it is now used in tests. Let me know if there are any other tests / improvements that you would like to see !

PR #827, opened 2026-08-22
Only swallow format errors while probing a database string

### LINKED ISSUE #814: Logging handler errors during load are mislabeled as "UnsupportedDatabaseFormatError: DBC: ..."
### Summary

`load_file` / `load_string` wrap each format attempt in a bare `except Exception`, so **any** exception raised while parsing — including errors that have nothing to do with the database format (e.g. an `OSError` raised by a `logging` handler while cantools emits its own `WARNING`) — is swallowed and re-reported as `UnsupportedDatabaseFormatError: <file>: DBC: "<unrelated error>"`.

This hides the real error and its traceback, and misleads the user into thinking the DBC file is malformed when it is not.

### Environment

- cantools: 42.0.3
- Python: 3.10.11
- OS: Windows 10/11

### How it happens in practice

Loading a DBC that contains duplicate message names / frame ids makes cantools call `LOGGER.warning("Overwriting message ...")` (see `database.py`, `_add_message`). If the logging handler that receives those records fails to write (in our case a Windows console handle that was no longer a real console → `OSError [WinError 1] Incorrect function`), the exception propagates out of `add_dbc_string`, is caught by the broad `except Exception` in `load_string`, and is reported as a DBC parse error:

```
cantools.database.errors.UnsupportedDatabaseFormatError: <file>: DBC: "[WinError 1] Incorrect function"
```

### Minimal, deterministic reproduction (no Windows needed)

```python
import logging
import cantools

# DBC with two messages sharing the same name + frame id -> triggers LOGGER.warning(...)
DBC = (
    'VERSION ""\n\nBU_: ECU\n\n'
    'BO_ 256 DupMsg: 8 ECU\n SG_ Sig1 : 0|8@1+ (1,0) [0|0] "" ECU\n\n'
    'BO_ 256 DupMsg: 8 ECU\n SG_ Sig2 : 8|8@1+ (1,0) [0|0] "" ECU\n'
)
open("dup.dbc", "w").write(DBC)

class ExplodingHandler(logging.Handler):
    # stands in for any handler whose write fails (e.g. broken console stream)
    def emit(self, record):
        raise OSError(1, "Incorrect function")

lg = logging.getLogger("cantools")
lg.addHandler(ExplodingHandler())
lg.propagate = False

cantools.database.load_file("dup.dbc", strict=False)
```

**Output:**

```
cantools.database.errors.UnsupportedDatabaseFormatError: DBC: "[Errno 1] Incorrect function"
```

### Expected behavior

An `OSError` from the logging subsystem is unrelated to the database format and should propagate unchanged (with its original traceback), rather than being reported as an unsupported/invalid DBC.

### Actual behavior

The unrelated `OSError` is caught and re-raised as `UnsupportedDatabaseFormatError: ...: DBC: "..."`, obscuring the real cause.

### Suggested fix

In `load_string`, only catch parse-related exceptions when probing formats (e.g. `textparser.ParseError` / `TokenizeError` and cantools' own error types), and let other exceptions (`OSError`, `KeyboardInterrupt`, etc.) propagate. This keeps format auto-detection working while surfacing genuine I/O or programming errors correctly.

Poseidonas:
Fixes #814.

`load_string()` probes each supported format in turn and keeps the failure from each one so that it can report all five reasons together if none of them works. Every attempt was wrapped in a bare `except Exception`, so an error that says nothing about the format was recorded as if it did.

The reporter's case, reproduced unchanged on `c1dcd39`:

```
cantools.database.errors.UnsupportedDatabaseFormatError: DBC: "[Errno 1] Incorrect function"
```

The file parses fine. The `OSError` comes from a logging handler that fails while cantools emits its own "Overwriting message ..." warning, and the traceback pointing at it is lost.

After the change the same script raises the `OSError` itself, with the handler still in the traceback.

**Which exceptions actually mean "wrong format"**

Rather than guess, I fed every parser each of the other formats plus junk and an empty string, and recorded what came out — 30 combinations:

| exception | occurrences |
| --- | --- |
| `ElementTree.ParseError` | 12 |
| `cantools.errors.Error` | 6 |
| `textparser.ParseError` | 5 |
| `ValueError` | 2 |
| `KeyError` | 2 |

The last two are wider than I would like. They come from three places that reject a wrong root tag with a plain builtin:

- `arxml/__init__.py:57` — `ValueError("No XML namespace specified or illegal root tag name ...")`
- `kcd.py:475` — `ValueError("Expected root element tag ...")`
- `cdd.py:244` — `KeyError("Could not find ECUDOC root element!")`

Making those raise `ParseError` would be tidier and would let the probe narrow to cantools' own errors plus the two parser ones. I left it alone because those exceptions are visible to anyone calling `add_arxml_string()` and friends directly, and several tests assert on them — that seemed like a separate decision rather than something to fold into a bugfix. Happy to do it if you would prefer that shape.

**Verification**

- Auto-detection unchanged: loaded all **131** database files under `tests/` with no explicit format, before and after. Same outcome for every file, including the 12 that already failed.
- Junk, empty string and binary input still raise `UnsupportedDatabaseFormatError`.
- Two tests added. The first fails on `master` with `UnsupportedDatabaseFormatError` and passes with the change; the second guards the other direction, that a genuinely bad format is still reported as one.
- `ruff` and `mypy` clean. Suite goes 415 → 417 passing. `test_cache_env_var` fails on my machine both before and after — unrelated to this change.

### DIFF
--- a/src/cantools/database/__init__.py
+++ b/src/cantools/database/__init__.py
@@ -5,6 +5,9 @@ import os
 import warnings
 from contextlib import nullcontext
 from typing import Any, TextIO
+from xml.etree import ElementTree
+
+import textparser
 
 try:
     import diskcache  # type: ignore
@@ -24,6 +27,20 @@ from .errors import (
     EncodeError,
     Error,
     UnsupportedDatabaseFormatError,
+)
+
+# Exceptions which mean "this string is not in the format being probed for".
+# Anything else - an OSError from a logging handler, a MemoryError, a bug in
+# cantools itself - is not a statement about the format and must reach the
+# caller with its own traceback instead of being reported as a parse failure.
+_FORMAT_PROBE_ERRORS: tuple[type[Exception], ...] = (
+    Error,
+    textparser.ParseError,
+    ElementTree.ParseError,
+    # arxml and kcd reject an unexpected root tag with a plain ValueError,
+    # cdd reports a missing ECUDOC root with a KeyError.
+    ValueError,
+    KeyError,
 )
 
 
@@ -339,25 +356,25 @@ def load_string(string: str,
     if database_format in ['arxml', None]:
         try:
             return load_can_database('arxml')
-        except Exception as e:
+        except _FORMAT_PROBE_ERRORS as e:
             e_arxml = e
 
     if database_format in ['dbc', None]:
         try:
             return load_can_database('dbc')
-        except Exception as e:
+        except _FORMAT_PROBE_ERRORS as e:
             e_dbc = e
 
     if database_format in ['kcd', None]:
         try:
             return load_can_database('kcd')
-        except Exception as e:
+        except _FORMAT_PROBE_ERRORS as e:
             e_kcd = e
 
     if database_format in ['sym', None]:
         try:
             return load_can_database('sym')
-        except Exception as e:
+        except _FORMAT_PROBE_ERRORS as e:
             e_sym = e
 
     if database_format in ['cdd', None]:
@@ -365,7 +382,7 @@ def load_string(string: str,
             db = diagnostics.Database()
             db.add_cdd_string(string)
             return db
-        except Exception as e:
+        except _FORMAT_PROBE_ERRORS as e:
             e_cdd = e
 
     if database_format is not None:
--- a/tests/test_database.py
+++ b/tests/test_database.py
@@ -2627,6 +2627,51 @@ class CanToolsDatabaseTest(unittest.TestCase):
         with self.assertRaises(UnsupportedDatabaseFormatError):
             cantools.database.load(StringIO(''))
 
+    def test_load_string_lets_unrelated_errors_through(self):
+        """An error unrelated to the format must not be reported as a parse failure.
+
+        Format probing runs each parser in turn and remembers why it failed. An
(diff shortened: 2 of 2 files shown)


andlaus:
Before I have a deeper look at this, I would like to see all format parsers to only raise `cantools.errors.ParseError` if it could not be loaded (separate PR for this please).

Poseidonas:
Makes sense — with every loader raising `ParseError`, the probe only has one thing to catch. I looked at the current state: DBC already does (10 of 12 sites), KCD has 1 `ValueError`, SYM 3, ARXML about 26 across the system and ECU-extract loaders, plus the `textparser`/`ElementTree` errors that escape the tokenisers. I'll open that as a separate PR first — each loader raising `ParseError` with the same message it raises today — and rebase #827 on it. One question: should an ARXML variant the loader does not support (today `NotImplementedError`) also become `ParseError`, or stay as it is since it is not a parse failure?

andlaus:
> should an ARXML variant the loader does not support (today NotImplementedError) also become ParseError, or stay as it is since it is not a parse failure?

it should stay a `NotImplementedError`: the file format could be loaded and partially understood (version could be recognized). This is a semantic difference compared to the file not adhering to the format's syntax.

PR #828, opened 2026-08-25
kcd: capture the full text of a Notes element

### LINKED ISSUE #709: Only first note of KCD object gets captured
when Importing a KCD, the contents of the `<Notes>` field is placed into the 'comments' dict of the message class. If multiple notes are present, only the first note is passed into the comments dictionary of the message. It would be nice if all notes would be placed into the comments dictionary with an integer index as a key, as the KCD file spec does not allow for attributes on the notes field.

ChrisJr404:
Closes #709.

When loading a KCD, the text of a `<Notes>` element was read with `element.text`, which only returns the text before the first child element. So a note like `<Notes>First line<br/>Second line</Notes>` came back as just "First line" and anything after the first child was silently dropped, on both messages and signals.

This gathers the whole note with `itertext()` instead. For the common case of a single text node the result is identical, so every existing KCD fixture loads exactly as before. In #709 @andlaus suggested simply concatenating the notes, which is what this does.

Added a test with a multi-part note that fails on master and passes here. Full test suite, ruff and mypy are all clean.

### DIFF
--- a/src/cantools/database/can/formats/kcd.py
+++ b/src/cantools/database/can/formats/kcd.py
@@ -98,11 +98,13 @@ def _load_signal_element(signal, nodes):
                 LOGGER.debug("Ignoring unsupported signal value attribute '%s'.",
                              key)
 
-    # Notes.
-    try:
-        notes = signal.find('ns:Notes', NAMESPACES).text
-    except AttributeError:
-        pass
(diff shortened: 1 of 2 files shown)


andlaus:
looks good. thanks for the contribution-

PR #829, opened 2026-08-26
fix: textparser minimal version

klow68:
Issue:
 
In dbc.py imports from textparser [MatchObject](https://github.com/cantools/textparser/blob/ddd6812d8fd9cdfe05e5c122578713402ee14e21/textparser/__init__.py#L82) which is only available starting from 0.26

 Fix:
 Set minimal version of textparser: 0.26


andlaus:
thanks for the catch!

PR #830, opened 2026-08-26
format loaders: raise ParseError for input that cannot be loaded

Poseidonas:
Prerequisite for #827, split out as requested there.

The dbc, kcd, sym, arxml and cdd loaders now raise `cantools.database.errors.ParseError` whenever the input cannot be loaded, instead of a mix of `ValueError`, `KeyError`, `textparser.ParseError` and `xml.etree.ElementTree.ParseError`. Message text is unchanged, with two exceptions: `num()` in `formats/utils.py` had an unreachable fallback, so a bad number now gets its intended message rather than Python's `float()` one; and the three cdd `KeyError`s lose the quotes `str(KeyError)` adds.

`NotImplementedError` for the unsupported ARXML direction in `ecu_extract_loader.py` is kept as is pending the question on #827; two converted sites in `system_loader.py` ("only AUTOSAR 3 and 4", "non-canonical references not yet supported") read as unsupported rather than malformed and can flip with it.

`load_string()` in `database/__init__.py` is not touched; #827 will be rebased on top of this.

### DIFF
--- a/src/cantools/database/__init__.py
+++ b/src/cantools/database/__init__.py
@@ -1,6 +1,7 @@
 __all__ = ["Bus", "Database", "DecodeError", "EncodeError", "Message",
            "Node", "Signal", "dump_file", "load", "load_file", "load_string"]
 
+import logging
 import os
 import warnings
 from contextlib import nullcontext
@@ -24,7 +25,10 @@ from .errors import (
     EncodeError,
     Error,
     UnsupportedDatabaseFormatError,
+    format_load_error,
 )
+
+LOGGER = logging.getLogger(__name__)
 
 
 def _resolve_database_format_and_encoding(database_format: str | None,
@@ -122,7 +126,8 @@ def load_file(filename: StringPathLike,
     Raises an
     :class:`~cantools.database.UnsupportedDatabaseFormatError`
     exception if given file does not contain a supported database
-    format.
+    format and a :class:`NotImplementedError` if the file uses a
+    variant of a supported format which is not supported.
 
     >>> db = cantools.database.load_file('foo.dbc')
     >>> db.version
@@ -249,7 +254,8 @@ def load(fp: TextIO,
     Raises an
     :class:`~cantools.database.UnsupportedDatabaseFormatError`
     exception if given file-like object does not contain a supported
-    database format.
+    database format and a :class:`NotImplementedError` if it uses a
+    variant of a supported format which is not supported.
 
     >>> with open('foo.kcd') as fin:
     ...    db = cantools.database.load(fin)
@@ -297,7 +303,8 @@ def load_string(string: str,
     Raises an
     :class:`~cantools.database.UnsupportedDatabaseFormatError`
     exception if given string does not contain a supported database
-    format.
+    format and a :class:`NotImplementedError` if the string uses a
+    variant of a supported format which is not supported.
 
     >>> with open('foo.dbc') as fin:
     ...    db = cantools.database.load_string(fin.read())
@@ -311,13 +318,12 @@ def load_string(string: str,
             f"expected database format 'arxml', 'dbc', 'kcd', 'sym', 'cdd' or "
             f"None, but got '{database_format}'")
 
-    e_arxml = None
-    e_dbc = None
-    e_kcd = None
-    e_sym = None
-    e_cdd = None
+    def load_database(fmt: str) -> can.Database | diagnostics.Database:
+        if fmt == 'cdd':
+            diagnostics_db = diagnostics.Database()
+            diagnostics_db.add_cdd_string(string)
+            return diagnostics_db
 
-    def load_can_database(fmt: str) -> can.Database:
         db = can.Database(frame_id_mask=frame_id_mask,
                           strict=strict,
                           sort_signals=sort_signals)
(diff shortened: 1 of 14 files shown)


fallenmi (changes requested):
The remaining unwrapped numeric conversions mean this does not yet establish the `ParseError`-only loader contract that #827 depends on.

On the current head (`5002715`), this syntactically valid KCD with an invalid input attribute still leaks a builtin exception:

```python
from cantools.database.can.formats import kcd

text = '''<NetworkDefinition xmlns="http://kayak.2codeornot2code.org/1.0">
<Bus name="B"><Message id="0x1" name="M">
<Signal name="S" offset="0" length=""/>
</Message></Bus></NetworkDefinition>'''
kcd.load_string(text, strict=False)
```

The result is `ValueError: invalid literal for int() with base 10: ''`, not `cantools.database.errors.ParseError`. The same path has other direct `int()` conversions for signal offsets, label values, mux counts, message IDs/intervals, lengths, and baud rates. A syntactically valid SYM that references an undefined signal can likewise still escape as `KeyError` during `_load_message`.

Today the broad probe in `load_string()` happens to catch these. Once #827 is rebased to catch only `ParseError`, they will escape rather than be retained in `UnsupportedDatabaseFormatError`. Please normalize the remaining input-derived failures (with regression coverage for a post-tokenization/attribute-conversion case), or narrow the promised contract before #827 relies on it.

AI assistance disclosure: OpenAI Codex was used to inspect the diff and run the adversarial reproducer; the technical finding and review submission were authorized by the account owner.

Poseidonas:
Confirmed — reproduced the KCD case on 5002715 (`length=""` escapes as a builtin `ValueError` from `kcd.load_string`). I'll add a boundary in each loader so failures after tokenisation also surface as `ParseError`, with the cases you list as regression tests, and push to this PR.

Poseidonas:
Done in 17ce3af. Every loader's `load_string()` is now wrapped at its boundary, so `ValueError`, `KeyError`, `AttributeError`, `TypeError`, `IndexError` and `ArithmeticError` raised after tokenisation are reported as `ParseError` chained from the original; the explicit checks keep their messages and a `ParseError` raised inside passes through unchanged. Regression tests cover the empty KCD `length`, a SYM message referencing an undefined signal, a non-integer DBC frame id and a string value for an `INT` attribute, a non-numeric ARXML `BAUDRATE`, a non-numeric CDD `TEXTMAP` bound and a CDD document without an `ECU` element — each via the loader and via `load_string(..., database_format=...)`.

fallenmi (changes requested):
Thanks — the original KCD and SYM post-tokenization cases are fixed on exact head `17ce3af`, and the complete `tests/test_database.py` file passes here (201/201). One non-input boundary is still being converted by the new loader-wide decorator: `sort_signals` is a documented user callback, but its exceptions are now relabeled as parse failures.

Using `tests/files/dbc/vehicle.dbc` and a callback that raises a sentinel `TypeError`, direct `dbc.load_string(text, sort_signals=callback)` raises a new `ParseError` whose cause is the sentinel on this head. On previous head `5002715`, the identical call raises the original sentinel object unchanged. I reproduced the same regression for callback-raised `AttributeError`, `ValueError`, and `KeyError`.

This means the narrowed probe planned in #827 would still classify unrelated user-code failures as unsupported database content, which is the same error-boundary problem #814 is trying to remove. Please limit conversion to input-derived loader failures (or otherwise let user callback failures escape unchanged) and add a sentinel callback regression alongside the new malformed-input matrix.

Follow-up review performed with OpenAI Codex assistance under the account owner authorization.

src/cantools/database/utils.py
```diff
@@ -471,3 +474,37 @@ def sort_choices_by_value(choices: Choices) -> Choices:
...
+    ArithmeticError,
+    AttributeError,
+    IndexError,
+    KeyError,
+    TypeError,
+    ValueError,
+)
+
+
+def raises_parse_error(func: Callable[_P, _T]) -> Callable[_P, _T]:
```
andlaus: instead of introducing this decorator which does basically the same as `load_string()` (what's the point?), I would like to see `load_string()` to be modified: if it catches a `ParseError`, no message is produced and the next format is tried, if it catches a `NotImplementedError`, it prints the error message belonging to the exception and refrains from probing further formats, and on any other exception it should print the error message belonging to the exception (plus adding that this is a buglet in cantools that ought to be fixed) and continue probing.

andlaus: the task for this PR is then to ensure that no exceptions other than `ParseError` and `NotImplementedError` are raised by `load_string()` for the whole test suite.

Poseidonas:
Done as described: `load_string()` now swallows `ParseError` and tries the next format, re-raises `NotImplementedError` immediately, and for anything else logs the exception's message with a note that it is a bug in cantools and keeps probing; the decorator is gone. To make the second part hold, the test suite runs with a guard that fails any test in which a loader raises something other than a cantools `Error`/`NotImplementedError`, and every file under `tests/files` is loaded as every format under it. The corpus itself was already clean, but mutating single values in valid files turned up the sites in the last commit (KCD/SYM/DBC/CDD/ECU-extract `int()`/`float()` on input, SYM undefined signal references, DBC undefined attributes and out-of-range enum indices, ARXML number parsing, a zero denominator and an assert on scale limits); all now raise `ParseError` naming the offending value. Once this lands, #827 is redundant.

Poseidonas:
On the callback point: under the scheme andlaus asked for, an exception raised by a user callback such as `sort_signals` is not a statement about the input, so `load_string()` keeps its message, appends the "this is a bug in cantools" note, and continues probing the other formats (see `test_load_string_sort_signals_exception`).

PR #831, opened 2026-08-26
add support for float back to _load_bus

Jrendl:
Version 43.0.0 removed support for FLOAT type bus baudrate. My use case specifies a FLOAT in the dbc even though it does not have any fractional elements. I am happy to do a validation of the decimal here to make sure that its `.0` but wanted to start with the simplest fix.

### DIFF
--- a/src/cantools/database/can/formats/dbc.py
+++ b/src/cantools/database/can/formats/dbc.py
@@ -1991,13 +1991,7 @@ def _load_bus(attributes: DbcAttributes, comments: DbcComments) -> Bus | None:
     bus_baudrate_attr = attributes.database.get('Baudrate')
     bus_baudrate = None
     if bus_baudrate_attr is not None:
-        try:
-            bus_baudrate = dbc_assert_type(bus_baudrate_attr.value, int)
-        except ParseError as original_error:
-            try:
-                bus_baudrate = int(dbc_assert_type(bus_baudrate_attr.value, float))
-            except ParseError as second_error:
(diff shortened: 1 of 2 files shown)


src/cantools/database/can/formats/dbc.py
```diff
@@ -1991,7 +1991,13 @@ def _load_bus(attributes: DbcAttributes, comments: DbcComments) -> Bus | None:
...
     bus_baudrate = None
     if bus_baudrate_attr is not None:
-        bus_baudrate = dbc_assert_type(bus_baudrate_attr.value, int)
+        try:
+            bus_baudrate = dbc_assert_type(bus_baudrate_attr.value, int)
+        except ParseError as original_error:
+            try:
+                bus_baudrate = int(dbc_assert_type(bus_baudrate_attr.value, float))
+            except ParseError as second_error:
+                raise second_error from original_error
```
andlaus: using exceptions for flow control ought to be avoided. This should work as well:
```suggestion
        bus_baudrate = int(float(bus_baudrate_attr.value))
```
can you test?

Jrendl: Sounds good. Yes I'll test. want me to provide any specific artifact from testing?

Jrendl: tested with my real DBC and it was able to load. I also added a unit test.

andlaus:
cool, thanks for the catch!

PR #833, opened 2026-09-04
DBC cleanup, part 12: rename `{Attribute,EnvironmentVariable}` to `Dbc{Attribute,EnvironmentVariable}`

andlaus:
Analogous to the `DbcSpecifics` class, IMO this makes it clearer that these classes are DBC-specific. 

This is the final PR of the DBC cleanup series (phew!).


andlaus:
as with the last few PRs I will merge this one next Friday (September 11) if there are no comments...

andlaus:
let's get that wrapped up...

PR #834, opened 2026-09-12
dbc: support multiple access nodes in environment variables

RyoheiHashimoto:
Closes #483.

The DBC grammar allows an `EV_` entry to list several access nodes separated
by commas:

```
environment_variable = 'EV_' env_var_name ':' env_var_type
    '[' minimum '|' maximum ']' unit initial_value ev_id
    access_type access_node {',' access_node} ';'
```

Only the first one was accepted, so a file using more than one failed to load:

```
UnsupportedDatabaseFormatError: DBC: "Invalid syntax at line 39, column 61: "EV_ MultipleNodes: 0 [0|100] "" 0 2 DUMMY_NODE_VECTOR0 NodeA>>!<<,NodeB,NodeC;""
```

### Implementation notes

- The field is parsed with `DelimitedList('WORD')`, which is what `BO_TX_BU_`
  — the other comma-separated node list in this grammar — already uses.
- `DbcEnvironmentVariable` takes and exposes `access_nodes` as a list, so the
  nodes survive a dump/load round trip.
- `access_node` still reads the first node and warns when that drops
  information; its setter is deprecated in favour of `access_nodes`.

### Testing

- `tests/files/dbc/environment_variable_multiple_access_nodes.dbc` covers the
  single- and multiple-node forms and a dump/load round trip. It is the file
  that produces the error above on `master`.
- Of the 82 existing `tests/files/dbc/*.dbc` files, the 78 that load on
  `master` dump byte-identically before and after this change; they contain
  19 `EV_` lines between them.

PR #835, opened 2026-09-12
dbc: support ENVVAR_DATA_ entries

RyoheiHashimoto:
`ENVVAR_DATA_` marks an environment variable as being of data type and gives
its length:

```
environment_variable_data = 'ENVVAR_DATA_' env_var_name ':' data_size ';'
```

The keyword is in the tokenizer's keyword set but has no grammar rule, so a
single such line makes the whole file unloadable:

```
UnsupportedDatabaseFormatError: DBC: "Invalid syntax at line 41, column 1: ">>!<<ENVVAR_DATA_ DataEnvVar: 10;""
```

Seven of the 147 `.dbc` files in `sebastianwilczek/CANoe-Configurations`,
`zhuhaijun753/CANoe_V348_Project`, `yiyuchenguang/CANoe-Demo` and
`oxibus/can-dbc` carry `ENVVAR_DATA_` definitions, and none of them loads on
`master`. Six load with this change; the seventh fails on overlapping signals,
which is unrelated. Nothing that loaded before regresses.

The length is kept as `data_size`, so it survives a dump/load round trip; a
variable without an entry reports `None`. An entry naming a variable with no
`EV_` entry is ignored rather than rejected and dropped on dump, which is how
an orphan `CM_ EV_` comment already behaves.

`tests/files/dbc/environment_variable_data.dbc` is the file that produces the
error above on `master`. #700 attempted this in 2024 and has been conflicting
since.


src/cantools/database/can/formats/dbc/dbc_environment_variable.py
```diff
@@ -20,7 +20,8 @@ def __init__(self,
                  access_type: str,
                  access_node: str,
                  comment: str | None,
-                 dbc_specifics: "DbcSpecifics") -> None:
+                 dbc_specifics: "DbcSpecifics",
+                 data_size: int | None = None) -> None:
```
andlaus:
```suggestion
                 data_size: int | None,
                 dbc_specifics: "DbcSpecifics") -> None:
```

andlaus (commented):
seems good. the `DbcSpecifics` object should stay the last constructor argument, though...

RyoheiHashimoto:
Moved, and rebased onto master now that #834 is in.

One consequence: with `dbc_specifics` last, `data_size` cannot carry a default,
so the constructor now requires it. Let me know if you would rather have the
default and take `data_size` at the end instead.

andlaus:
good. let's merge.

> One consequence: with dbc_specifics last, data_size cannot carry a default, so the constructor now requires it. Let me know if you would rather have the default and take data_size at the end instead.

I think that in the context of this PR it is fine. Generally, I think we should consider making functions with many and/or unintuitive arguments keyword-only....

